PR #8, opened 2018-05-13
Dev-chain in a Docker

rootmos:
Two contributions for the price of one:
* Add a Dockerfile for running nodes, and in particular a local development chain
* Use non-cryptographic RNG when guessing next seed
  - I needed to make this change since the test node I was running stalled when no
    more entropy was available, which felt kind of silly.

@samcamwilliams @allquantor @DamonSweeney


rootmos:
For the interested: the container is made and used to stabilize the integration tests of https://github.com/toknapp/arweave4s (see https://github.com/toknapp/arweave4s/pull/17).

PR #15, opened 2018-07-16
Copy over the prometheus nif

rootmos:
@arweave-kyle @samcamwilliams  I don't think the `cp` of the prometheus library to `priv` is working as intended, the application still looks for it in its own `priv` directory.


samcamwilliams:
Thanks, @rootmos -- this is fixed in #16 which we are just about to merge.

Sam

Issue #20, opened 2018-07-25
Document issue handling


themue:
Also change hint on naming in `version-control.md`.

PR #23, opened 2018-07-26
Add issue management documentation

themue:
Solves issue #20 .

### DIFF
--- a/docs/issue-management.md
+++ b/docs/issue-management.md
@@ -1,0 +1,42 @@
+# Issue Management
+
+## Motivation
+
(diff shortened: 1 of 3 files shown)


docs/issue-management.md
```diff
@@ -0,0 +1,42 @@
+# Issue Management
+
+## Motivation
+
+Like the **Arweave** code base is hosted on Github we also manage all issues using the according issue services. A defined set of tags helps sorting the issues as well as milestones help to group them for upcoming iterations and releases.
+
+## Process
+
+### Entering
```
arweave-kyle: I think this should say _Reporting_ instead of _Entering_

arweave-kyle:
How come we're removing the `.gitignore` from `priv`, I don't think we would ever want to commit anything from this directory so it should be ignored?

PR #40, opened 2018-08-28
Maintenance/quiet elli metrics

llaisdy:
The white space changes obscure the code changes (next time I will rejig whitespace first ...). There are two sets of functional changes, which address these two issues:

- around L225: sometimes elli_request:peer(Req) returns undefined, which crashes ar_util:parse_peer/1.

- around L329 for size/2: sometimes one of size(X, response_headers) or size(X, response_body) returns undefined, which crashes size(X, response).

Those two issues cause about 12 errors/minute in the logs.

### DIFF
--- a/src/ar_metrics.erl
+++ b/src/ar_metrics.erl
@@ -190,51 +190,56 @@ handle_event(_Event, _Args, _Config) -> ok.
 
 handle_full_response(Type, [Req, Code, _Hs, _B, {Timings, Sizes}], _Config) ->
 %%handle_full_response(Type, [Req, Code, _Hs, _B, Timings], _Config) ->
-  Path = path(),
-  case {elli_request:method(Req), elli_request:raw_path(Req)} of
-    {'GET', Path} -> ok;
-    _ ->
-      Labels = labels(Req, Code),
-      TypedLabels = case Type of
-                      request_complete -> ["full" | Labels];
-                      chunk_complete -> ["chunks" | Labels] %;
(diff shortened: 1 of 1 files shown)


llaisdy:
The badarith error (2nd above) happens about once every ten seconds per node; the other is more intermittent.

majastanislawska:
You don't need to do commit that only changes whitespace in whole file to fix trivial issue
https://github.com/majastanislawska/arweave/commit/5a35711d350a3af81c8e94901186a1784fcb2ad7

llaisdy:
Hello!  Thank you for your comment.  Sorry I didn't see your commit.  I agree with your point about
whitespace.  As I say in my PR comment, in future I'll tidy whitespace as a separate commit.

The reason I've added `size_safe_add/2` instead of using a default in `proplists:get_value/3` is
that I haven't studied elli enough to know whether the returned `undefined` is because the proplist
doesn't have the requested key, or the proplist does have the key with the value `undefined`.  Using
`proplists:get_value/3` wouldn't protect against the latter, e.g.:

```erlang
> proplists:get_value(q, [{q,undefined}], 0).
undefined
```

majastanislawska:
You don't have to protect youself from things that doesnt happen.
Erlang is not language for defensive programming anyway. its part of language philosophy.

llaisdy:
The badarith error happens about once every ten seconds per node.

majastanislawska:
Thank you for confirming that was really plain sexism, not just general arrogance

PR #42, opened 2018-09-04
Maintenance/quiet elli metrics

llaisdy:


src/ar_metrics.erl
```diff
@@ -326,20 +331,24 @@ duration(StartKey, EndKey, Timings) ->
...
+		undefined ->
+			case proplists:get_value(file, Sizes) of
+				undefined ->
+					proplists:get_value(resp_body, Sizes);
+				FileSize -> FileSize
+			end;
+		ChunksSize -> ChunksSize
+	end.
+
+size_safe_add(undefined, undefined) -> 0;
```
themue: Isn't it more a `safe_size_add`?  Also would like comments here and on other functions. Yeah, I know, I'm one of those bad comment-all-and-everything-guys. 😁

PR #62, opened 2018-09-25
Fix JSON struct to block when expect only transaction IDs 

themue:
Now function accepts both, list of full transactions as well as of IDs, but only returns lists of IDs.

Fixes error when requesting block from peer and it is delivered in v7 or older format. Only a quick fix, more useful for handling of multiple different active node versions would be an analysis of version header on HTTP client interface level. Should be discussed before merging of this hotfix.

Fixes issue #63

### DIFF
--- a/src/ar_serialize.erl
+++ b/src/ar_serialize.erl
@@ -154,7 +154,7 @@ json_struct_to_block(JSONBlock) ->
 		(TX) when is_binary(TX) ->
 			ar_util:decode(TX);
 		(TX) ->
-			Temp = tx_safely_to_json_struct(TX),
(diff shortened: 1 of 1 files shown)


llaisdy:
Looks good.  Why the weird line indenting in the test?

themue:
Oh, will fix it, editor setting for tab/space indentation.

llaisdy:
I suspect the involvement of copy and paste ;)

(The following `full_block_roundtrip_test()` has similar weird indenting.)

themue:
Bingo, will retab the whole file. Thankfully one simple command.

llaisdy:
👍

PR #72, opened 2018-10-23
Fix non-parallel pmap

martin-torhage:
I don't know how this could affect the system, so please look into this carefully.


martin-torhage:
Invalid branch name. Will create a new PR.

PR #75, opened 2018-10-25
Fix / Parallelize pmap

martin-torhage:
I don't know how this could affect the behaviour of the system, so please look into this carefully.


martin-torhage:
These changes got included in #77

PR #82, opened 2018-11-06
Move fetching recall block in ar_http_iface to ar_node_worker

martin-torhage:
As usual, view the commits one by one in commit order (i.e. not in the timestamp order Github displays them).


martin-torhage:
@llaisdy I've fixed the error you pointed out (on Slack) which failed the ar_fork_recovery tests most of the time.

The commit "Fetch recall block later when receiving new block" is modified and the following commits are new.

PR #87, opened 2018-11-21
Feature/bulk upload files

ldmberman:


src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,82 @@
...
+		app_queue:add(Queue, chunk_to_tx(Hash, Blob));
+	true ->	
+		<< Chunk:?MB/binary, Rest/binary >> = Blob,
+		app_queue:add(Queue, chunk_to_tx(Hash, Chunk)),
+		upload_blob(Queue, Hash, Rest)
+	end.
+
+%% @doc Converts the given binary chunk into a transaction. A hash of the whole block the chunk
+%% is part of is assigned as a tag.
+chunk_to_tx(Hash, Chunk) ->
```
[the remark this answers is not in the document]
ldmberman: Good point!

ldmberman:
@samcamwilliams I clarified 1048576 = 1024 * 1024 in a comment; see the commit description [here](https://github.com/ArweaveTeam/arweave/pull/87/commits/e1133984890c5a77cce517cef53fe8242e50733b).

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+-export([upload/2, upload/3, download/2]).
+
+-include("ar.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+%%% Starts a server that accepts possibly huge files, splits them into small chunks, and
+%%% submits them to Arweave as separate transactions. Transactions are submitted via
+%%% the queue module.
+
+-define(MB, 1048576). % 1024 * 1024
```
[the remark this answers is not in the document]
ldmberman: @samcamwilliams has already suggested this. I tried it but did not like the experience - check [this commit description](https://github.com/ArweaveTeam/arweave/pull/87/commits/f4d50461bd26f7bcf53b668b6d695bee844f7907).

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+				{"blob_hash", ar_util:encode(Hash)},
+				{"number_of_chunks", integer_to_binary(ChunkNumber)},
+				{"chunk_position", integer_to_binary(ChunkPosition)}
+			],
+		data = Chunk
+	}.
+
+%% @doc Searches the local storage for the chunks of the blob with the given hash.
+%% If the blob is reconstructed successfuly, writes it to the specified destination.
+%% The provided hash has to be a Base 64 encoded SHA 256 hash of the file as a binary string.
```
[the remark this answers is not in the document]
ldmberman: I would then go a little further and accept a list of tags to assign to every tx here, not `Hash`. So I would pass the `blob_base64_sha256` tag along. If we want to attach other hashes in the future, we will add more tags.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+	Queue = app_queue:start(Node, Wallet),
+	Hash = crypto:hash(?BLOB_HASH_ALGO, Blob),
+	upload_blob(Queue, Hash, Blob),
+	Queue.
+
+%% @doc Takes a binary blob and processes it chunk by chunk. Each chunk is converted into
+%% a transaction and put into the queue. Chunk size is 1MB.
+upload_blob(Queue, Hash, Blob) ->
+	BlobSize = byte_size(Blob),
+	ChunkNumber = BlobSize div ?MB + if BlobSize rem ?MB =/= 0 -> 1; true -> 0 end,
```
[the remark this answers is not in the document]
ldmberman: Yes, I will also update [the style guide](https://github.com/ArweaveTeam/arweave/pull/87) accordingly.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+
+%% @doc Starts a queue server, splits the given file into chunks, wraps the chunks as 
+%% Arweave transactions, and submits them to the queue. Returns the queue PID.
+upload(Wallet, Filename) when is_list(Filename) ->
+	{ok, Filecontents} = file:read_file(Filename),
+	upload(Wallet, Filecontents);
+upload(Wallet, Blob) when is_binary(Blob) ->
+	upload(whereis(http_entrypoint_node), Wallet, Blob).
+
+upload(Node, Wallet, Blob) ->
```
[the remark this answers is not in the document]
ldmberman: We can not stop the queue here - the queue does not notify when all chunks (transactions) are posted.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+%% @doc Converts the given binary chunk into a transaction. A hash of the whole block the chunk
+%% is part of is assigned as a tag.
+chunk_to_tx(Hash, Chunk, ChunkNumber, ChunkPosition) ->
+	#tx {
+		tags =
+			[
+				{"app_name", "BulkUpload"},
+				{"blob_hash", ar_util:encode(Hash)},
+				{"number_of_chunks", integer_to_binary(ChunkNumber)},
+				{"chunk_position", integer_to_binary(ChunkPosition)}
```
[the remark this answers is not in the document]
ldmberman: The term "number" does not itself suggest the ordering. May be `chunk_sequence_number`?

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+
+%% @doc Searches the local storage for the chunks of the blob with the given hash.
+%% If the blob is reconstructed successfuly, writes it to the specified destination.
+%% The provided hash has to be a Base 64 encoded SHA 256 hash of the file as a binary string.
+download(Hash, Filename) ->
+	{ok, Blob} = download(Hash),
+	file:write_file(Filename, Blob, [{encoding, unicode}]).
+
+download(Hash) ->
+	app_search:get_entries(<< "blob_hash" >>, Hash),
```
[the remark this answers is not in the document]
ldmberman: It is not actually "chunked" though if it only consists of a single chunk.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+	receive TXIDs ->
```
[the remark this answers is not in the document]
ldmberman: Yes, I would be happy to include this improvement here.

src/ar_node_worker.erl
```diff
@@ -236,32 +239,13 @@ handle_gossip(SPid, {NewGS, UnhandledMsg}) ->
...
-				{potential_txs, ar_util:unique([TX | PotentialTXs])}
-			]}
-	end.
-
-add_tx(StateIn, TX, NewGS) ->
+add_tx(StateIn, TX, GS) ->
 	#{node := Node, waiting_txs := WaitingTXs, potential_txs := PotentialTXs} = StateIn,
 	case ar_node_utils:get_conflicting_txs(aggregate_txs(StateIn), TX) of
 		[] ->
+			{NewGS, _} = ar_gossip:send(GS, {add_tx, TX}),
```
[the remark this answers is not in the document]
ldmberman: The function adds a tx to the pool. If it is not in conflict with some other transactions, makes sense to gossip it but it only makes sense to include it into a block after it got some time to be propogated across the network.

Which comment are you referring to?

src/apps/app_bulk_upload.erl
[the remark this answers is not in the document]
martin-torhage: I don't know exactly what "may lead to unexpected results embedded into other arithmetic expressions" refers to, but I'm thinking wrapping it in a parentesis might be a work-around for those issues. E.g. `-define(MB, (1024 * 1024)).`

src/apps/app_bulk_upload.erl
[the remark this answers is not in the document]
ldmberman: [Done.](https://github.com/ArweaveTeam/arweave/pull/100/commits/a97f51c7cb26d4307a9a2bf44f87a05530b4975a)

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+		end,
+		Transactions
+		),
+		{ok, << (TX#tx.data) || TX <- SortedTransactions >>}
+	end.
+
+
+upload_test_() ->
+	{timeout, 60, fun() ->
+		app_search:deleteDB(),
```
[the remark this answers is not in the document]
martin-torhage: Ah, so you mean that the test will fail the second time since the `blob_hash` will collide. Wait a minute.... That will happen in reality too... I will add more comments about this.

src/apps/app_bulk_upload.erl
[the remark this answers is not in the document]
ldmberman: As for the variable names, yes - they do not have to carry all these details around.

As a tag name though, it may help people browsing the weave to make sense of transactions.

src/apps/app_bulk_upload.erl
martin-torhage: I can mess up your upload by storing a transaction with the same `blob_hash`. I guess this is something we should discuss.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+			{ok, Blob}
+		end
+	end.
+
+reconstruct_blob([]) -> {error, not_found};
+reconstruct_blob(Transactions) ->
+	Head = hd(Transactions),
+	{_, ChunkNumberBinary} = lists:keyfind(<< "number_of_chunks" >>, 1, Head#tx.tags),
+	ChunkNumber = binary_to_integer(ChunkNumberBinary),
+	if length(Transactions) < ChunkNumber ->
```
martin-torhage: If the same file is uploaded twice (by the same user or by different users) this will still pass. Although, it's not enough to change it to `if length(Transactions) /= ChunkNumber ->`.

src/apps/app_bulk_upload.erl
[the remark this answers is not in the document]
ldmberman: Ah yes, the duplicate transactions may create a problem!

> I will add more comments about this.

Let's just discuss it in this thread for convenience.

src/apps/app_bulk_upload.erl
ldmberman: The same blob hash is fine, the current implementation won't work if there are multiple chunks with the same `chunk_number` though - those have to be filtered.

ldmberman: There are two cases:

1) There is an extra chunk with the same hash and chunk position.
2) There is an extra chunk with the same hash but with incorrect data or tags.

`1)` is not a problem - I just need to make it ignore extra chunks.

`2)` is harded to address completely. Finding chunks by hash is convenient and includes the integrity check, what is good enough for now IMO.

src/apps/app_bulk_upload.erl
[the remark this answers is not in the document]
martin-torhage: `chunk_sequence_number` sounds good to me 👍👍

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,208 @@
...
+			],
+		data = Chunk
+	}.
+
+%% @doc Searches the local storage for the chunks of the blob with the given hash.
+%% If the blob is reconstructed successfuly, writes it to the specified destination.
+%% The provided hash has to be a Base 64 encoded SHA 256 hash of the file as a binary string.
+download(Hash, Filename) ->
+	{ok, Blob} = download(Hash),
+	file:write_file(Filename, iolist_to_binary(Blob), [write]).
```
martin-torhage: `file:write_file/2` can handle iolists, so no need to call `iolist_to_binary/1` here.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,208 @@
...
+
+download(Hash) ->
+	TXIDs = app_search:get_entries(<< "blob_hash" >>, Hash),
+	Transactions = lists:map(fun(TX) -> ar_storage:read_tx(TX) end, TXIDs),
+	case lists:member(unavailable, Transactions) of
+		true ->
+			{error, invalid_storage_state};
```
martin-torhage: This is to me a confusing error message. What do you feel about replacing it with `unavailable_transaction`?

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,208 @@
...
+		false ->
+			{ok, Blob} = reconstruct_blob(Transactions),
+			BlobHash = ar_util:encode(crypto:hash(?BLOB_HASH_ALGO, iolist_to_binary(Blob))),
+			case BlobHash == Hash of
```
martin-torhage: This would be more clear, or what do you think?
```
			case Hash of
				BlobHash ->
					{ok, Blob};
				_ ->
					{error, invalid_upload}
			end
```

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,208 @@
...
+reconstruct_blob(Transactions) ->
+	Head = hd(Transactions),
+	{_, NumberOfChunksBinary} = lists:keyfind(<< "number_of_chunks" >>, 1, Head#tx.tags),
+	NumberOfChunks = binary_to_integer(NumberOfChunksBinary),
+	case length(Transactions) < NumberOfChunks of
+		true ->
+			{error, missing_chunks};
+		false ->
+			Sorted = sort_transactions_by_chunk_position(Transactions),
+			SortedUnique = lists:foldr(fun(V, A) -> drop_duplicates(V, A) end, [], Sorted),
```
martin-torhage: You can also write this like this:
`lists:foldr(fun drop_duplicates/2, [], Sorted)`

martin-torhage: This handles the case where e.g. the same user uploads the same file twice. But, a malicious user can still mess upp a BulkUpload by adding invalid transactions. Just a reminder, I think we still need to discuss this.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,173 @@
...
+			SecondChunkPosition = binary_to_integer(SecondPositionBinary),
+			FirstChunkPosition =< SecondChunkPosition
+		end,
+		Transactions
+		),
+		{ok, << (TX#tx.data) || TX <- SortedTransactions >>}
+	end.
+
+
+upload_test_() ->
```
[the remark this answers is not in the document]
martin-torhage: Different code paths. If we reuse the code, it's just a matter of generating a new blob with one more byte and see that it works. I can do it if you like.

src/apps/app_bulk_upload.erl
ldmberman: Check my message [here](https://github.com/ArweaveTeam/arweave/pull/87#discussion_r238360915).

src/apps/app_bulk_upload.erl
ldmberman: I think, a better idea would be to simply filter `unavailable`s out. It none is left, it will end up returning `missing_chunks`, what makes sense.

src/apps/app_search.erl
```diff
@@ -43,7 +43,12 @@ add_entry(PID, Name, Value, ID) ->
 
 get_entries(Name, Value) -> get_entries(whereis(http_search_node), Name, Value).
 get_entries(PID, Name, Value) ->
-	PID ! {get_tx, Name, Value, self()}.
+	PID ! {get_tx, Name, Value, self()},
+	receive TXIDs ->
+		TXIDs
+	after 1000 ->
```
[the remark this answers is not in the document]
ldmberman: Ok, it would be safer to default to 3000ms here for now.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+
+upload_blob(Node, Wallet, Blob) ->
+	Queue = app_queue:start(Node, Wallet),
+	upload_chunks(Queue, Blob),
+	Queue.
+
+%% @doc Takes a binary blob and processes it chunk by chunk. Each chunk is converted into
+%% a transaction and put into the queue. Chunk size is 1MB.
+upload_chunks(Queue, Blob) ->
+	upload_chunks(Queue, Blob, true).
```
martin-torhage: `true` and `false` are very inexplicit names for atoms. Let's use something more clear like `first` `not_first`. But, even better than having `first` and `not_first`, let's use a counter starting on 1. What do you think about this?

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+
+%% @doc Searches the local storage for the chunks of the blob identified by the given
+%% Base 64 encoded transaction ID. If the blob is reconstructed successfully,
+%% writes it to the specified destination.
+download(TXID, Filename) ->
+	{ok, Blob} = download(TXID),
+	file:write_file(Filename, Blob, [write]).
+
+download(TXID) ->
+	case download_chunks(TXID, []) of
```
martin-torhage: I'm suggestion pattern matching instead of checking type. I.e. let `download_chunks/2` return `{ok, Chunks}` or `{error, Error}`

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+	end.
+
+download_chunks(TXID, Chunks) ->
+	case ar_storage:read_tx(ar_util:decode(TXID)) of
+		unavailable ->
+			tx_not_found;
+		TX ->
+			case lists:keyfind(<< "app_name" >>, 1, TX#tx.tags) of
+				false ->
+					invalid_tx;
```
martin-torhage: How about `non_bulk_upload_tx`?

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+		Err ->
+			{error, Err}
+	end.
+
+download_chunks(TXID, Chunks) ->
+	case ar_storage:read_tx(ar_util:decode(TXID)) of
+		unavailable ->
+			tx_not_found;
+		TX ->
+			case lists:keyfind(<< "app_name" >>, 1, TX#tx.tags) of
```
martin-torhage: I guess what you mean is this:

```
case lists:keyfind(<< "app_name" >>, 1, TX#tx.tags) of
    {<< "app_name" >>, << "BulkUpload" >>} -> yeay;
    _ -> bohoo
end
```

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+				false ->
+					invalid_tx;
+				_ ->
+					case lists:keyfind(<< "first_chunk" >>, 1, TX#tx.tags) of
```
martin-torhage: See my comment for the parent case.

Btw, we don't like nested cases. How about casing on a tuple with both instead? Like: `{AppName, FirstChunk}`

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+upload_three_chunks_test_() ->
+	{
+		timeout, 60, fun() ->
+			upload_blob_test_with_blob_size(?CHUNK_SIZE * 2 + 1)
+		end
+	}.
+
+upload_blob_test_with_blob_size(BlobSize) ->
+	Wallet = {_, Pub} = ar_wallet:new(),
+	Bs = ar_weave:init([{ar_wallet:to_address(Pub), ?AR(10000), <<>>}]),
```
martin-torhage: I would rather not reset the weave on every test. I see no point in doing so and the tests would be better if it's not running in an empty weave.

src/apps/app_bulk_upload.erl
ldmberman: It does not go well with recursion in this case.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+	Node = ar_node:start([], Bs),
+
```
martin-torhage: We never talked about empty lines. I'm suggesting never allow them and add it to our style guide. What do you think about this?

src/apps/app_bulk_upload.erl
ldmberman: There is a much bigger problem here - we need a dedicated database setup for tests.. We need to devise a generic solution for it outside of this PR.

src/apps/app_bulk_upload.erl
ldmberman: For me they make reading easier, but I am fine with not having them if there is a reason. Why do you want to remove them?

src/apps/app_bulk_upload.erl
ldmberman: `"first_chunk"`'s value is not important so I do not check for it here. And yes, I will get rid of the nested case.

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+hash(Blob) ->
+	ar_util:encode(crypto:hash(?BLOB_HASH_ALGO, Blob)).
+
+%% Generates an iolist() of the given size.
+%% The first 100 bytes are randomly picked from [1; 100].
+%% The rest is filled with 1. It would take too long to randomly pick every byte for a huge blob.
+generate_blob(Size) ->
+	generate_blob(Size, []).
+
+generate_blob(0, Acc) ->
```
martin-torhage: I don't understand this function. How about just generate a binary with `Size` number of random bytes?

E.g (non-tested):

```
random_blob(Size) ->
    random_blob(Size, []).

random_blob(Size, Acc) when length(Acc) == Size ->
    list_to_binary(Acc);
random_blob(Size, Acc) ->
    Byte = rand:uniform(255),
    random_blob([Byte | Acc]).
```

martin-torhage:
I just finished reviewing and adding all the feedback in regular inline Github comments instead of a Github review.

I'm not sure about relying on `tx.last_tx` for two reasons:
* There is a slim chance it's going away soon. I don't think so, but I would like to verify this.
* This blocks the possibility of adding any type of transaction by the same wallet in between. E.g. doing a web page archive, doing another bulk upload, transferring tokens (from the wallet).

src/apps/app_bulk_upload.erl
```diff
@@ -0,0 +1,212 @@
...
+		tags = chunk_tags(IsFirstChunk),
+		data = Chunk
+	}.
+
+chunk_tags(false) ->
+	[
+		{<< "app_name" >>, << "BulkUpload" >>}
+	];
+chunk_tags(true) ->
+	chunk_tags(false) ++ [{<< "first_chunk" >>, << "true" >>}].
```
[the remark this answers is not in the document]
ldmberman: - [There was a comment already](https://github.com/ArweaveTeam/arweave/pull/87#discussion_r239106393).
- Check [the commit](https://github.com/ArweaveTeam/arweave/pull/87/commits/b09f204a54fd764138a197aa74426091524677b7).

src/apps/app_bulk_upload.erl
martin-torhage: I think it does or I'm missing your point. Just make sure `download_chunks/2` always return the tuple.

src/apps/app_bulk_upload.erl
martin-torhage: It's not needed in Erlang. If the code is not easily read, blank lines are just a work-around. So let's try to make the code more readable instead. Usually the solution is to break out unrelated stuff into separate functions. But let's not change the style-guide now then.

ldmberman:
Yes, those are fair points. However, we only know transaction ID at the time we include into the block so we need to maintain another queue and keep it in sync with `app_queue` to tag every transaction properly. I considered it at first but the complexity is much higher compared to the current approach.

src/apps/app_bulk_upload.erl
ldmberman: Ah, yes, nevermind.

src/apps/app_bulk_upload.erl
martin-torhage: Oh yes, that's 2 MB of random data which takes a while. But actually, the blobs doesn't have to be random in the test since the size of the three blobs are different. It would also be good to test an uploading and downloading the same binary twice.

src/apps/app_bulk_upload.erl
ldmberman: It does not answer my question. What is the argument against blank lines?

src/apps/app_bulk_upload.erl
ldmberman: The blobs have to be random cause there is no cleanup in between test runs.

PR #101, opened 2018-11-30
Maintenance/refactor http iface

llaisdy:
Move all code from `ar_http_iface.erl` into _client, _server and _tests modules.

Remove `ar_http_iface.erl`.

No refactoring other than these file movements.

All 137 tests passing


src/ar.erl
```diff
@@ -328,7 +328,7 @@ start(
 	lists:foreach(fun(Feature) -> ar_meta_db:put(Feature, true) end, Enable),
 	lists:foreach(fun(Feature) -> ar_meta_db:put(Feature, false) end, Disable),
 	% Add self to all remote nodes.
-	%lists:foreach(fun ar_http_iface:add_peer/1, Peers),
```
ldmberman: I would clean these comments up now.

src/ar_gossip.erl
```diff
@@ -75,7 +75,7 @@ possibly_send(_S, Peer, #gs_msg { data = {new_block, _Node, _Height, _NewB, _Rec
 		when (not is_pid(Peer)) ->
 	ignore;
 possibly_send(_S, Peer, #gs_msg { data = {add_tx, _TX} }) when not is_pid(Peer) ->
-	%ar_http_iface:send_new_tx(Peer, TX);
+	%ar_http_iface_client:send_new_tx(Peer, TX);
```
ldmberman: I would rather remove this comment.

src/ar_http_iface_tests.erl
```diff
@@ -0,0 +1,830 @@
+%%%
+%%% @doc Tests for ar_http_ifcae_* modules.
```
ldmberman: There is a typo.

src/ar_http_iface_tests.erl
```diff
@@ -0,0 +1,830 @@
+%%%
```
ldmberman: FYI: I have just added a commit to the [style guide PR](https://github.com/ArweaveTeam/arweave/pull/100) recommending to put tests into separate files.

llaisdy: Good idea.  A few others here: https://github.com/ArweaveTeam/arweave-private/issues/8

src/ar_http_iface_server.erl
```diff
@@ -0,0 +1,957 @@
...
+%% @doc Return the balance of the wallet specified via wallet_address.
+%% GET request to endpoint /wallet/{wallet_address}/balance
+handle('GET', [<<"wallet">>, Addr, <<"balance">>], _Req) ->
+	case safe_decode(Addr) of
+		{error, invalid} ->
+			{400, [], <<"Invalid address.">>};
+		{ok, AddrOK} ->
+			case ar_node:get_balance(whereis(http_entrypoint_node), AddrOK) of
+				node_unavailable ->
+					{504, [], <<"Temporary timeout.">>};
```
martin-torhage: `ar_node:get_balance/2` returns `node_unavailable` when there is a internal timeout. First of all, I don't see why there is a timeout at all. Elli will timeout anyway if this is too slow, so I wish `ar_node:get_balance/2` waited infinitely instead.

504 is a gateway timeout, so it doesn't sound like a good fit for this response (since it's an internal timeout). A "503 Service Unavailable" makes sense to me to though. What do you think about a 503 instead?

llaisdy: AFAICS `ar_http_iface:verify_request_to_blockshadow/1` is never used.  I removed the function as well as the export.

llaisdy: 503 seems fine.

src/ar_http_iface_server.erl
```diff
@@ -0,0 +1,957 @@
...
+%% @doc Return the balance of the wallet specified via wallet_address.
+%% GET request to endpoint /wallet/{wallet_address}/balance
+handle('GET', [<<"wallet">>, Addr, <<"balance">>], _Req) ->
+	case safe_decode(Addr) of
+		{error, invalid} ->
+			{400, [], <<"Invalid address.">>};
+		{ok, AddrOK} ->
+			case ar_node:get_balance(whereis(http_entrypoint_node), AddrOK) of
+				node_unavailable ->
+					{503, [], <<"Temporary timeout.">>};
```
martin-torhage: Is it good or bad to change the text description of the status codes? I don't know. It's "supposed" to be "Service Unavailable"

llaisdy: That's not the text description of the status code it's the response body. Response will be received with status {503, "Service Unavailable"}.

llaisdy: cf https://github.com/ArweaveTeam/arweave/blob/maintenance/refactor-http-iface/src/ar_http_iface_tests.erl#L53-L65

PR #106, opened 2018-12-19
mute-on-success broken on linux

martin-torhage:


martin-torhage:
@ldmberman We got false errors in the log, so I pushed one more commit. Please have a look.

ldmberman:
@martin-torhage looks good, thanks!

src/ar_node_worker.erl
```diff
@@ -346,7 +346,7 @@ process_new_block(#{ height := Height } = StateIn, NewGS, NewB, unavailable, Pee
 			StateNext = StateIn#{ gossip => NewGS },
 			process_new_block(StateNext, NewGS, NewB, RecallShadow, Peer, HashList);
 		false ->
-			ar:err(failed_to_get_recall_block),
+			ar:info(failed_to_get_recall_block),
```
ldmberman: I think, we should report an error here - a failure to find a recall block should not happen under normal circumstances; also, the weave would stop growing on this node.

src/ar_fork_recovery.erl
```diff
@@ -302,7 +302,7 @@ do_fork_recover(S = #state {
 					)
 				of
 					false ->
-						ar:err(
+						ar:report_console(
```
ldmberman: Why not `ar:info` (here and in the following one)?

src/ar_node_worker.erl
martin-torhage: If it will stall, then I agree with reporting an error here, but to me it doesn't look like it will stall. I think it will continue growing when next block is received.

Please correct me if I'm wrong.

src/ar_fork_recovery.erl
martin-torhage: I'm thinking it might make sense to do a `ar:console/1` call here, but since `ar:console/1 doesn't format this message nicely, it's better to leave it as `ar:report_console/1` so we can deal with it later.

If you don't think it makes sense to log it to the console, then we can change it to `ar:info/1`.

src/ar_node_worker.erl
ldmberman: > I think it will continue growing when next block is received.

Is it because it can do a fork recovery when the next block arrives?

Yeah, we can use a warning then.

martin-torhage: Btw, before we created `ar:info`, `ar:err` etc, it was a `ar:report`, so `ar:info` is keeping the old behaviour.

src/ar_fork_recovery.erl
martin-torhage: I changed to `ar:info` for "the following one" (applied_fork_recovery_block)

ldmberman: > If you don't think it makes sense to log it to the console

It's not logged to the console anyway, because of a macro, unless I am missing smth. Under the hood `ar:info` would do the same thing as `report_console`.

src/ar_node_worker.erl
martin-torhage: > > I think it will continue growing when next block is received.
> 
> Is it because it can do a fork recovery when the next block arrives?
> 
> Yeah, we can use a warning then.

I assume so, yes. But I'm not sure. Keeping the old behaviour (i.e. using `ar:info`) is at least not wrong.

ldmberman: If miners' nodes can be stuck without them noticing it, it's a good enough reason for changing the old behaviour. But I am pretty confident fork recovery can bring the node back in this situation, so don't stricly mind keeping info here.

PR #108, opened 2019-01-07
Get TX status endpoint

martin-torhage:


src/apps/app_search.erl
```diff
@@ -41,6 +41,22 @@ add_entry(undefined, _, _, _) -> do_nothing;
 add_entry(PID, Name, Value, ID) ->
 	PID ! {add_tx, Name, Value, ID}.
 
+get_tags_by_id(TXID) ->
+	get_tags_by_id(whereis(http_search_node), TXID).
+get_tags_by_id(PID, TXID) ->
```
martin-torhage: I think we agreed on, but I cannot find it in the code standard, that functions should have an empty line in between them even if they have the same name.

martin-torhage: In the Erlang docs, pid-variables are called `Pid` and not `PID`. I guess we should do the same.

src/apps/app_search.erl
```diff
@@ -41,6 +41,22 @@ add_entry(undefined, _, _, _) -> do_nothing;
...
+	PID ! {get_tags, TXID, self()},
+	receive Tags ->
```
martin-torhage: Receiving anything and assuming it's what you expect is dangerous. Wrap the message and receive it like this instead:
`receive {tags, Tags} ->`

src/apps/app_search.erl
```diff
@@ -41,6 +41,22 @@ add_entry(undefined, _, _, _) -> do_nothing;
...
+		lists:map(
+			fun(Tag) ->
+				{_, Name, Value, _} = Tag,
```
martin-torhage: I don't know what the first and the forth element is here, but I'm guessing it would make more sense to do this transformation in the other process, where there is a context about where this tuple is from.

src/ar_http_iface_server.erl
```diff
@@ -638,6 +665,19 @@ handle(Method, [<<"height">>], _Req) when (Method == 'GET') or (Method == 'HEAD'
...
 	{400, [], <<"Request type not found.">>}.
 
+handle_get_tx(Hash) ->
+	case hash_to_maybe_filename(tx, Hash) of
+		{error, ID, unavailable} ->
+			case is_a_pending_tx(ID) of
+				true ->
+					{error, pending};
+				false ->
+					ar_tx_db:get(ID)
```
martin-torhage: `ar_tx_db:get/1` can return `not_found` but it looks like this function expect it to return `{error, not_found}`.

src/ar_http_iface_tests.erl
```diff
@@ -736,6 +736,28 @@ get_txs_by_send_recv_test_() ->
...
+	receive after 1000 -> ok end,
+	{ok, {_, _, Body, _, _}} =
+		ar_httpc:request(
+			<<"GET">>,
+			{127, 0, 0, 1, 1984},
+			"/tx/" ++ binary_to_list(ar_util:encode(TX#tx.id)) ++ "/status"
+		),
+	{Res} = ar_serialize:dejsonify(Body),
+	HashList = ar_node:get_hash_list(hd(Peers)),
+	?assertEqual(true, lists:member({<<"block_height">>, length(HashList) - 1}, Res)),
```
martin-torhage: Matching bools like this won't give much usable output when the test fails.

How about do this instead:
`?assertEqual({<<"block_height">>, length(HashList) - 1}, lists:keyfind(<<"block_height">>, 1, Res)),`

Or:
```
ResMap = maps:from_list(Res),
?assertMatch(#{<<"block_height">> => length(HashList) - 1}, Res),
```

src/ar_http_iface_tests.erl
```diff
@@ -736,6 +736,28 @@ get_txs_by_send_recv_test_() ->
...
+	?assertEqual(true, lists:member({<<"block_indep_hash">>, ar_util:encode(hd(HashList))}, Res)),
+	?assertEqual(true, lists:member({<<"number_of_confirmations">>, 0}, Res)).
```
martin-torhage: If we let it mine a few more blocks, we could also verify that this value increased correctly. Either we mine before these asserts, or we add `ar_node:mine/1` calls after these asserts and do the request again.

src/ar_http_iface_server.erl
```diff
@@ -165,22 +165,49 @@ handle('GET', [<<"tx">>, <<"pending">>], _Req) ->
...
+			],
+			Tags = lists:filter(
+				fun(Tag) ->
+					{Name, _} = Tag,
+					lists:member(Name, TagsToInclude)
+				end,
+				app_search:get_tags_by_id(ar_util:decode(Hash))
+			),
+			TXHeight = proplists:get_value(<<"block_height">>, Tags),
+			io:format("TX: ~p ~p ~p~n", [Height, TXHeight, Hash]),
```
martin-torhage: This call should probably be removed.

src/ar_http_iface_server.erl
```diff
@@ -165,22 +165,49 @@ handle('GET', [<<"tx">>, <<"pending">>], _Req) ->
...
+				<<"block_indep_hash">>
+			],
+			Tags = lists:filter(
+				fun(Tag) ->
+					{Name, _} = Tag,
+					lists:member(Name, TagsToInclude)
+				end,
+				app_search:get_tags_by_id(ar_util:decode(Hash))
+			),
+			TXHeight = proplists:get_value(<<"block_height">>, Tags),
```
martin-torhage: I didn't found how the block height ended up here. Let's talk about it.

src/apps/app_search.erl
```diff
@@ -41,6 +41,22 @@ add_entry(undefined, _, _, _) -> do_nothing;
...
+				{Name, Value}
+			end,
+			Tags
+		)
+	after 3000 ->
+		[]
```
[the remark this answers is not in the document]
ldmberman: This function is not an HTTP handler, it can be also used elsewhere.

src/ar_http_iface_tests.erl
ldmberman: It actually outputs the evaluated expression, e.g.

```
{expression,"lists : member ( { << \"block_height\" >> , length ( HashList ) - 1 } , Res )"},
                                  {expected,false},
                                  {value,true}]}
```

martin-torhage: Sure, but you can see the expression by opening the source code anyway. What you cannot see in the source code is the value of `Res` and I think you will get it if doing the `?assertMatch` approach. What do you think?

I think we have too many tests that are hard to debug when they fail because they don't expose the incorrect values.

ldmberman: I agree, it makes sense to try out `assertMatch`.

ldmberman: `assertMatch` does not really allow to ignore part of the map.. But I replaced these assertions with the whole map assertions.

src/apps/app_search.erl
martin-torhage: ✅

martin-torhage: ✅

src/ar_http_iface_server.erl
```diff
@@ -638,6 +670,19 @@ handle(Method, [<<"height">>], _Req) when (Method == 'GET') or (Method == 'HEAD'
...
 	{400, [], <<"Request type not found.">>}.
 
+handle_get_tx(Hash) ->
+	case hash_to_maybe_filename(tx, Hash) of
+		{error, ID, unavailable} ->
+			case is_a_pending_tx(ID) of
+				true ->
+					{error, pending};
+				false ->
+					{error, ar_tx_db:get(ID)}
```
martin-torhage: This always returns an error tuple.... :) I just pushed a fix for this.

src/apps/app_search.erl
```diff
@@ -193,7 +185,7 @@ search_by_id(TXID) ->
 
 basic_usage_test() ->
 	ar_storage:clear(),
-	SearchServer = start(),
+	{ok, SearchServer} = start(),
```
ldmberman: There is also a handful of `app_search:start` in `ar_http_iface_tests`.

src/apps/app_search.erl
```diff
@@ -190,25 +194,33 @@ search_by_id(TXID) ->
...
 	{ok, Tags} = get_tags_by_id(SearchServer, TX#tx.id, 3000),
 	?assert(lists:member({<<"TestName">>, <<"TestVal">>}, Tags)),
 	%% Check aux tags
-	?assert({<<"block_height">>, 1} == lists:keyfind(<<"block_height">>, 1, Tags)).
+	?assert({<<"block_height">>, 1} == lists:keyfind(<<"block_height">>, 1, Tags)),
+	%% Check that if writing the TX to the index again, the entries gets
+	%% overwritten, not duplicated.
+	AddTx(TX#tx{tags = [{<<"TestName">>, <<"Updated value">>}]}),
+	{ok, UpdatedTags} = get_tags_by_id(SearchServer, TX#tx.id, 3000),
+	?assertMatch({_, <<"Updated value">>}, lists:keyfind(<<"TestName">>, 1, UpdatedTags)).
```
ldmberman: I think, there should be another assertion of the absence of the old value in the list.

ldmberman:
We've made so many changes to `app_search` here - perhaps, it's the right time to move the module to the core (`ar_search`), where it should be.

src/apps/app_search.erl
ldmberman: ✅

src/apps/app_search.erl
ldmberman: ✅

PR #109, opened 2019-01-09
Added basic muckrock archiving implementation.

samcamwilliams:


src/apps/app_muckrock.erl
```diff
@@ -0,0 +1,54 @@
...
+
+submit_item(Queue, Item) ->
+    io:format("Submitting item ~s... ", [Item]),
+    {ok, {{_, 200, _}, _, Body}} = httpc:request(Item),
+    app_queue:add(Queue,
+        #tx {
+            tags =
+                [
+                    {"app_name", "MuckRock"},
+                    {"Original-File-Location", Item},
```
martin-torhage: I would suggest calling this one `Original-URL` instead.

src/apps/app_muckrock.erl
```diff
@@ -0,0 +1,54 @@
...
+-define(CHECK_TIME, 60 * 5).
+
+%%% Scrapes and uploads the latest responses received by the Muckrock FOIA
+%%% collection group.
+
+start(WalletFile) ->
+    ssl:start(),
+    Wallet = ar_wallet:load_keyfile(WalletFile),
+    Queue = app_queue:start(Wallet),
+    spawn(fun() -> server(Queue, []) end).
```
martin-torhage: I'm not so sure about how app_muckrock will be used, but wouldn't it makes sense to populate the `AlreadySeen` argument to `server/2` by using `app_search` to find all already uploaded PDFs?

src/apps/app_muckrock.erl
```diff
@@ -0,0 +1,54 @@
...
+                ],
+            data = list_to_binary(Body)
+        }
+    ),
+    io:format("Done!~n").
+
+get_latest() ->
+	{ok, {{_, 200, _}, _, Body}} =
+	    httpc:request("https://www.muckrock.com/api_v1/foia/?ordering=-id&status=done"),
+    case re:run(Body, "([^ \"]*?\.pdf)", [global, {capture, all_but_first, list}]) of
```
martin-torhage: This regex doesn't look correct and it extracts "relevant pdf" out of this part: "reattached the 3 relevant pdfs. The FY17 Tre"

src/apps/app_muckrock.erl
```diff
@@ -0,0 +1,54 @@
...
+	lists:foreach(
+		fun(Item) ->
+			try submit_item(Queue, Item)
+			catch _:_ -> ok end
+		end,
+		NewItems
+	),
+	receive
+		stop -> ok
+	after ?CHECK_TIME * 1000 ->
```
martin-torhage: Only waiting 1s between every run sounds very little. Or is the muckrock.com changing that often with new PDFs?

martin-torhage:
I accidentally closed the PR so I re-opened it again.

src/apps/app_muckrock.erl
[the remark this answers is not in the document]
samcamwilliams: Yep, this was done for speed. Let's see if MR like the first version then we know whether to spend time on improving it.

src/apps/app_muckrock.erl
samcamwilliams: +1

src/apps/app_muckrock.erl
samcamwilliams: Yes, that would be very nice.

PR #110, opened 2019-01-10
ar_tx_db overhaul

martin-torhage:


src/ar_tx.erl
```diff
@@ -78,7 +78,12 @@ sign(TX, PrivKey, PubKey) ->
...
 -ifdef(DEBUG).
 verify(#tx { signature = <<>> }, _, _) -> true;
-verify(TX, Diff, WalletList) ->
+verify(TX, Diff, WalletList) -> do_verify(TX, Diff, WalletList).
+-else.
+verify(TX, Diff, WalletList) -> do_verify(TX, Diff, WalletList).
+-endif.
+
+do_verify(TX, Diff, WalletList) ->
 	% ar:report(
```
ldmberman: May be let's remove this comment as well?

src/ar_tx_db.erl
```diff
@@ -22,33 +21,49 @@ start() ->
...
+	{Priv2, Pub2} = ar_wallet:new(),
+	[B0] = ar_weave:init([
+		{ar_wallet:to_address(Pub1), ?AR(10000), <<>>},
+		{ar_wallet:to_address(Pub2), ?AR(10000), <<>>}
+	]),
+	%% Test bad transaction
+	OrphanedTX1 = ar_tx:new(Pub1, ?AR(1), ?AR(5000), <<>>),
+	BadTX = OrphanedTX1#tx { owner = Pub1, signature = <<"BAD">> },
+	?assert(not ar_tx:verify(BadTX, 8, B0#block.wallet_list)),
 	timer:sleep(500),
```
ldmberman: Is `timer.sleep` here on purpose?

src/ar_tx_db.erl
```diff
@@ -22,33 +21,49 @@ start() ->
...
-    ["tx_too_cheap ","tx_fields_too_large ","tag_field_illegally_specified ","last_tx_not_valid "] = get(TX#tx.id),
-    ar_tx:verify(SignedTX, 8, B0#block.wallet_list).
+	Expected = {ok, ["same_owner_as_target", "tx_id_not_valid", "tx_signature_not_valid"]},
+	?assertEqual(Expected, get_error_codes(BadTX#tx.id)),
+	%% Test good transaction
+	OrphanedTX2 = ar_tx:new(Pub1, ?AR(1), ?AR(5000), <<>>),
+	SignedTX = ar_tx:sign(OrphanedTX2, Priv2, Pub2),
+	?assert(ar_tx:verify(SignedTX, 8, B0#block.wallet_list)),
+	clear_error_codes(BadTX#tx.id),
+	clear_error_codes(SignedTX#tx.id),
```
ldmberman: Makse sense to add an additional assertion to assert `clear_error_codes` actually cleans it up.

src/ar_key_db.erl
```diff
@@ -1,13 +1,12 @@
...
 -include_lib("eunit/include/eunit.hrl").
 -include("ar.hrl").
-%%% Defines a small in-memory metadata table for Arweave nodes.
-%%% Typically used to store small peices of globally useful information
-%%% (for example: the port number used by the node).
+%%% Database for storing decryption keys from encrypted recall blocks. The
+%%% entries has a TTL. The DB is a singleton.
 
-%% @doc Initialise the metadata storage service.
+%% @doc Create the DB. This will fail if a DB already exists.
```
ldmberman: Not a big deal, but it should be `a DB` the first time and `the DB` the second, same below.

src/ar_tx_db.erl
martin-torhage: It's an old sleep but it doesn't make sense, so I successfully removed it now :)

src/ar_tx_db.erl
martin-torhage: `ets:delete/2` always return `true`, so there is nothing to test.

src/ar_tx.erl
martin-torhage: ✅

src/ar_key_db.erl
martin-torhage: ✅

martin-torhage:
@ldmberman I handled all your feedback. Please have a look again and see my reply.

src/ar_tx_db.erl
ldmberman: You can call `get_error_codes` after `clear_error_codes` and observe the absence of some previously added entries.

martin-torhage: How about now?

PR #113, opened 2019-01-11
Add GET /wallet/{w}/txs(/{earliest_tx}) endpoint

ldmberman:
The endpoint returns identifiers of transactions made by the given wallet. Optionally stops when reaches the given transaction in the transaction chain.


samcamwilliams:
Looks good.

Maybe `get_wallet_txs` should not handle encoding to b64u though? Might be clearer if that is done in `handle_get_wallet_txs`.

Issue #105, opened 2018-12-19
Support multiple transactions per wallet per block


ldmberman:
Mempool should accept multiple transactions per wallet if they are ordered (i.e. they reference each other, and the first transaction references the last transaction of the wallet), their total value does not exceed the wallet balance, and each of them is valid when applied on top of the previous one.

Possibly, accept out of order transactions but kick them out after a short timeout. This way, transactions can survive out of order gossip but bad actors can then abuse mempool by posting transactions that won't be mined. Although the amount is limited by the wallet balance, one can constantly keep a certain portion of miners' memory filled up.

Optionally, accept transactions without a predetermined order. Then, out of order gossip does not introduce issues, and the mempool cannot be abused as described above.

- [x] Define `FORK_1_7` and make sure the new validation rules only apply above this height
- [x] Validate incoming transactions against the floating wallet list
- [x] Update the floating wallet list without waiting for the network propagation delay
- [x] Update the miner so that multiple txs per wallet are possibly chosen for the new block
- [x] Update`ar_node_worker:integrate_block_from_miner` so that the floating wallet list possibly preserves multiple txs per wallet
- [x] Update `ar_node_utils:integrate_new_block` so that the floating wallet list possibly preserves multiple txs per wallet

martin-torhage:
Since this is a hard fork, we need to define the fork height (like `FORK_1_6`) and keep the current behaviour if we haven't passed the fork height yet.

I haven't looked through the current code, but it sounds like `ar_node_worker:encounter_new_tx` is the place for dealing with new transactions and `ar_http_iface_server` does not.

ldmberman:
The feature branch - `feature/multiple-txs-per-wallet-per-block`.

JustMaier:
Got an estimated release date for this? This is kind of a showstopper for serious arweave integration.

ldmberman:
@JustMaier we expect to release it some time in the next few months.

JustMaier:
@ldmberman is there a workaround for this in the meantime?

ldmberman:
If you set up a node locally for development, you could try to use this branch to test the integration. On the mainnet perhaps it would help if you use multiple wallets?

JustMaier:
Any way we can help with development? Being limited to 1 tx per wallet per block severely limits the usability of the Arweave. The idea of having a database that can only take a write request from a user once every 3-5 minutes is pretty funny🤣

ldmberman:
[...]

PR #114, opened 2019-01-18
Feature/multiple txs per wallet per block

ldmberman:
Implements https://github.com/ArweaveTeam/arweave/issues/105.


src/ar_mine.erl
```diff
@@ -54,6 +54,31 @@ do_start(CurrentB, RecallB, RawTXs, RewardAddr, Tags, Diff, Parent) ->
...
+				{[], WalletList},
+				ar_node_utils:filter_all_out_of_order_txs(WalletList, TXs)
+			),
+			PickedTXs;
+		false ->
+			lists:filter(
+				fun(T) ->
+					ar_tx:verify(T, Diff, WalletList)
+				end,
+				TXs
```
martin-torhage: Shouldn't we use `ar_node_utils:filter_all_out_of_order_txs/2` here?

src/ar_http_iface_tests.erl
```diff
@@ -1185,6 +1221,75 @@ get_wallet_deposits_test_() ->
...
+			[],
+			ar_serialize:jsonify(ar_serialize:tx_to_json_struct(SignedTX2))
+		),
+	receive after 1000 -> ok end,
+	ar_node:mine(Node),
+	receive after 1000 -> ok end,
+	ReadTX = ar_storage:read_tx(SignedTX#tx.id),
+	?assertEqual(<<>>, ReadTX#tx.last_tx),
+	ReadTX2 = ar_storage:read_tx(SignedTX2#tx.id),
+	?assertEqual(SignedTX#tx.id, ReadTX2#tx.last_tx).
```
martin-torhage: Can we somehow more explicitly assert that the two transactions are in the block?

src/ar_http_iface_tests.erl
```diff
@@ -4,9 +4,45 @@
...
+			{"fail_external_tx_test", fun fail_external_tx_test/0},
+			{"add_tx_and_get_last_test", fun add_tx_and_get_last_test/0},
+			{"get_subfields_of_tx_test", fun get_subfields_of_tx_test/0},
+			{"get_pending_tx_test", fun get_pending_tx_test/0},
+			{"get_multiple_pending_txs_test_", get_multiple_pending_txs_test_()},
+			{"get_tx_by_tag_test", fun get_tx_by_tag_test/0},
+			{"get_tx_body_test", fun get_tx_body_test/0},
+			{"get_txs_by_send_recv_test_", get_txs_by_send_recv_test_()},
+			%% Tests that only run for after fork 1.7
+			{"can_post_multiple_txs_per_wallet", fun can_post_multiple_txs_per_wallet/0},
```
martin-torhage: How about replacing the two new tests with a multi-node test doing this:

1. Start both master and slave
2. Push two TXs to master
3. Mine on master
4. Verify new block is on slave and includes both TXs

src/ar_mine.erl
```diff
@@ -101,16 +126,7 @@ update_txs(
 	end,
 	%% Filter out invalid TXs. A TX can be valid by itself, but still invalid
```
martin-torhage: This comment should probably also have moved into `pick_txs_to_mine/4`

src/ar_mine.erl
```diff
@@ -54,6 +54,31 @@ do_start(CurrentB, RecallB, RawTXs, RewardAddr, Tags, Diff, Parent) ->
 		RawTXs
 	).
 
+pick_txs_to_mine(Height, Diff, WalletList, TXs) ->
```
martin-torhage: This function should probably be moved just after `calc_diff/2`

ldmberman:
Was closed by mistake.

llaisdy:
AFAICS, if a submitted TX.owner is not in the FloatingWalletList, the TX will always be rejected.  If that's right, you could save the node (and the reader;) a bit of work by catching that on the edge.  e.g., rewrite:

https://github.com/ArweaveTeam/arweave/blob/4bcfc81096a62419a4f20ad7edae07b557dda39d/src/ar_http_iface_middleware.erl#L856-L858

as something like:

```
	OwnerAddr = ar_wallet:to_address(TX#tx.owner),
	case lists:keyfind(OwnerAddr, 1, FloatingWalletList) of
		false -> handle_post_tx_verification_response();  %%  <-- new line
		{_, Balance, _} when (TX#tx.reward + TX#tx.quantity) > Balance ->
```

llaisdy:
Surely this can never be false?

https://github.com/ArweaveTeam/arweave/blob/4bcfc81096a62419a4f20ad7edae07b557dda39d/src/ar_tx_replay_pool.erl#L241

As the same test has already been done in `ar_tx:verify`.

ldmberman:
There is a height-dependent check in `ar_tx:verify`, which will ensure every pre-fork transaction has a valid `last_tx`, while new transactions do not need it there.

llaisdy:
Surely this can never be false: 

https://github.com/ArweaveTeam/arweave/blob/4bcfc81096a62419a4f20ad7edae07b557dda39d/src/ar_tx_replay_pool.erl#L241

Because the same test is done in ar_tx:verify (& the TX rejected if it fails).

llaisdy:
Sorry, missed that.

llaisdy:
The last clause of `ar_tx_replay_pool:verify_tx` (mempool_check) creates new mempool and FloatingWalletList and returns them in this tuple `{valid, NewFWL, NewMempool}`:

https://github.com/ArweaveTeam/arweave/blob/4bcfc81096a62419a4f20ad7edae07b557dda39d/src/ar_tx_replay_pool.erl#L265-L275

Are these new values used anywhere?  They are not used by `ar_http_iface_middleware:handle_post_tx2` which receives the tuple as `{valid, _, _}`.

What would happen if that `false` branch didn't do the work and just returned `valid` (& with `handle_post_tx2` updated)?

llaisdy:
Instead of

https://github.com/ArweaveTeam/arweave/blob/4bcfc81096a62419a4f20ad7edae07b557dda39d/src/ar_tx_replay_pool.erl#L277-L306

How about removing the duplicated inner case, e.g.:

```
verify_size(TX, #mempool { size_list = SizeList } = Mempool, BHL) ->
    TXSize = byte_size(TX#tx.data),
    {Anchor, Acc, Limit} =
        case lists:member(TX#tx.last_tx, BHL) of
            true  ->
                lists:keyfind(TX#tx.last_tx, 1, SizeList);
            false ->
                hd(SizeList)
        end,
    case Acc + TXSize of
        Sum when Sum > Limit ->
            invalid;
        _ ->
            NewSizeList = update_size_list(
                Anchor,
                TXSize,
                SizeList
            ),
            {valid, Mempool#mempool { size_list = NewSizeList }}
    end.
```

src/ar_tx_replay_pool.erl
```diff
@@ -0,0 +1,290 @@
...
+				{invalid, _} ->
+					false
+			end
+		end,
+		TXs
+	).
+
+%% PRIVATE
+
+create_state(RecentTXs) ->
```
maxmellen: Can we discuss this function together? I don't understand what `RecentTXs` is supposed to be...

src/ar_tx_replay_pool.erl
```diff
@@ -0,0 +1,290 @@
...
+	}.
+
+verify_tx(general_verification, TX, Diff, Height, FloatingWalletList, WeaveState, Mempool) ->
+	case ar_tx:verify(TX, Diff, Height, FloatingWalletList) of
+		true ->
+			verify_tx(last_tx_in_mempool, TX, Diff, Height, FloatingWalletList, WeaveState, Mempool);
+		false ->
+			{invalid, tx_verification_failed}
+	end;
+verify_tx(last_tx_in_mempool, TX, Diff, Height, FloatingWalletList, WeaveState, Mempool) ->
```
maxmellen: How about refactoring this clause to:
```erlang
verify_tx(last_tx_in_mempool, TX, Diff, Height, FloatingWalletList, WeaveState, Mempool) ->
	ShouldContinue = case ar_fork:height_1_8() of
		H when Height >= H ->
			%% Only verify after fork 1.8 otherwise it causes a soft fork
			%% since current nodes can accept blocks with a chain of last_tx
			%% references. The check would still fail on edge pre 1.8 since
			%% TX is validated against a previous blocks' wallet list then.
			case sets:is_element(TX#tx.last_tx, Mempool#mempool.tx_id_set) of
				true ->
					{invalid, last_tx_in_mempool};
				false ->
					continue
			end;
		_ ->
			continue
	end,
	case ShouldContinue of
		continue ->
			verify_tx(
				last_tx,
				TX,
				Diff,
				Height,
				FloatingWalletList,
				WeaveState,
				Mempool
			);
		{invalid, Reason} ->
			{invalid, Reason}
	end;

File: src/ar_tx_replay_pool.erl [hunk 8]
```diff
@@ -0,0 +1,290 @@
...
+				Diff,
+				Height,
+				FloatingWalletList,
+				WeaveState,
+				FloatingMempool
+			) of
+				{valid, NewFWL, NewMempool} ->
+					{VerifiedTXs ++ [TX], NewFWL, NewMempool};
+				{invalid, _} ->
+					{VerifiedTXs, FloatingWalletList, FloatingMempool}
```
maxmellen: Am I correct in understanding that this fold just moves forward in case of an invalid transaction rather than halting

src/ar_http_iface_middleware.erl
```diff
@@ -460,6 +460,22 @@ handle(<<"GET">>, [<<"wallet">>, Addr, <<"last_tx">>], Req) ->
...
+handle(<<"GET">>, [<<"tx_anchor">>], Req) ->
+	case ar_node:get_hash_list(whereis(http_entrypoint_node)) of
+		[] ->
+			{400, #{}, <<"The node has not joined the network yet.">>, Req};
+		BHL when is_list(BHL) ->
+			{
+				200,
+				#{},
+				ar_util:encode(
+					lists:nth(min(length(BHL), (?MAX_TX_ANCHOR_DEPTH)) div 2 + 1, BHL)
```
maxmellen: So in the end, we always pick the block exactly in the middle of the allowed range

src/ar_tx_replay_pool.erl
```diff
@@ -0,0 +1,290 @@
...
+
+pick_txs_to_mine_pre_1_8(Height, Diff, WalletList, TXs) ->
+	lists:filter(
+		fun(TX) ->
+			ar_tx:verify(TX, Diff, Height, WalletList)
+		end,
+		TXs
+	).
+
+pick_txs_under_size_limit(TXs) ->
```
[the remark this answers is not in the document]
ldmberman: It will not work in case the total size of `TXs` is under the limit.

ldmberman:
> AFAICS, if a submitted TX.owner is not in the FloatingWalletList, the TX will always be rejected. If that's right, you could save the node (and the reader;) a bit of work by catching that on the edge. 

Unfortunately it breaks a bunch of legacy tests which post specially crafted dummy transactions and expect them to pass the validation. We should rework those tests eventually, I just do not want to do it right now.

src/ar_fork_recovery.erl
```diff
@@ -20,11 +20,11 @@
 	recovery_hash_list, % Complete hash list for this fork
 	block_list, % List of hashes of verified blocks
 	hash_list, % List of block hashes needing to be verified and applied (lowest to highest)
-	recent_txs % List of {BH, TXIDs} pairs of the last N blocks
+	block_txs_pairs % List of {BH, TXIDs} pairs of the last ?MAX_TX_ANCHOR_DEPTH blocks
```
maxmellen: That's a lot more clear 😁

src/ar_tx_replay_pool.erl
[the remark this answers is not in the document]
maxmellen: Realised later on that my suggestion couldn't work like this 😅 It's fine to keep this as such for now

ldmberman:
[Merged](https://github.com/ArweaveTeam/arweave/commit/1f9c70c20da102bc31a0200a99ee12bef33b3062).

PR #115, opened 2019-01-21
Failing node tests

martin-torhage:


martin-torhage:
Actually, all code changes were done by me and Lev last week and we considered this branch except for some git polishing I just performed, so no more reviewing is necessary.

PR #121, opened 2019-02-13
/unsigned_tx requires too many fields

martin-torhage:
The commit `NOT FINISHED Fix invalid docs and tests for tx owner` should be handled before this branch can be considered done.


llaisdy:
All looks good.

PR #122, opened 2019-02-15
Fix mixed crashes

martin-torhage:


src/apps/app_net_crawler.erl
```diff
@@ -62,19 +62,27 @@ start_worker(Node) ->
...
-	fetch_peers(Node, 3).
+	fetch_peers(Node, {1, 4}).
 
-fetch_peers(_, 0) ->
+fetch_peers(_, {_Attempt, _MaxAttempts}) when _Attempt >= _MaxAttempts ->
 	unavailable;
-fetch_peers(Node, AttemptsLeft) ->
-	case ar_http_iface_client:get_peers(Node) of
+fetch_peers(Node, {Attempt, MaxAttempts}) when Attempt > 1 ->
+	timer:sleep(30 * 1000),
```
ldmberman: I do not know if and how we use `app_net_crawler`, but anyway wonder why we set such big timeouts.

PR #123, opened 2019-02-19
Refactor ar_meta_db as a gen_server

Raphexion:


samcamwilliams:
Hey @Raphexion ! Thanks for contributing. We will review this and get back to you.

All the best,
Sam

Raphexion:
I am sorry for such an invasive change. I just wanted to propose a gen_server variant of the ar_meta_db. I appreciate any feedback.

samcamwilliams:
We have been intending to move to a `gen_server`/more OTP-like style for a while, so this is very helpful -- not too invasive :).

All the best,
Sam

ldmberman:
Thank you for your contribution @Raphexion!

src/ar_meta_db.erl
```diff
@@ -1,91 +1,179 @@
...
+put(Key, Value) ->
+    gen_server:call(?MODULE, {put, Key, Value}).
+
+%% @doc Retreive value for key
 get(Key) ->
-	case ets:lookup(?MODULE, Key) of
-		[{Key, Obj}] -> Obj;
-		[] -> not_found
-	end.
+    gen_server:call(?MODULE, {get, Key}).
```
ldmberman: We use tabs for indentation.

src/ar_meta_db.erl
```diff
@@ -1,91 +1,179 @@
+%% @doc Defines a small in-memory metadata table for Arweave nodes.
+%% Typically used to store small peices of globally useful information
+%% (for example: the port number used by the node).
+%% @end
+
 -module(ar_meta_db).
--export([start/0, get/1, put/2, keys/0, remove_old/1,remove_old/2, increase/2]).
+-behaviour(gen_server).
+-define(SERVER, ?MODULE).
```
ldmberman: Why did you define the `SERVER` constant? You mostly use `?MODULE` anyways.

Raphexion: I don't have a good reference. I think it something I have picked up in a book. I often use it for for singletons. 

```gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).```

File: src/ar_meta_db.erl (hunk 1 above)
Comment (Raphexion) [reply src/ar_meta_db.erl]: I hopefully changed it now. If it is wrong please let me know.

File: src/ar_meta_db.erl [hunk 3]
```diff
@@ -1,91 +1,179 @@
...
+	gen_server:call(?MODULE, {get, Key}).
 
-%% @doc Increase the value associated by a key by Val
+%% @doc Increase key with value
 increase(Key, Val) ->
-	put(Key, get(Key) + Val).
+	gen_server:call(?MODULE, {increase, Key, Val}).
 
 %% @doc Remove entries from the performance database older than
-%% ?PEER_TMEOUT
```
ldmberman: Did you intend to remove this part of the docstring?

src/ar_meta_db.erl
ldmberman: Thank you.  Only `case`s still seem to have some extra indents.

src/ar_meta_db.erl
Raphexion: I think it is a common pattern. For example here:
https://github.com/stevenchan/minipush/blob/478046fc0962a13b6cb93aa8d6f6909e811271d4/src/supervisor/minipush_sup.erl

src/ar_meta_db.erl
Raphexion: Nope, good catch :+1:

src/ar_meta_db.erl
Raphexion: Is it better now? Sorry for the miss.

ldmberman: A consistent formatting would look like this:
```
...
        Res = case ets:lookup(?MODULE, Key) of
                [{Key, Obj}] -> Obj;
	        [] -> not_found
        end,
        {reply, Res, State};
```

Raphexion: I am sorry for this. I need to setup my emacs to work with tabs instead of spaces.

Raphexion: Are you sure that the end should be on the same level as Res? According to my "style" layout, it should be on the same level as the case. However, I am unsure what is correct.

ldmberman: I do not have a strong preference. We have [a styleguide](https://github.com/ArweaveTeam/arweave/blob/master/arweave_styleguide.md) but it does not define how to handle this. In other places we align the way I posted, if I remember correctly.

Raphexion: I think it has something to do that I assign Res using a case. To be honest, I don't what is best. I just follow emacs :joy:

Raphexion: I found an example in LearnYouSomeErlang book.

```
%% note, this one would be better as a pattern match in function heads!
%% I'm doing it this way for the sake of the example.
help_me(Animal) ->
    Talk = if Animal == cat  -> "meow";
              Animal == beef -> "mooo";
              Animal == dog  -> "bark";
              Animal == tree -> "bark";
              true -> "fgdadfgna"
           end,
    {Animal, "says " ++ Talk ++ "!"}.
```

ldmberman:
@Raphexion looks good to me! Inviting @martin-torhage to also take a look.

src/ar_meta_db.erl
Raphexion: I found a reference. *Erlang and OTP in Action* by Martin Logan pg 103.
>>>
"[Sets SERVER to module name] you may want to change that at some point, so you shouldn't assume that the server name will always remain the same as the module name"

martin-torhage:
Hi @Raphexion! Thank you for the contribution! We should really use gen_servers in many cases instead of rolling our own servers.

I realised today we don't have a guide for how to contribute, so I added one here:
https://github.com/ArweaveTeam/arweave/blob/master/CONTRIBUTING.md

If you try to run the tests (`make test`) they fail. I haven't looked into why they fail yet. One caveat you need to now (which I will add to CONTRIBUTING.md) is that the tests are running in a fully running node, so `ar_meta_db` is already started. Therefore, it's not possible to start and stop `ar_meta_db` in the setup and teardown functions.

Can you please have a look if you can get `make test` to work? If you run into any issues, then I would be happy to help. And I would also be happy to improve the CONTRIBUTING.md document if you have any feedback.

Good to know:
 - You need to enable `-define(DEBUG, debug).` for running the tests. See CONTRIBUTING.md
 - You can also start the node without connecting to the network with `make session` and run the tests from there. Either by `ar:tests().` or `ar:tests(ar_meta_db).`

Raphexion:
@martin-torhage Thank you for your feedback. I really like the CONTRIBUTING.md. It helps to get the correct flow. I felt I was doing a big change and I was unsure to fully test it. (In addition to eunit).

I will dig into this and come back with an updated pull-request.

Thank you again for the feedback :)

samcamwilliams:
Hey @Raphexion -- glad you like the guide!

Please can you also send us a mainnet AR address so that we can send you a small 'thank you' for contributing?

All the best,
Sam

Raphexion:
Wow, that would be great. However, I don't have a AR address yet. Is it okay you send it in the future. Hopefully I will contribute more. It is a great project, super friendly team and interesting Erlang code.

samcamwilliams:
No worries at all, send it along whenever you like. The easiest way to generate an address is using the web extension for Chrome, by the way: 
https://chrome.google.com/webstore/detail/arweave/iplppiggblloelhoglpmkmbinggcaaoc

Sam

Raphexion:
@samcamwilliams Thank you for the advice. The chrome extension worked like a charm. I really like the design of the system.

I think that my address is HOfKZ8SM6o2Cic0p4kdMWuT4ibRC9Vl3g3ryPwL31ZE
Because the keyfile I download had the form
```arweave-keyfile-HOfKZ8SM6o2Cic0p4kdMWuT4ibRC9Vl3g3ryPwL31ZE.json```

samcamwilliams:
@Raphexion  `500 AR` on its way to you in the next few days!

Sam

Raphexion:
@martin-torhage I've pushed an updated version. Pay especially attention to the move of the testcase. Now last in the list.

martin-torhage:
Hi @Raphexion,

I'm sorry for the delay. When doing some minor touch-ups I realized the old code didn't have a server. The ar_meta_db process is only there to own the ets table. You've changed the architecture by serializing all reads and writes through the gen_server process. I'm not sure how this is an improvement. How do you see it?

Raphexion:
Hi @martin-torhage,

No worries, I know that you guys are super hard at work :)

I see you point but how I see it, there is one key improvement with the code: _standadization_.

By using gen_server, the code becomes readable because it is familiar to most erlang developers. I see it as a *strength* that you say that it as simply a process and an ets. That means that this module is very easy to think and reason about. Before it wasn't as clear (I think). Moreover, by using the gen_server behaviour, it fits into the OTP framework. It works well with everything OTP (supervisors, hot reloading, start_link and so on).

There is no performance penalty by implementing the gen_server behaviour. It is just a "better" process.

That said, I will not hold it against it if you reject the pull-request. I am just happy to try to contribute. I hope to add more in the future :)

samcamwilliams:
Hey @Raphexion , I think @martin-torhage 's point is that because this patch flips the structure such that requests are received and processed directly by the `gen_server`, it serialises read requests to the table. Because the ETS table is setup with `read_concurrency` enabled, as it currently stands many processes could read from it directly at one time (no message passing involved), this patch would serialise the reads introducing a bottleneck.

We are totally on-board with standardising the codebase further though, and would like to make it more accessible to other Erlang programmers. You might well be right about the `remove_old` and `safe_fixtable` changes -- I will leave it to @martin-torhage to review.

All the best,
Sam

Raphexion:
Hey @samcamwilliams and @martin-torhage .

You guys are correct. I miss-understood Matin's code review above. Sorry about that Martin. I completely missed the point of the code; that we wanted to create and ets for fast access.

When we create the ets table, we create with public and {read_concurrency, true}. Which I missed to reflect upon.

Now I have updated my pull-request so that ```get(Key)``` does _not_ call ```gen_server:call```, but does a lockup directly. That way, we get original fast code for reads.

Would that be a good hybrid solution?

martin-torhage:
Hey @Raphexion, I just merged your commit together with a few improvements to master. I added a reference to this PR in the merge commit message. If you have any feedback, please add it here. Otherwise I'll close this PR.

In the future, maybe we should have contribution branch to create PRs to. Then we could actually click merge button so that GitHub wouldn't be less confused about what happened with the PR. What do you think about it?

Thank you for your contribution!

Raphexion:
Hey @martin-torhage,

I looked at your additional commits and your improvements. Thank you. Great!

Most project in github I work with (private or professionally) merge to master. But for example, eclipse/paho.mqtt.c takes in PRs on the branch `develop`. [paho.mqtt.c](https://github.com/eclipse/paho.mqtt.c/blob/master/CONTRIBUTING.md).

I think it would be wise that you guys have a develop branch like paho.
So you can have a little more control over master. 
Personally, I am happy to contribute and use what ever strategy what ever you guys decide :)

Thank you again for your reviews and time. You are welcome to close the PR.

martin-torhage:
Thanks for the feedback! I realised afterwards we can change which branch to merge the PR to, so it's ok to create the PR to master. Next time!

PR #124, opened 2019-02-19
Mention port forwarding in "blocks not received" warning

ldmberman:


martin-torhage:
I just realized I had added my feedback on Friday in a "review" which isn't submitted. Will re-add the feedback as ad-hoc comments.

src/ar_miner_log.erl
```diff
@@ -46,7 +46,8 @@ watchdog() ->
 	after ?FOREIGN_BLOCK_ALERT_TIME ->
 		log(
 			"WARNING: No foreign blocks received from network. "
-			"Please check your internet connection."
+			"Please check your internet connection and port forwarding settings."
+			"Consult your ISP or cloud provider to forward incoming HTTP requests with port 1984 to your machine."
```
martin-torhage: Hardcoding 1984 is not correct. The real values is in meta_db with the key `port`. E.g. `ar_meta_db:get(port)`.

martin-torhage: We also need to keep polling mode in mind.

My suggested message for polling mode:
```
WARNING: No foreign blocks received from network. Please check your internet connection and the logfile for errors.
```

My suggestion for non-polling mode:
```
WARNING: No foreign blocks received from network. Please check that your node is available on port 1234 on your Internet IP address. E.g. browse to http://[IP address]:1234/info. If so, please check the logfile for errors.
```

martin-torhage: Btw, I think we should implement a ping-back when joining that verifies the remote node can access the current node.

martin-torhage:
I did some changes. Feel free to squash + merge if you approve.

PR #127, opened 2019-03-11
Log missing txs when processing blocks

ldmberman:


martin-torhage:
Doesn't the block get rejected before this code is run since `ar_node_utils:validate/8` returns false? Did you really test this? I thought the issue was that `ar_node_utils:validate/8` produces an implicit log message, something about the "nonce" is not valid.

src/ar_node_worker.erl
```diff
@@ -445,6 +446,18 @@ process_new_block(#{ height := Height, cumulative_diff := CDiff } = StateIn, New
 			none
 	end.
 
+log_missing_txs(TXIDs, ReconstructedTXs) ->
+	case length(TXIDs) of
+		L when L > length(ReconstructedTXs) ->
+			MissingTXIDs = TXIDs -- lists:map(fun(TX) -> TX#tx.id end, ReconstructedTXs),
```
martin-torhage: We calculate the missing TXIDs here. Wouldn't it make sense to log if MissingTXIDs is not empty? Parent case doesn't make sense to me.

src/ar_node_worker.erl
```diff
@@ -445,6 +446,18 @@ process_new_block(#{ height := Height, cumulative_diff := CDiff } = StateIn, New
...
+			ar:info(
+				"Could not find these transactions from the new block: ~p",
```
martin-torhage: I think we should add the indep hash here as well. E.g.:

`ar:info([{transactions_missing_in_mempool_for_block, BH}, {missing_txs, TXs}]).`

src/ar_node_worker.erl
```diff
@@ -445,6 +446,18 @@ process_new_block(#{ height := Height, cumulative_diff := CDiff } = StateIn, New
 			none
 	end.
 
+log_missing_txs(TXIDs, ReconstructedTXs) ->
```
martin-torhage: How do you feel about renaming `ReconstructedTXs` to `TXsInMempool`?

martin-torhage:
I finished adding review feedback

src/ar_node_worker.erl
ldmberman: The parent case is just a quicker way to see if txs are missing. If yes, I subrtact the sets to log them.

martin-torhage: What if:
TXIDs: [a, b, c]
ReconstructedTXs: [b, c, d]

Or:
TXIDs: [a, b, c]
ReconstructedTXs: [e, f, g, h, i, j, k]

martin-torhage: Btw, I would suggest using `lists:member/2` together with `lists:all/2`. For larger lists it would make sense to use the `sets` module.

ldmberman: `TXIDs` can't be smaller by construction.

ldmberman: > Btw, I would suggest using lists:member/2 together with lists:all/2

What is your concern?

PR #129, opened 2019-03-14
Rewrite ar_meta_db (into gen_server)

martin-torhage:
This includes the commit from #123

### DIFF
--- a/src/ar_manage_peers.erl
+++ b/src/ar_manage_peers.erl
@@ -8,11 +8,10 @@
 %% @doc Print statistics about the current peers.
 stats() ->
 	Connected = ar_bridge:get_remote_peers(http_bridge_node),
-	All = all_peers(),
 	io:format("Connected peers, in preference order:~n"),
 	stats(Connected),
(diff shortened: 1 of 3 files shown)


src/ar_manage_peers.erl
```diff
@@ -43,7 +42,9 @@ format_stats(Peer, Perf) ->
 %% Peers who have behaved well in the past are favoured in ranking.
 %% New, unknown peers are given 100 blocks of grace.
 update(Peers) ->
-	ar_meta_db:remove_old(os:system_time(seconds)),
+	%% Remove performance records created more than a certain amount of time. It
```
ldmberman: A suggestion: "Remove performance records **not updated for** more than a certain amount of time. Also, I do not understand what "how actively updated it is" means in this context.

martin-torhage: It was a misunderstanding this function does something unexpected. That's not the case, so I removed the comment. Looks ok?

ldmberman: No text => no questions :)

Issue #25, opened 2018-07-26
Configuration of data directories


themue:
Today the directories for blocks, transactions, wallets, data, and logs are relative to project root. Other locations are possible by exchanging these manually by symlinks to other locations.

Configuration should be extended to allow individual configuration of these directories. If not configured location is relative to project root like today.

codybartfast:
This would have increased value for a native Windows version because of the lack of support for symlinks in Windows.

PR #130, opened 2019-03-15
Add support for custom weave data folder

ldmberman:
The motivation is to be able to run multiple independent miners from the same folder for testing purposes. As a bonus, the feature simplifies the task of copying data between machines (a common miners need).

### DIFF
--- a/arweave-server
+++ b/arweave-server
@@ -2,8 +2,6 @@
 
 set -o errexit
 
-mkdir -p ebin logs blocks hash_lists wallets wallet_lists txs blocks/enc priv data data/mnesia
-
 if [ -t 1 ]; then
     SHELL_OPTS=
 else
--- a/docker-arweave-server
+++ b/docker-arweave-server
@@ -2,7 +2,7 @@
 
 set -o errexit
 
-mkdir -p ebin logs blocks wallets txs blocks/enc priv data/genesis_txs
+mkdir -p ebin logs priv
 
 if [ -t 1 ]; then
 	SHELL_OPTS=
--- a/src/ar.erl
+++ b/src/ar.erl
@@ -69,6 +69,7 @@
 	init = false,
 	mine = false,
 	peers = [],
+	data_dir = ".",
 	polling = false,
 	auto_join = true,
 	clean = false,
@@ -110,6 +111,7 @@ main("") ->
 			{"mine", "Automatically start mining once the netwok has been joined."},
 			{"port", "The local port to use for mining. "
 						"This port must be accessible by remote peers."},
+			{"data_dir", "The directory for storing the weave and the wallets (when generated)."},
 			{"polling", "Poll peers for new blocks. Useful in environments where "
 						"port forwarding is not possible."},
 			{"clean", "Clear the block cache before starting."},
@@ -152,6 +154,8 @@ parse(["content_policy", F|Rest], O = #opts { content_policies = Fs }) ->
 	parse(Rest, O#opts { content_policies = [F|Fs] });
 parse(["port", Port|Rest], O) ->
 	parse(Rest, O#opts { port = list_to_integer(Port) });
+parse(["data_dir", DataDir|Rest], O) ->
+	parse(Rest, O#opts { data_dir = DataDir });
 parse(["diff", Diff|Rest], O) ->
 	parse(Rest, O#opts { diff = list_to_integer(Diff) });
 parse(["polling"|Rest], O) ->
@@ -204,6 +208,7 @@ start(#opts { benchmark = true, max_miners = MaxMiners }) ->
 start(
 	#opts {
 		port = Port,
+		data_dir = DataDir,
 		init = Init,
 		peers = Peers,
 		mine = Mine,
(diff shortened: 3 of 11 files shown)


src/ar_storage.erl
```diff
@@ -29,16 +29,17 @@ start() ->
...
-	filelib:ensure_dir(?BLOCK_DIR),
-	filelib:ensure_dir(?BLOCK_ENC_DIR),
-	filelib:ensure_dir(?WALLET_LIST_DIR),
-	filelib:ensure_dir(?HASH_LIST_DIR),
+	DataDir = ar_meta_db:get(data_dir) ++ "/",
+	filelib:ensure_dir(DataDir ++ ?TX_DIR ++ "/"),
+	filelib:ensure_dir(DataDir ++ ?BLOCK_DIR ++ "/"),
+	filelib:ensure_dir(DataDir ++ ?BLOCK_ENC_DIR ++ "/"),
+	filelib:ensure_dir(DataDir ++ ?WALLET_LIST_DIR ++ "/"),
+	filelib:ensure_dir(DataDir ++ ?HASH_LIST_DIR ++ "/"),
```
martin-torhage: Shouldn't also the WALLET_DIR be created here?

arweave-server
```diff
@@ -2,7 +2,7 @@
 
 set -o errexit
 
-mkdir -p ebin logs blocks hash_lists wallets wallet_lists txs blocks/enc priv data data/mnesia
+mkdir -p ebin logs priv
```
martin-torhage: Why is this needed? All three exists in git.

martin-torhage:
Some overall feedback/questions:

- Did you consider using `filename:join/1`? I think it makes sense but maybe not so important.
- It seems like we're using `.gitignore` files to ensure directories exists. That's bad cargo-culting in my opinion. Let's do something explicit like using empty `.keep` files.

docker-arweave-server
```diff
@@ -2,7 +2,7 @@
 
 set -o errexit
 
-mkdir -p ebin logs blocks wallets txs blocks/enc priv data/genesis_txs
+mkdir -p ebin logs priv
```
martin-torhage: Why is this needed? All three exists in git.

src/ar.erl
```diff
@@ -227,6 +232,9 @@ start(
 	error_logger:logfile({open, Filename = generate_logfile_name()}),
 	error_logger:tty(false),
 
+	ar_meta_db:start(),
+	ar_meta_db:put(data_dir, DataDir),
```
martin-torhage: I would prefer if all `ar_meta_db:put/2` calls are moved up here so they stay together.

Btw, our style guide doesn't say anything about blank lines but I would like to not have any.

src/ar_storage.erl
```diff
@@ -59,19 +60,20 @@ block_exists(Hash) ->
 %% @doc Move a block into the 'invalid' block directory.
 invalidate_block(B) ->
 	ar_block_index:remove(B#block.indep_hash),
+	DataDir = ar_meta_db:get(data_dir) ++ "/",
 	TargetFile =
 		lists:flatten(
 			io_lib:format(
 				"~s/invalid/~w_~s.json",
-				[?BLOCK_DIR, B#block.height, ar_util:encode(B#block.indep_hash)]
+				[DataDir ++ ?BLOCK_DIR, B#block.height, ar_util:encode(B#block.indep_hash)]
```
martin-torhage: Let's put the whole `lists:flatten/1` call in a function that takes a block as argument.

src/ar_storage.erl
```diff
@@ -325,7 +327,7 @@ write_tx(Tx) ->
 		Name = lists:flatten(
 			io_lib:format(
 				"~s/~s.json",
-				[?TX_DIR, ar_util:encode(Tx#tx.id)]
+				[ar_meta_db:get(data_dir) ++ "/" ++ ?TX_DIR, ar_util:encode(Tx#tx.id)]
```
martin-torhage: I would like to put the whole `lists:flatten/1` call in a function that takes a transaction as argument.

src/ar_storage.erl
```diff
@@ -29,16 +29,17 @@ start() ->
...
 ensure_directories() ->
-	filelib:ensure_dir(?TX_DIR),
-	filelib:ensure_dir(?BLOCK_DIR),
-	filelib:ensure_dir(?BLOCK_ENC_DIR),
-	filelib:ensure_dir(?WALLET_LIST_DIR),
-	filelib:ensure_dir(?HASH_LIST_DIR),
+	DataDir = ar_meta_db:get(data_dir) ++ "/",
+	filelib:ensure_dir(DataDir ++ ?TX_DIR ++ "/"),
+	filelib:ensure_dir(DataDir ++ ?BLOCK_DIR ++ "/"),
+	filelib:ensure_dir(DataDir ++ ?BLOCK_ENC_DIR ++ "/"),
```
martin-torhage: I don't understand what "block enc" means.

src/ar_storage.erl
```diff
@@ -29,16 +29,17 @@ start() ->
...
+	filelib:ensure_dir(DataDir ++ ?WALLET_LIST_DIR ++ "/"),
```
martin-torhage: The slash in the end should not be needed. We have it in many places.

martin-torhage:
I've added some inline comments. Here are the overall feedback:

 - Have you seen the `filename: join/1` function? It might be cleaner it many cases.
 - I suggest adding something like this in `ar` and use it as much as possible:
```
data_filename(Components) ->
    filename:join([ar_meta_db:get(data_dir) | Components]).
```

I lost one comment before I posted it. Will add it when I remember what it was about :)

ldmberman:
Yes, `join` makes sense off course. I mostly preserved the style of the code as it used to be but I am happy to improve it.

src/ar_storage.erl
```diff
@@ -29,16 +29,17 @@ start() ->
...
+	filelib:ensure_dir(DataDir ++ ?HASH_LIST_DIR ++ "/"),
 	filelib:ensure_dir(?LOG_DIR).
```
martin-torhage: This dir already exists in git. Can we remove this line?

src/ar_storage.erl
ldmberman: `ar_wallet:new_keyfile` takes care of it.

docker-arweave-server
ldmberman: The Dockerfile which was supposed to be used with this script has a dedicated stage for running the server where it just copies some of the build artifacts from the other stage, without copying the whole git folder. Creating `ebin` and `priv` (unlike `logs`) looks unnecessary anyway but I would leave these scripts as-is until we specifically dedicate some time to update or remove them.

src/ar.erl
ldmberman: I like the visual separation of the blocks, but comments can often be used for that. I am ok with not using blank lines.

src/ar_storage.erl
ldmberman: It is in fact needed here, to make `filelib:ensure_dir` create a directory if one does not exist.

src/ar.erl
samcamwilliams: +1 for comments

src/ar_storage.erl
ldmberman: Renamed into `ENCRYPTED_BLOCK_DIR`, hope it's clear now.

src/ar_storage.erl
martin-torhage: ✅

docker-arweave-server
martin-torhage: ✅

arweave-server
martin-torhage: ✅

src/ar_storage.erl
martin-torhage: ✅

src/ar_storage.erl
martin-torhage: Ohhh I made false assumptions ✅

src/ar.erl
martin-torhage: ✅

ldmberman:
Linking one of the old tickets to the PR, https://github.com/ArweaveTeam/arweave/issues/25.

src/ar_storage.erl
martin-torhage: ✅

src/ar_storage.erl
martin-torhage: ✅

PR #135, opened 2019-04-08
Storage content policy firewall

ldmberman:


src/ar_weave.erl
```diff
@@ -351,20 +352,3 @@ read_genesis_txs() ->
 		[],
 		Files
 	).
-
-create_genesis_txs() ->
```
martin-torhage: I think it makes sense to keep this code even though we're not planning to use it again. Did you get Sam's opinion on this? I think it makes more sense to remove the data than removing the code, so maybe we should remove the `data/genesis_wallets.csv` file or truncate it (into only one line with the column names).

src/ar_firewall.erl
```diff
@@ -1,71 +1,144 @@
...
-	FwServer ! {scan_tx, self(), Data},
+%% @doc Check whether a received TX matches the firewall rules.
+scan_tx(TX) ->
+	FwServer = whereis(default_firewall),
+	FwServer ! {scan_tx, self(), TX},
 	receive
 		{scanned_tx, Response} ->
-			ar:report([{scanned_object_of_type, tx}, {response, Response}]),
 			Response
+	after 10000 ->
```
martin-torhage: This doesn't work. Let's talk about it.

src/ar_firewall.erl
```diff
@@ -77,6 +150,60 @@ load_blacklist_test() ->
...
+					binary = <<"BADCONTENT3">>
+				}
+		}
+	],
+	{Sigs, _, _ } = av_sigs:load(["test/test_sig.txt"]),
+	lists:foreach(
+		fun(ExpectedSig) ->
+			?assert(
+				lists:member(ExpectedSig, Sigs),
+				io:format("Expected ~p to be in ~p", [ExpectedSig, Sigs])
```
martin-torhage: Nice! We should do this more 👍👍

src/ar_firewall.erl
```diff
@@ -1,71 +1,144 @@
...
+		end,
+		[],
+		TX#tx.tags
+	),
+	ScanList = [{ar_util:encode(TX#tx.id), TXIDSigs},{TX#tx.data, ContentSigs},{TX#tx.target, ContentSigs}|Tags],
+	case lists:any(
+		fun({Data, Sigs}) ->
+			case av_detect:is_infected(Data, Sigs) of
+				{true, _} ->
+					ar:info([{ar_firewall, reject_tx}, {tx, ar_util:encode(TX#tx.id)}, {content, Data}]),
```
martin-torhage: Seems like this will dump the complete binary in the logfile. I assumed error_logger would truncate it, but it doesn't seem to do that.

src/ar_firewall.erl
```diff
@@ -1,71 +1,144 @@
...
+			do_nothing
+	end.
+
+reload() ->
+	Firewall = whereis(default_firewall),
+	case Firewall of
+		undefined ->
+			do_nothing;
+		_ ->
+			erlang:unregister(default_firewall),
```
martin-torhage: There is no need to create this gap when no firewall is running. We can instead start the new firewall first, register it and then stop the old one.

src/ar_firewall.erl
```diff
@@ -1,71 +1,144 @@
...
+		undefined ->
+			F = spawn(
+				fun() ->
+					server(#state {
+						tx_id_sigs = av_sigs:load(ar_meta_db:get(transaction_blacklist)),
+						content_sigs = av_sigs:load(ar_meta_db:get(content_policies))
+					})
+				end
+			),
+			erlang:register(default_firewall, F);
```
martin-torhage: I see no reason to use a different name than `ar_firewall`. It's a singleton which you cannot start more of, so it makes sense to use `?MODULE`.

src/ar_storage.erl
```diff
@@ -115,12 +118,26 @@ write_block(RawB) ->
...
+					true;
+				reject ->
+					false
+			end
+		end,
+		TXs
+	),
+	ar_storage:write_tx(ScannedTXs),
+	ar_storage:write_block(BShadow),
+	ar_tx_search:update_tag_table(BShadow).
```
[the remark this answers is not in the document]
ldmberman: It won't fail cause `ar_storage:read_tx` can accept a record. Let's think if we can make it clearer without spending too much time on it.

src/ar_node_utils.erl
```diff
@@ -226,35 +233,40 @@ start_mining(#{
...
-				ar:report([{node_starting_miner, Node}, {recall_block, RecallB#block.height}])
+			case is_record(RecallB, block) of
+				false ->
+					ar:report_console([{erroneous_recall_block, RecallB}]);
+				true ->
+					ar_miner_log:started_hashing(),
+					ar:report([{node_starting_miner, Node}, {recall_block, RecallB#block.height}])
 			end,
-			RecallBFull = make_full_block(
+			case make_full_block(
```
martin-torhage: This is duplicated in `integrate_new_block`. What about moving this into a function with which takes the recall block shadow? E.g. `store_recall_block_key(RecallBShadow)`

src/ar_node_utils.erl
```diff
@@ -35,19 +35,26 @@ get_conflicting_txs(StateTXs, TX) ->
...
-				FinalB ->
+				{error, unavailable} ->
+					GetBlockFromPeersFun();
+				{error, {txs_missing, MissingTXIDs}} ->
+					ar:info([
+						{transactions_missing_on_disk_for_block, ar_util:encode(ID)},
+						{missing_txs, lists:map(fun ar_util:encode/1, MissingTXIDs)}
+					]),
+					GetBlockFromPeersFun();
+				{ok, FinalB} ->
```
martin-torhage: 6 lines further down (I cannot make a comment there), this function returns `make_full_block(ID, BHL)`. We need to case it to return the block or `unavailable` instead of the ok or error tuple.

src/ar_node_utils.erl
```diff
@@ -35,19 +35,26 @@ get_conflicting_txs(StateTXs, TX) ->
...
 	].
 
-%% @doc Get a specific full block (a block containing full txs) via
-%% blocks indep_hash.
+%% @doc Get a full block (a block containing all transactions) by the independent hash.
+%%      Try to find the block locally first. If we do not have the full block on disk, try to download it from peers.
 get_full_block(Peers, ID, BHL) when is_list(Peers) ->
-	% check locally first, if not found ask list of external peers for block
+	GetBlockFromPeersFun = fun() ->
+		get_full_block_from_remote_peers(ar_util:unique(Peers), ID, BHL)
```
martin-torhage: I see no reason why we don't store the block after fetching it. Do you?

ldmberman: We are storing such blocks, but the calls are in `ar_node_worker` and `ar_fork_recovery`, where the blocks are processed.

Makefile
```diff
@@ -13,7 +13,11 @@ ERL_OPTS= -pa ebin/ \
 test_all: test test_apps
 
 test: all
-	@erl $(ERL_OPTS) -noshell -s ar test_coverage -s init stop
+	@erl $(ERL_OPTS) -noshell -s ar test_with_coverage -s init stop
+
+distributed-test: all
```
martin-torhage: "distributed-test" is misleading to me. How do you feel about calling it "multi-node-test"? Or actually, in this context it should probably be named "test-multi-node".

src/ar_firewall.erl
ldmberman: I've rewritten it to log the matched signatures, makes much more sense this way.

src/ar_firewall.erl
ldmberman: ✅

src/ar_firewall.erl
ldmberman: ✅

src/ar_storage.erl
```diff
@@ -115,12 +118,26 @@ write_block(RawB) ->
...
+			case ar_firewall:scan_tx(TX) of
+				accept ->
+					true;
+				reject ->
+					false
+			end
+		end,
+		TXs
+	),
+	ar_storage:write_tx(ScannedTXs),
```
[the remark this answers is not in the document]
ldmberman: ✅

src/ar_node_utils.erl
```diff
@@ -226,35 +233,40 @@ start_mining(#{
...
 					end
 			end,
 			StateIn;
 		RecallB ->
-			if not is_record(RecallB, block) ->
-				ar:report_console([{erroneous_recall_block, RecallB}]);
-			true ->
-				ar_miner_log:started_hashing(),
-				ar:report([{node_starting_miner, Node}, {recall_block, RecallB#block.height}])
+			case is_record(RecallB, block) of
```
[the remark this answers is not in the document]
ldmberman: ✅

src/ar_node_utils.erl
ldmberman: ✅

Makefile
ldmberman: As discussed, joined the two targets into one for convenience. The same for `make session`.

src/ar_firewall.erl
martin-torhage: Wait a sec, you probably wanted to use `io_lib:format/2` and not `io:format/2` :) Obviously this is not tested, so I wonder if it really is alright to have an iolist as a comment. I think we should render it into a string.

src/ar_firewall.erl
martin-torhage: Oh turns out it's not possible to atomically reregister a name, so we still have a race condition here. It's probably good enough but we can easily get rid of the race condition by keeping the old process and just feeding it with a new state. I'm ok leaving it like this is how we do it in `ar_http_iface_server:reregister/2`.

ldmberman: Yeah, there is still a tiny gap, I think it's fine.

src/ar_storage.erl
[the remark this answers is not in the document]
martin-torhage: Ah I see 👍👍We should still stalk about the second bullet point.

src/ar_storage.erl
```diff
@@ -115,12 +118,26 @@ write_block(RawB) ->
...
+					true;
+				reject ->
+					false
+			end
+		end,
+		TXs
+	),
+	write_tx(ScannedTXs),
+	write_block(BShadow),
+	ar_tx_search:update_tag_table(BShadow).
```
martin-torhage: We're passing in the non-filtered TXs to `ar_tx_search:update_tag_table/1` here. Shouldn't we do this instead:
`ar_tx_search:update_tag_table(BShadow#block { txs = ScannedTXs }).`

src/ar_firewall.erl
martin-torhage: ✅

src/ar_node_utils.erl
martin-torhage: Yes, looks good. I'm putting my ✅on it too 😉

src/ar_node_utils.erl
martin-torhage: But `get_full_block/3` is using `ar_storage`, so it makes sense to me to also write to the storage. But the most important thing is that someone is writing it to storage. Let's just hope it stays like this...

src/ar_storage.erl
ldmberman: Very good point!

src/ar_firewall.erl
ldmberman: No, `assert` does not print the result of `io_lib:format` in case of a failure but it does execute the code so `io:format` does the trick.

src/ar_firewall.erl
ldmberman: ✅

src/ar_storage.erl
ldmberman: ✅

src/ar.erl
```diff
@@ -123,6 +127,8 @@ main("") ->
 			{"new_mining_key", "Generate a new keyfile, apply it as the reward address"},
 			{"load_mining_key (file)", "Load the address that mining rewards should be credited to from file."},
 			{"content_policy (file)", "Load a content policy file for the node."},
+			{"transaction_blacklist (file)", "A .txt file containing blacklisted transactions. "
+											 "Every line has to be a Base64 encoded transaction ID."},
```
martin-torhage: Here is my suggested improvement of the description:

A .txt file containing blacklisted transactions. One Base64 encoded transaction ID per line.

src/ar.erl
```diff
@@ -86,7 +89,8 @@
 	internal_api_secret = not_set,
 	enable = [],
 	disable = [],
-	content_policies = []
+	content_policies = [],
+	transaction_blacklist = []
```
martin-torhage: Looks like this is a list of filenames. I think it makes sense to name it in plural like `content_policies`.

For the record, I would rather call them `content_policy_files` and `transaction_blacklist_files`.

src/ar_bridge.erl
```diff
@@ -81,6 +81,20 @@ add_remote_peer(PID, Node) ->
 			PID ! {add_peer, remote, Node}
 	end.
 
+-ifdef(DEBUG).
+%% Do not filter out loopback IP addresses with custom port in the debug mode
+%% to allow multiple local VMs to peer with each other.
+is_loopback_ip({_, _, _, _, Port}) -> Port == ar_meta_db:get(port).
```
martin-torhage: This whole is_loopback_ip and how we use it is weird. I think what we're trying to do is to not add self as a remote. It would make sense to call the function `is_peer_self` and do some adjustments. Let's talk about this.

src/ar.erl
```diff
@@ -405,11 +415,19 @@ generate_logfile_name() ->
...
 		io_lib:format(
-			"~s/session_~4..0b-~2..0b-~2..0b_~2..0b-~2..0b-~2..0b.log",
-			[?LOG_DIR, Yr, Mo, Da, Hr, Mi, Se]
+			"~s/session_~4..0b-~2..0b-~2..0b_~2..0b-~2..0b-~2..0b~s.log",
+			[?LOG_DIR, Yr, Mo, Da, Hr, Mi, Se, maybe_node()]
 		)
 	).
 
+maybe_node() ->
+	case node() of
```
martin-torhage: What do you think about getting the sname from `init:get_argument(sname)` instead?

ldmberman: ✅

src/ar.erl
ldmberman: ✅

src/ar.erl
ldmberman: ✅

src/ar_node_utils.erl
martin-torhage: ✅

src/ar_node_utils.erl
martin-torhage: ✅

Makefile
martin-torhage: ✅

src/ar_bridge.erl
martin-torhage: ✅

src/ar_firewall.erl
martin-torhage: ✅

src/ar_firewall.erl
martin-torhage: ✅

src/ar_tx_blacklist.erl
```diff
@@ -0,0 +1,25 @@
+-module(ar_tx_blacklist).
+-export([load_from_files/1, is_element/2]).
+-include("ar.hrl").
+
+load_from_files(Files) ->
+	sets:from_list(lists:flatten(lists:map(fun load_from_file/1, Files))).
+
+is_element(TX, TXBlacklist) ->
+	sets:is_element(ar_util:encode(TX#tx.id), TXBlacklist).
```
martin-torhage: Why do we encode the IDs before we put them in the set? I see no reason to do it.

ldmberman: We actually simply not decode them before we put them in the set. Having them encoded in the file is convenient cause people can make sense of them. What do you think?

ldmberman: We could decode them before putting in the set to make it smaller though..

ldmberman: ✅

src/ar_tx_blacklist.erl
```diff
@@ -0,0 +1,30 @@
+-module(ar_tx_blacklist).
+-export([load_from_files/1, is_element/2]).
+-include("ar.hrl").
+
+load_from_files(Files) ->
+	sets:from_list(lists:flatten(lists:map(fun load_from_file/1, Files))).
+
+is_element(TX, TXBlacklist) ->
```
martin-torhage: How about renaming this function to `is_blacklisted` instead?

src/av/av_detect.erl
```diff
@@ -2,43 +2,75 @@
...
-%% Only files can be infected, return false for directories.
-% is_infected(FileName, Sigs) ->
-% 	case try_read(FileName) of
-% 		error -> pass;
-% 		Bin -> do_is_infected(FileName, Bin, Sigs)
-% 	end.
+%% Given a list of binaries and a set of signatures, test whether some binaries
+%% contain illicit content or have illicit hash. If not, return false, if yes, return true and
+%% the matching signatures.
+is_infected(Binaries, {Sigs, BinaryPattern}) ->
```
martin-torhage: What do you think about renaming `Binaries` to `Subjects`?

src/av/av_detect.erl
```diff
@@ -2,43 +2,75 @@
...
+	HashSigs = hash_sigs(Sigs),
+	MatchedHashSigs = case HashSigs of
```
martin-torhage: Can't we replace this whole case with:
`MatchedHashSigs = full_check(Binaries, HashSigs),`?

src/av/av_detect.erl
```diff
@@ -2,43 +2,75 @@
...
+		[] ->
+			[];
+		_ ->
+			full_check(Binaries, HashSigs)
+	end,
+	BinarySigs = binary_sigs(Sigs),
+	MatchedBinarySigs = case BinarySigs of
+		[] ->
+			[];
```
martin-torhage: We haven't used the `BinaryPattern` here. Isn't that a problem?

src/av/av_detect.erl
```diff
@@ -2,43 +2,75 @@
...
+		_ ->
+			case quick_check(<< Binary || Binary <- Binaries >>, BinaryPattern) of
```
martin-torhage: Let's talk about this.

src/av/av_detect.erl
martin-torhage: ✅

src/av/av_detect.erl
martin-torhage: ✅

src/av/av_detect.erl
martin-torhage: ✅

src/av/av_detect.erl
martin-torhage: ✅

PR #136, opened 2019-04-12
Replace Elli with Cowboy

maxmellen:
Also improve how we use `rebar3` to compile dependencies


.gitmodules
```diff
@@ -1,9 +1,27 @@
 [submodule "lib/prometheus"]
 	path = lib/prometheus
-	url = https://github.com/deadtrickster/prometheus.erl.git
+	url = https://github.com/ArweaveTeam/prometheus.erl.git
```
ldmberman: `git diff` now shows the change in the subproject commit hash. I suppose we need to commit `lib/prometheus` along with this change.

ldmberman:
We need to check out `rebar.lock`.

.gitmodules
ldmberman: Nevermind, I tried some old commit by mistake initially.

src/ar.hrl
```diff
@@ -19,9 +19,7 @@
 	]).
 
 -define(DEFAULT_RESPONSE_HEADERS,
-	[
-		{<<"Access-Control-Allow-Origin">>, <<"*">>}
-	]).
+	#{<<"access-control-allow-origin">> => <<"*">>}).
```
ldmberman: HTTP headers are case-insensitive but it feels safer to preserve the case for the response headers. What would you say?

src/ar.erl
```diff
@@ -248,8 +262,11 @@ start(
 	if Clean -> ar_storage:clear(); true -> do_nothing end,
 	%% Register prometheus stats collector.
 	application:ensure_started(prometheus),
+	application:ensure_all_started(prometheus_cowboy),
 	prometheus_registry:register_collector(prometheus_process_collector),
 	prometheus_registry:register_collector(ar_metrics_collector),
+	%% Start Cowboy and its dependencies
+	application:ensure_all_started(cowboy),
```
martin-torhage: I see no reason not to pattern match on the expected returned ok tuple. We should probably do this on all the `application:ensure_all_started/1` calls in this function.

src/ar_http_iface_server.erl
```diff
@@ -880,337 +33,57 @@ reregister(Name, Node) ->
...
+						cowboy_router,
+						cowboy_handler
+					],
+					env => #{dispatch => Dispatch},
+					metrics_callback => fun prometheus_cowboy2_instrumenter:observe/1,
+					stream_handlers => [cowboy_metrics_h, cowboy_stream_h]
+				},
+
+			case TransportOpts of
+				clear ->
```
martin-torhage: How do you feel about calling this `no_tls`?

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -0,0 +1,1142 @@
...
+
+read_complete_body(Req) ->
+	read_complete_body(Req, <<>>).
+
+read_complete_body(UnreadReq, Acc) ->
+	case cowboy_req:read_body(UnreadReq) of
+		{more, Data, ReadReq} ->
+			read_complete_body(ReadReq, <<Acc/binary, Data/binary>>);
+		{ok, Data, ReadReq} ->
+			{<<Acc/binary, Data/binary>>, ReadReq}
```
martin-torhage: Even though Erlang should be able to handle this efficiently, the standard way of doing this, which is also probably even more efficient, would this:
```
		{more, Data, ReadReq} ->
			read_complete_body(ReadReq, [Data | Acc]);
		{ok, Data, ReadReq} ->
			Body = iolist_to_binary(lists:reverse([Data | Acc]))
			{Body, ReadReq}
```

What do you think about this?

martin-torhage: I just saw that the binary accumulator approach is in the example code in the Cowboy docs 👍👍

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -0,0 +1,1142 @@
...
+do_handle(<<"POST">>, [<<"peers">>], Req) ->
+	{BlockJSON, _} = read_complete_body(Req),
+	case ar_serialize:dejsonify(BlockJSON) of
+		{Struct} ->
+			{<<"network">>, NetworkName} = lists:keyfind(<<"network">>, 1, Struct),
+			case (NetworkName == <<?NETWORK_NAME>>) of
+				false ->
+					{400, #{}, <<"Wrong network.">>};
+				true ->
+					Peer = arweave_peer(Req),
```
martin-torhage: We're reusing the outdated `Req` variable here. `read_complete_body/1` above returns the new one.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -0,0 +1,1142 @@
...
+	end;
+%% @doc Return the balance of the wallet specified via wallet_address.
+%% GET request to endpoint /wallet/{wallet_address}/balance
+do_handle(<<"GET">>, [<<"wallet">>, Addr, <<"balance">>], _Req) ->
+	case safe_decode(Addr) of
+		{error, invalid} ->
+			{400, #{}, <<"Invalid address.">>};
+		{ok, AddrOK} ->
+			%% ar_node:get_balance/2 can time out which is not suitable for this
+			%% use-case. It would be better if it never timed out so that Elli
```
ldmberman: One more Elli mention to remove.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -0,0 +1,1142 @@
...
+	CowboyStatus = handle208(Status),
+	Req1 = cowboy_req:reply(CowboyStatus, Headers, Body, Req0),
+	{ok, Req1, State}.
+
+%%%===================================================================
+%%% Private functions.
+%%%===================================================================
+
+do_handle(Req) ->
+	%% Inform ar_bridge about new peer, performance rec will be updated from ar_metrics
```
ldmberman: The ar_metrics mention is irrelevant now.

martin-torhage:
I've finished reviewing and added inline comments.

I also found that we're using Cowboy 2.6.1 but 2.6.3 is released, so we should upgrade.

src/ar.hrl
maxmellen: This is actually cowboy specific behaviour. Cowboy explicitly asks the developer to always refer to headers using only lowercase letters.

src/ar.erl
maxmellen: In the `{ok, Started}` tuple, is the order of the started apps deterministic?

src/ar.hrl
ldmberman: Does cowboy explain this? I think, for us it's just a matter of backwards-compatibility, in case some clients rely on headers specified in the particular case.

src/ar_http_iface_server.erl
maxmellen: I took the wording from cowboy's `start_clear`/`start_tls` but I can't find other languages/libraries that refer to non-secure HTTP as clear... I'm up for any wording

src/ar_http_iface_cowboy_handler.erl
maxmellen: 👍

src/ar_http_iface_cowboy_handler.erl
maxmellen: Is the comment up to date? Can I simply update the comment to read:
> It would be better if it never timed out so that Cowboy would handle the timeout instead.

src/ar_http_iface_cowboy_handler.erl
maxmellen: I'm not sure I understand the comment... Should I say the following?
> performance rec will be updated from cowboy_metrics_h

src/ar_http_iface_cowboy_handler.erl
ldmberman: Yep

src/ar.hrl
maxmellen: Yep, super explicitly:

> The header name must be given as a lowercase binary string. While header names are case insensitive, Cowboy requires them to be given as lowercase to function properly.

https://ninenines.eu/docs/en/cowboy/2.6/manual/cowboy_req.set_resp_headers/

maxmellen:
@martin-torhage Updated 👍

src/ar.hrl
ldmberman: Ouch, that's fucked up. Okay, then we have no choice.

src/ar.erl
llaisdy: `application` works out the order of starting based on the graph of dependencies, so the order might not be deterministic if there is more than one possible ordering.  I generally use something like:

```
{ok, Started} = application:ensure_all_started(cowboy).
```

src/ar_prometheus_cowboy_labels.erl
```diff
@@ -0,0 +1,109 @@
...
+normalize_method(<<"GET">>) -> 'GET';
+normalize_method(<<"HEAD">>) -> 'HEAD';
+normalize_method(<<"POST">>) -> 'POST';
+normalize_method(<<"PUT">>) -> 'PUT';
+normalize_method(<<"DELETE">>) -> 'DELETE';
+normalize_method(<<"CONNECT">>) -> 'CONNECT';
+normalize_method(<<"OPTIONS">>) -> 'OPTIONS';
+normalize_method(<<"TRACE">>) -> 'TRACE';
+normalize_method(<<"PATCH">>) -> 'PATCH';
+normalize_method(_) -> undefined.
```
martin-torhage: It's not really undefined so maybe we should call it `other` instead?

src/ar_prometheus_cowboy_labels.erl
```diff
@@ -0,0 +1,109 @@
...
+name_route([<<"services">>]) ->
+	"/services";
+name_route([<<"height">>]) ->
+	"/height";
+name_route([<<_Hash:43/binary, _MaybeExt/binary>>]) ->
+	"/{hash}[.{ext}]";
+name_route([<<"metrics">>]) ->
+	"/metrics";
+name_route(_) ->
+	undefined.
```
martin-torhage: Same here as above. How do you feel about calling this one `other_route` instead?

martin-torhage:
More review feedback:
 - I don't understand where the rebar3 binary came from. A reference would be helpful.
 - Let's talk about the `_checkouts` dir.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -228,13 +232,17 @@ do_handle(<<"POST">>, [<<"wallet">>], Req) ->
...
 %% specified in ar_serialize.
 do_handle(<<"POST">>, [<<"tx">>], Req) ->
-	{TXJSON, ReadReq} = read_complete_body(Req),
-	TX = ar_serialize:json_struct_to_tx(TXJSON),
-	case handle_post_tx(TX) of
-		ok ->
-			{200, #{}, <<"OK">>, ReadReq};
-		{error_response, {Status, Headers, Body}} ->
-			{Status, Headers, Body, ReadReq}
+	case read_complete_body(Req) of
```
ldmberman: There is also `POST /wallet` (above this one).

PR #144, opened 2019-04-29
Serve transactions on isolated origins

maxmellen:
This PR implements the logic to serve each transaction under its own origin. This is done by deriving a _label_ from the transaction and its containing block's hash.

For example, for the transaction `vP_3_IZBng33AaW8MCfTqseJxdG-OWbfOb_ZU1UhR7s`, the generated label is `xvc4dssncn5p`, so accessing https://gateway.example/vP_3_IZBng33AaW8MCfTqseJxdG-OWbfOb_ZU1UhR7s will redirect to https://xvc4dssncn5p.gateway.example/vP_3_IZBng33AaW8MCfTqseJxdG-OWbfOb_ZU1UhR7s.

This PR also implements the logic to associate a custom domain name with a specific transaction. This is done by specifying a list of known custom domains at node startup. These domains should point to the node either with an A or CNAME record. The association of the domain to the transaction is then done by looking for a TXT record for that domain with the name `_arweave.<custom domain name>` and the transaction id as its value.

There's a few points I still want to discuss, although I would still appreciate feedback on what's been done so far.

The points I still want to discuss are the following:

- Currently, the location of the certificate files is inferred: the main certificate and key are expected to live at `priv/tls/{cert,key}.pem` and the certificate files for the custom domains are expected to live under `priv/tls/<custom domain>/{cert,key}.pem`. Symlinking to these locations seems to work fine. I feel like this should be good enough for this first iteration. What do you think?
- Currently, most of the business logic still lives in the cowboy handler. I feel like it might be nice to split as much of it as possible into a separate module. What do you think?
- Currently, the configuration for the domains is done through CLI arguments. Specifying the `apex_domain` value (the main domain name for the node) triggers the new behaviour. Leaving it absent will retain the old behaviour. It might be nice to turn this into a more explicit flag. What do you think?
- There's also currently no interface to reconfigure the domains without restarting the node and passing in other CLI arguments. Again, I feel like this should be good enough for this first iteration. What do you think?

### DIFF
--- a/src/ar.erl
+++ b/src/ar.erl
@@ -91,7 +91,9 @@
 	enable = [],
 	disable = [],
 	content_policy_files = [],
-	transaction_blacklist_files = []
+	transaction_blacklist_files = [],
+	gateway = off,
+	custom_domains = []
 }).
 
 %% @doc Command line program entrypoint. Takes a list of arguments.
@@ -142,7 +144,9 @@ main("") ->
 				)
 			},
 			{"enable (feature)", "Enable a specific (normally disabled) feature. For example, subfield_queries."},
-			{"disable (feature)", "Disable a specific (normally enabled) feature. For example, api_compat mode."}
+			{"disable (feature)", "Disable a specific (normally enabled) feature. For example, api_compat mode."},
+			{"gateway (port) (domain)", "Run a gateway on the specified port and domain"},
+			{"custom_domain (domain)", "Add a domain to the list of supported custom domains."}
 		]
 	),
 	erlang:halt();
@@ -203,6 +207,10 @@ parse(["enable", Feature | Rest ], O = #opts { enable = Enabled }) ->
 	parse(Rest, O#opts { enable = [ list_to_atom(Feature) | Enabled ] });
 parse(["disable", Feature | Rest ], O = #opts { disable = Disabled }) ->
 	parse(Rest, O#opts { disable = [ list_to_atom(Feature) | Disabled ] });
+parse(["gateway", Port, Domain | Rest ], O = #opts { gateway = off }) ->
+	parse(Rest, O#opts { gateway = {on, list_to_integer(Port), list_to_binary(Domain)} });
+parse(["custom_domain", Domain|Rest], O = #opts { custom_domains = Ds }) ->
+	parse(Rest, O#opts { custom_domains = [ list_to_binary(Domain) | Ds ] });
 parse([Arg|_Rest], _O) ->
 	io:format("~nUnknown argument: ~s. Terminating.~n~n", [Arg]),
 	erlang:halt().
@@ -238,7 +246,9 @@ start(
 		enable = Enable,
 		disable = Disable,
 		content_policy_files = ContentPolicyFiles,
-		transaction_blacklist_files = TransactionBlacklistFiles
+		transaction_blacklist_files = TransactionBlacklistFiles,
+		gateway = GatewayOpts,
+		custom_domains = GatewayCustomDomains
 	}) ->
 	%% Start the logging system.
 	error_logger:logfile({open, Filename = generate_logfile_name()}),
@@ -405,6 +415,12 @@ start(
 		{http_search_node, SearchNode},
 		{http_bridge_node, Bridge}
 	]),
+	case GatewayOpts of
+		{on, GatewayPort, GatewayDomain} ->
+			ar_gateway_server:start(GatewayPort, GatewayDomain, GatewayCustomDomains);
+		off ->
+			do_nothing
+	end,
 	case Polling of
 		true ->
 			ar_meta_db:put(polling_mode, true),
--- a/src/ar.erl
+++ b/src/ar.erl
@@ -112,7 +112,7 @@ main("") ->
 		end,
 		[
 			{"peer (ip:port)", "Join a network on a peer (or set of peers)."},
-			{"start_hash_list (file)", "Start the node from a given block."},
+			{"start_hash_list (hash)", "Start the node from a given block."},
 			{"mine", "Automatically start mining once the netwok has been joined."},
 			{"port", "The local port to use for mining. "
 						"This port must be accessible by remote peers."},
@@ -183,8 +183,8 @@ parse(["disk_space", Size|Rest], O) ->
 	parse(Rest, O#opts { disk_space = (list_to_integer(Size)*1024*1024*1024) });
 parse(["load_mining_key", File|Rest], O)->
 	parse(Rest, O#opts { load_key = File });
-parse(["start_hash_list", IndepHash|Rest], O)->
-	parse(Rest, O#opts { start_hash_list = ar_util:decode(IndepHash) });
+parse(["start_hash_list", BHLHash|Rest], O)->
+	parse(Rest, O#opts { start_hash_list = ar_util:decode(BHLHash) });
 parse(["benchmark"|Rest], O)->
 	parse(Rest, O#opts { benchmark = true });
 parse(["auto_update", "false" | Rest], O) ->
@@ -332,7 +332,8 @@ start(
 						if Init -> ar_weave:init(ar_util:genesis_wallets(), Diff);
 						true -> not_joined
 						end;
-					_ -> ar_storage:read_hash_list(ar_util:decode(BHL))
+					_ ->
+						ar_storage:read_block_hash_list(BHL)
 				end,
 				0,
 				MiningAddress,
--- a/src/ar.erl
+++ b/src/ar.erl
@@ -400,13 +400,11 @@ start(
 		]
 	),
 	%% Start the first node in the gossip network (with HTTP interface).
-	ar_http_iface_server:start(
-		Port,
-		Node,
-		SearchNode,
-		undefined,
-		Bridge
-	),
+	ar_http_iface_server:start(Port, [
+		{http_entrypoint_node, Node},
+		{http_search_node, SearchNode},
+		{http_bridge_node, Bridge}
+	]),
 	case Polling of
 		true ->
 			ar_meta_db:put(polling_mode, true),
--- a/src/ar_domain.erl
+++ b/src/ar_domain.erl
@@ -1,0 +1,44 @@
+-module(ar_domain).
+
+-export([get_labeling/3, lookup_arweave_txt_record/1, derive_tx_label/2]).
+
+%%%===================================================================
+%%% Public interface.
+%%%===================================================================
+
+get_labeling(ApexDomain, CustomDomains, Hostname) ->
+	Size = byte_size(ApexDomain),
+	case binary:match(Hostname, ApexDomain) of
+		{0, Size} ->
+			apex;
+		{N, Size} ->
+			Label = binary:part(Hostname, {0, N-1}),
+			{labeled, Label};
+		nomatch ->
+			get_labeling_1(CustomDomains, Hostname)
+	end.
+
+lookup_arweave_txt_record(Domain) ->
+	case inet_res:lookup("_arweave." ++ binary_to_list(Domain), in, txt) of
+		[] ->
+			not_found;
+		[RecordChunks|_] ->
+			list_to_binary(lists:concat(RecordChunks))
+	end.
+
+derive_tx_label(TXID, BH) ->
+	Data = <<TXID/binary, BH/binary>>,
+	Digest = crypto:hash(sha256, Data),
+	binary:part(ar_base32:encode(Digest), {0, 12}).
+
+%%%===================================================================
+%%% Private functions.
+%%%===================================================================
+
+get_labeling_1(CustomDomains, Hostname) ->
+	case lists:member(Hostname, CustomDomains) of
+		true ->
+			{custom, Hostname};
+		false ->
+			unknown
+	end.
--- a/src/ar_gateway_h.erl
+++ b/src/ar_gateway_h.erl
@@ -1,0 +1,208 @@
+-module(ar_gateway_h).
+-behaviour(cowboy_handler).
+-export([init/2]).
+
+-include("ar.hrl").
+-define(BH_SEARCH_TIMEOUT, 25000).
+
+%%%===================================================================
+%%% Cowboy handler callback.
+%%%===================================================================
+
+%% @doc Handle gateway requests.
+%%
+%% When gateway mode is turned on, a domain name needs to be provided
+%% to act as the main domain name for this gateway. This domain name
+%% is referred to as the apex domain for this gateway. A number of
+%% custom domains that will map to specific transactions can also be
+%% provided.
+%%
+%% The main purpose of a gateway is to serve different transaction on
+%% different origins. This is done by deriving what will be referred
+%% to as a transaction label from the transaction's ID and its block's
+%% hash and prefixing that label to the apex domain of the gateway.
+%% For example:
+%%
+%%     Accessing    https://gateway.example/{txid}
+%%     redirects to https://{txlabel}.gateway.example/{txid}
+%%     where it then serves the transaction's content.
+%%
+%% A gateway also allows to configure custom domains which will
+%% redirect to specific transactions. For example, if
+%% custom.domain.example is defined as a custom domain:
+%%
+%%     Accessing https://custom.domain.example
+%%     triggers a TXT record lookup at _arweave.custom.domain.example
+%%     where a transaction ID is expected to be found.
+%%
+%% This transaction ID is then used to decide from which transaction
+%% to serve the content for this custom domain.
+%%
+%% The gateway also allows using relative links whether accessing a
+%% transaction's content through a labeled domain name or through a
+%% custom domain name. For example:
+%%
+%%     Accessing    https://{tx1label}.gateway.example/{tx2id}
+%%     redirects to https://{tx2label}.gateway.example/{tx2id}
+%%     where tx2's content is then served, and
+%%
+%%     Accessing    https://custom.domain.example/{txid}
+%%     redirects to https://{txlabel}.gateway.example/{txid}
+%%     where tx's content is then served.
+%%
+%% This allows webpages stored in transactions to link to each other
+%% without having to refer to any specific gateway. That is to say, to
+%% link to a different transaction in HTML, one would use
+%%
+%%     <a href="{txid}">Link</a>
+%%
+%% rather than
+%%
+%%     <a href="https://{txlabel}.gateway.example/{txid}">Link</a>
+init(Req, {Domain, CustomDomains} = State) ->
+	Hostname = cowboy_req:host(Req),
+	Req1 =
+		case ar_domain:get_labeling(Domain, CustomDomains, Hostname) of
+			apex -> handle_apex_request(Domain, Req);
+			{labeled, Label} -> handle_labeled_request(Domain, Label, Req);
+			{custom, CustomDomain} -> handle_custom_request(Domain, CustomDomain, Req);
+			unknown -> not_found(Req)
+		end,
+	{ok, Req1, State}.
+
+%%%===================================================================
+%%% Private functions.
+%%%===================================================================
+
+handle_apex_request(Domain, Req) ->
+	case resolve_path(cowboy_req:path(Req)) of
+		{tx, TXID, _} -> derive_label_and_redirect(Domain, TXID, Req);
+		root -> bad_request(Req);
+		invalid -> bad_request(Req);
+		not_found -> not_found(Req)
+	end.
+
+handle_labeled_request(Domain, Label, Req) ->
+	case resolve_path(cowboy_req:path(Req)) of
+		{tx, TXID, Filename} ->
+			handle_labeled_request_1(Domain, Label, TXID, Filename, Req);
+		root ->
+			bad_request(Req);
+		invalid ->
+			bad_request(Req);
+		not_found ->
+			not_found(Req)
+	end.
+
+handle_labeled_request_1(Domain, Label, TXID, Filename, Req) ->
+	case get_tx_block_hash(TXID) of
+		{ok, BH} ->
+			handle_labeled_request_2(Domain, Label, TXID, Filename, BH, Req);
+		not_found ->
+			not_found(Req)
+	end.
+
+handle_labeled_request_2(Domain, Label, TXID, Filename, BH, Req) ->
+	case ar_domain:derive_tx_label(TXID, BH) of
+		Label -> serve_tx(Filename, Req);
+		OtherLabel -> redirect_to_labeled_tx(Domain, OtherLabel, TXID, Req)
+	end.
+
+handle_custom_request(Domain, CustomDomain, Req) ->
+	case resolve_path(cowboy_req:path(Req)) of
+		root -> handle_custom_request_1(CustomDomain, Req);
+		{tx, TXID, _} -> derive_label_and_redirect(Domain, TXID, Req);
+		invalid -> bad_request(Req)
+	end.
+
+resolve_path(Path) ->
+	case Path of
+		<<"/">> -> root;
+		<<"/", Hash/binary>> -> resolve_path_1(Hash);
+		_ -> invalid
+	end.
+
+resolve_path_1(Hash) ->
+	case get_tx_from_hash(Hash) of
+		{ok, TXID, Filename} -> {tx, TXID, Filename};
+		invalid -> invalid;
+		not_found -> not_found
+	end.
+
+derive_label_and_redirect(Domain, TXID, Req) ->
+	case get_tx_block_hash(TXID) of
+		{ok, BH} -> derive_label_and_redirect_1(Domain, TXID, BH, Req);
+		not_found -> not_found(Req)
+	end.
+
+get_tx_block_hash(TXID) ->
+	HttpSearchNode = whereis(http_search_node),
+	case ar_tx_search:get_tags_by_id(HttpSearchNode, TXID, ?BH_SEARCH_TIMEOUT) of
+		{ok, PseudoTags} -> get_tx_block_hash_1(PseudoTags);
+		{error, _} -> not_found
+	end.
+
+get_tx_block_hash_1(PseudoTags) ->
+	case proplists:get_value(<<"block_indep_hash">>, PseudoTags) of
+		undefined -> not_found;
+		BH -> {ok, BH}
+	end.
+
+derive_label_and_redirect_1(Domain, TXID, BH, Req) ->
+	Label = ar_domain:derive_tx_label(TXID, BH),
+	redirect_to_labeled_tx(Domain, Label, TXID, Req).
+
+redirect_to_labeled_tx(Domain, Label, TXID, Req) ->
+	Scheme = cowboy_req:scheme(Req),
+	Port = integer_to_binary(cowboy_req:port(Req)),
+	Hash = ar_util:encode(TXID),
+	Location = <<
+		Scheme/binary, "://",
+		Label/binary, ".",
+		Domain/binary, ":",
+		Port/binary, "/",
+		Hash/binary
+	>>,
+	cowboy_req:reply(301, #{<<"location">> => Location}, Req).
+
+bad_request(Req) ->
+	cowboy_req:reply(400, Req).
+
+not_found(Req) ->
+	cowboy_req:reply(404, Req).
+
+handle_custom_request_1(CustomDomain, Req) ->
+	case ar_domain:lookup_arweave_txt_record(CustomDomain) of
+		not_found ->
+			domain_improperly_configured(Req);
+		TxtRecord ->
+			handle_custom_request_2(TxtRecord, Req)
+	end.
+
+handle_custom_request_2(TxtRecord, Req) ->
+	case get_tx_from_hash(TxtRecord) of
+		{ok, _, Filename} -> serve_tx(Filename, Req);
+		invalid -> domain_improperly_configured(Req);
+		not_found -> domain_improperly_configured(Req)
+	end.
+
+get_tx_from_hash(Hash) ->
+	case ar_util:safe_decode(Hash) of
+		{ok, TXID} -> get_tx_from_hash_1(TXID);
+		{error, _} -> invalid
+	end.
+
+get_tx_from_hash_1(TXID) ->
+	case ar_storage:lookup_tx_filename(TXID) of
+		unavailable -> not_found;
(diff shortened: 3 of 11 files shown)


src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
 	end;
 
 %% @doc Catch case for requests made to unknown endpoints.
 %% Returns error code 400 - Request type not found.
-do_handle(_, _, Req) ->
+handle(_, _, Req) ->
+	not_found(Req).
+
+handle_serve_root(ApexDomain, Req) ->
+	Domain = cowboy_req:host(Req),
```
martin-torhage: Isn't this the hostname? Why not call it `Hostname`?

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+handle_serve_tx_apex_1(ApexDomain, TxHash, Req) ->
+	BlockHash = ?OK(get_tx_block_hash(TxHash)),
+	redirect_to_labeled_tx(ApexDomain, BlockHash, TxHash, Req).
+
+handle_serve_tx_labeled_1(ApexDomain, Label, TxHash, MaybeExt, Req) ->
+	BlockHash = ?OK(get_tx_block_hash(TxHash)),
+	handle_serve_tx_labeled_2(ApexDomain, Label, BlockHash, TxHash, MaybeExt, Req).
+
+get_tx_block_hash(TxHash) ->
+	case ar_tx_search:get_tags_by_id(
```
martin-torhage: The search index is not very reliably, so let's talk about our expectations.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+		ar_util:decode(TxHash),
+		infinity
+	) of
+		{ok, PseudoTags} ->
+			case proplists:get_value(<<"block_indep_hash">>, PseudoTags) of
+				undefined -> not_found;
+				BlockHash -> {ok, BlockHash}
+			end;
+		{error, _} ->
+			not_found
```
martin-torhage: Idea: Return 404 if we don't have this metadata for this tx

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+resolve_custom_domain_tx(Domain, Req) ->
+	case inet_res:lookup("_arweave." ++ binary_to_list(Domain), in, txt) of
+		[] ->
+			{502, #{}, <<"Domain improperly configured">>, Req};
+		[RecordChunks|_] ->
+			TxHash = list_to_binary(lists:concat(RecordChunks)),
+			handle(<<"GET">>, [<<"tx">>, TxHash, <<"data.html">>], Req)
+	end.
+
+handle_serve_tx_apex_1(ApexDomain, TxHash, Req) ->
```
martin-torhage: We have a naming convention to call transaction ID variables for `TXID`

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+	BlockHash = ?OK(get_tx_block_hash(TxHash)),
```
martin-torhage: We have a naming convention to call block hash variables for `BH`.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+	handle(<<"GET">>, [<<"tx">>, TxHash, <<"data.", Ext/binary>>], Req);
+handle_serve_tx_labeled_3(_, _, Req) ->
 	not_found(Req).
 
+redirect_to_labeled_tx(ApexDomain, BlockHash, TxHash, Req) ->
+	Scheme = cowboy_req:scheme(Req),
+	Label = derive_tx_label(TxHash, BlockHash),
+	Port = integer_to_binary(cowboy_req:port(Req)),
+
+	Location =
```
martin-torhage: Even though we don't do Erlang standard indentation, I think we can do better than this. How about this:
```
	Location = <<
		Scheme/binary, "://",
		Label/binary, ".",
		ApexDomain/binary, ":",
		Port/binary, "/",
		TxHash/binary
	>>,
```
Or this:
```
	Location = <<
			Scheme/binary, "://",
			Label/binary, ".",
			ApexDomain/binary, ":",
			Port/binary, "/",
			TxHash/binary
		>>,
```

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+	case lists:member(Domain, CustomDomains) of
+		true -> {custom, Domain};
+		false -> error('FIXME')
+	end.
+
+redirect_to_apex(ApexDomain, Req) ->
+	Scheme = cowboy_req:scheme(Req),
+	Port = integer_to_binary(cowboy_req:port(Req)),
+
+	Location =
```
martin-torhage: See my comment for `redirect_to_labeled_tx/4`

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
-		<<>> ->
-			do_handle(<<"GET">>, [<<"tx">>, Hash, <<"data.html">>], Req);
-		_ ->
-			not_found(Req)
+handle(<<"GET">>, [<<Hash:43/binary, MaybeExt/binary>>], Req) ->
+	case ar_meta_db:get(apex_domain) of
+		not_found ->
+			handle(<<"GET">>, [<<"tx">>, Hash, <<"data.", MaybeExt/binary>>], Req);
+		ApexDomain ->
+			handle_serve_tx_1(ApexDomain, Hash, MaybeExt, Req)
```
martin-torhage: We should decode `Hash` before passing it on and call it `TXID` in the next function.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+	case Label of
+		ExpectedLabel -> handle_serve_tx_labeled_3(TxHash, MaybeExt, Req);
+		_ -> redirect_to_labeled_tx(ApexDomain, BlockHash, TxHash, Req)
+	end.
+
+derive_tx_label(TxHash, BlockHash) ->
+	State0 = crypto:hash_init(sha256),
+	State1 = crypto:hash_update(State0, ar_util:decode(TxHash)),
+	State2 = crypto:hash_update(State1, ar_util:decode(BlockHash)),
+	Digest = crypto:hash_final(State2),
```
martin-torhage: What do you think about writing it like this instead?
```
	HashData = <<(ar_util:decode(TxHash))/binary, (ar_util:decode(BlockHash))/binary>>,
	Digest = crypto:hash(sha256, HashData),
```

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -691,7 +815,7 @@ arweave_peer(Req) ->
 split_path(Path) ->
 	lists:filter(
 		fun(C) -> C /= <<>> end,
-		string:split(Path, <<"/">>, all)
+		binary:split(Path, <<"/">>, [global])
```
martin-torhage: Did you sneak in a bugfix here? :) Do you know what endpoints are affected of this?

src/ar.erl
```diff
@@ -141,7 +144,10 @@ main("") ->
 				)
 			},
 			{"enable (feature)", "Enable a specific (normally disabled) feature. For example, subfield_queries."},
-			{"disable (feature)", "Disable a specific (normally enabled) feature. For example, api_compat mode."}
+			{"disable (feature)", "Disable a specific (normally enabled) feature. For example, api_compat mode."},
+			{"use_tls", "Run the HTTP server under TLS"},
+			{"apex_domain (domain)", "Set the apex domain to use when running this node as a gateway"},
```
martin-torhage: Suggestion: "Enable gateway mode and use this as the apex domain name"

src/ar_http_iface_server.erl
```diff
@@ -60,8 +69,37 @@ do_start(Port) ->
...
+				SniHosts = derive_sni_hosts(CustomDomains),
+				cowboy:start_tls(ar_cowboy_listener, [
+					{port, Port},
+					{certfile, "priv/tls/cert.pem"},
+					{keyfile, "priv/tls/key.pem"},
+					{sni_hosts, SniHosts}
+				], ProtocolOpts)
+		end.
+
+get_required_proplist_value(Name, Proplist) ->
```
martin-torhage: How do you feel about calling this function for `get_value` instead?

src/ar_http_iface_server.erl
```diff
@@ -60,8 +69,37 @@ do_start(Port) ->
...
+	case proplists:get_value(Name, Proplist) of
```
martin-torhage: This will break when the property value is `undefined`. You can use lists:keyfind/3 instead.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+			handle_serve_tx_1(ApexDomain, Hash, MaybeExt, Req)
 	end;
 
 %% @doc Catch case for requests made to unknown endpoints.
 %% Returns error code 400 - Request type not found.
-do_handle(_, _, Req) ->
+handle(_, _, Req) ->
+	not_found(Req).
+
+handle_serve_root(ApexDomain, Req) ->
```
martin-torhage: This function is only used when `apex_domain` is set. How about calling this `handle_gateway_root` instead?

martin-torhage:
I've added inline comments and here is some additional feedback:

- I didn't look through the code of `ar_base32` in detail yet but I noticed there are not tests. A unit test would make a lot of sense.
- `get_domain_labeling_1/1` is obviously not completed yet.
- Let's talk a about the ordering of the functions.
- Can a node making a request to a gateway node handle HTTPS only?

Here are my answers to your questions in the PR description:

1. We might want to add something with "apex" to the name of the apex cert files. E.g. move them to a subdir called "apex"
2. Yes, let's move the whole serve-one-tx-from-root-thingy to its own module.
3. It would be good if we can make it more clear that it's also enabling gateway mode. Maybe just rename it to `gateway_apex_domain`?
4. Sounds good enough to me.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
-		<< ".", Part/binary >> ->
-			do_handle(<<"GET">>, [<<"tx">>, Hash, <<"data.", Part/binary>>], Req);
-		<<>> ->
-			do_handle(<<"GET">>, [<<"tx">>, Hash, <<"data.html">>], Req);
-		_ ->
-			not_found(Req)
+handle(<<"GET">>, [<<Hash:43/binary, MaybeExt/binary>>], Req) ->
+	case ar_meta_db:get(apex_domain) of
+		not_found ->
+			handle(<<"GET">>, [<<"tx">>, Hash, <<"data.", MaybeExt/binary>>], Req);
```
ldmberman: Have you changed the behaviour here so that `GET /data.meh` returns HTML as if it was just `data` or `data.html` instead of `Request type not found`?

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+		ExpectedLabel -> handle_serve_tx_labeled_3(TxHash, MaybeExt, Req);
+		_ -> redirect_to_labeled_tx(ApexDomain, BlockHash, TxHash, Req)
+	end.
+
+derive_tx_label(TxHash, BlockHash) ->
+	State0 = crypto:hash_init(sha256),
+	State1 = crypto:hash_update(State0, ar_util:decode(TxHash)),
+	State2 = crypto:hash_update(State1, ar_util:decode(BlockHash)),
+	Digest = crypto:hash_final(State2),
+	binary:part(ar_base32:encode(Digest), {0, 12}).
```
ldmberman: I would like to question the need for `Base32`. It is supposed to be more readable but I personally do not feel the difference, at least on some examples I tried.

src/ar_http_iface_cowboy_handler.erl
ldmberman: Ah, disregard my comment, I now see the behaviour is preserved.

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -659,21 +665,139 @@ do_handle(Method, [<<"height">>], Req)
...
+
+handle_serve_tx_labeled_1(ApexDomain, Label, TxHash, MaybeExt, Req) ->
+	BlockHash = ?OK(get_tx_block_hash(TxHash)),
+	handle_serve_tx_labeled_2(ApexDomain, Label, BlockHash, TxHash, MaybeExt, Req).
+
+get_tx_block_hash(TxHash) ->
+	case ar_tx_search:get_tags_by_id(
+		whereis(http_search_node),
+		ar_util:decode(TxHash),
+		infinity
```
ldmberman: Will Cowboy timeout here? If not, we should. I think, makes sense to introduce a dedicated HTTP response, which would essentialy tell the user that the server is busy.

ldmberman:
Would it make sense to redirect

`https://gateway.example/vP_3_IZBng33AaW8MCfTqseJxdG-OWbfOb_ZU1UhR7s`

to

`https://xvc4dssncn5p.gateway.example`

instead of

`https://xvc4dssncn5p.gateway.example/vP_3_IZBng33AaW8MCfTqseJxdG-OWbfOb_ZU1UhR7s`?

src/ar_http_iface_cowboy_handler.erl
maxmellen: The reason for base32 is that domain labels (the parts separated by dots) are case insensitive and have some limitations such as having to start and end with an alphanumeric character.

maxmellen:
@ldmberman Generating the label is a one way process, so if we got rid of the path and only kept the label, we wouldn't have enough information to recover the original transaction ID.

src/ar_http_iface_server.erl
maxmellen: I also thought about this, but I don't think it's a problem

maxmellen: Otherwise, I think `prop lists:lookup/2` is a better match. I can update the function if you want, but in this case considering `undefined` as a specified value would only crash later on.

src/ar.erl
maxmellen: So I'm a bit unsure about calling anything a _gateway mode_ as of yet. I also wanted to be more explicit and I was thinking of having another flag [`tx_origins`: Serve transaction contents on unique origins] to enable the behaviour in this PR, and having a bit of validation code fail early if TLS isn't enabled or an apex domain isn't specified. I also wonder if we should just enable TLS implicitly when that feature is enabled, but then again, could it make sense to run a node under HTTPS without this origins logic? Would other nodes be able to communicate with it?

ldmberman:
@maxmellen yeah, I see the problem. Perhaps, later (not in this PR) if we want to do it, we may maintain a reverse mapping somehow.

src/ar_http_iface_cowboy_handler.erl
ldmberman: Fair.

src/ar_http_iface_cowboy_handler.erl
maxmellen: Is the naming convention to call the base64url encoded block hash `BH` or its binary form?

src/ar_http_iface_cowboy_handler.erl
maxmellen: Nope 😅 I think I used another split somewhere else while developing this feature and when I wrote the new split, I realised that both the `string` module and the `binary` module have their own split functions. Looking into the difference, I saw that `string:split` operates on `unicode:chardata()`, whereas `binary:split` is a BIF that only operates on binaries. Therefore I went for the function in `binary` and updated the code in this line to be consistent. I think I later got rid of the new `split` call.

src/ar_http_iface_cowboy_handler.erl
maxmellen: So I implemented a handler-wide timeout of 30 seconds here https://github.com/ArweaveTeam/arweave/pull/144/commits/db91056e355911c600b4858d15f2d4447985fe23. I think it still makes sense to pick smaller values for this line and for the status endpoint. What do you think?

src/ar_gateway_h.erl
```diff
@@ -0,0 +1,140 @@
+-module(ar_gateway_h).
+-behaviour(cowboy_handler).
+-export([init/2]).
+
+-include("ar.hrl").
+-define(BH_SEARCH_TIMEOUT, 5000).
```
ldmberman: I think, we have decided to only have a global timeout.

maxmellen: Considering I wrote the whole handler, I feel pretty confident that `ar_tx_search:get_tags_by_id` is the only call that can take some time before returning. Therefore, I don't see the point in having a global timeout, although I could add it if you think it still makes sense here.

ldmberman: Ah, sorry, we do actually have a global timeout in the non-gateway server now.

Yes, it's probably not useful to have a global timeout in the gateway server at the moment. In general I think it is a way (although crappy) to have at least some backoff in case one of the endpoints is overloaded for whatever weird reason.

src/ar_http_iface_h.erl
```diff
@@ -8,22 +8,31 @@
...
+		Pid ! handle(InitialReq)
+	end),
+	receive
+		{Status, Headers, Body, HandledReq} ->
+			CowboyStatus = handle208(Status),
+			RepliedReq = cowboy_req:reply(CowboyStatus, Headers, Body, HandledReq),
+			{ok, RepliedReq, State}
+	after
+		30000 ->
+			error(handler_timeout)
```
martin-torhage: This is expected to happen, so I don't think we should call `error/1` and cause an error log entry. Instead I think we should reply with a 500 and call `ar:info` or `ar:warn`.

src/ar_gateway_server.erl
```diff
@@ -0,0 +1,36 @@
...
+-export([start/3]).
+
+%%%===================================================================
+%%% Public interface.
+%%%===================================================================
+
+%% @doc Start the server that will handle gateway requests.
+start(Port, Domain, CustomDomains) ->
+	ProtocolOpts = #{
+		middlewares => [cowboy_handler],
```
martin-torhage: Shoudn't we have the `ar_firewall` here?

src/ar_gateway_h.erl
martin-torhage: Only waiting 5s sounds too short for me. I have successfully waited more than 5s for a webpage to load many times.

src/ar.erl
```diff
@@ -202,6 +206,10 @@ parse(["enable", Feature | Rest ], O = #opts { enable = Enabled }) ->
 	parse(Rest, O#opts { enable = [ list_to_atom(Feature) | Enabled ] });
 parse(["disable", Feature | Rest ], O = #opts { disable = Disabled }) ->
 	parse(Rest, O#opts { disable = [ list_to_atom(Feature) | Disabled ] });
+parse(["gateway", Port, Domain | Rest ], O = #opts { gateway = off }) ->
+	parse(Rest, O#opts { gateway = {on, Port, Domain} });
```
martin-torhage: The `on` atom looks unnecessary to me, but if you see value in it, then I don't mind keeping it.

src/ar.erl
```diff
@@ -398,13 +409,20 @@ start(
...
-		Bridge
-	),
+	ar_http_iface_server:start(Port, [
+		{http_entrypoint_node, Node},
+		{http_search_node, SearchNode},
+		{http_bridge_node, Bridge}
+	]),
+	case GatewayOpts of
+		{on, GatewayPortString, GatewayDomainString} ->
+			GatewayPort = list_to_integer(GatewayPortString),
```
martin-torhage: Wouldn't it make sense if the `parse` function also did `list_to_integer` on `GatewayPort` and `list_to_binary ` on `GatewayDomain` and `GatewayCustomDomains`?

src/ar_gateway_server.erl
maxmellen: So as agreed not for now

src/ar_gateway_h.erl
maxmellen: So let's go for `25000` 👍

src/ar.erl
maxmellen: Yep 👍

src/ar_http_iface_h.erl
maxmellen: Will do 👍

src/ar_gateway_h.erl
```diff
@@ -84,6 +58,50 @@ handle_labeled_request_2(Domain, Label, TXID, Filename, BH, Req) ->
 		OtherLabel -> redirect_to_labeled_tx(Domain, OtherLabel, TXID, Req)
 	end.
 
+handle_custom_request(Domain, CustomDomain, Req) ->
```
ldmberman: The amount of details accounted for in this module is not trivial, I think it would be very helpful to add a generic comment to each of `handle_custom_request`, `handle_labeled_request` and `handle_apex_request`. For `handle_custom_request` I see something along the lines of:

```
handle_custom_domain ensures that:

- "mycustomdomain.site" serves the transaction specified in its txt record.
  It has to be passed on  upon startup as "custom_domain".
- "mycustomdomain.site/[TXID]" redirects to "[TXPREFIX].gateway.com/[TXID]" where "gateway.com" is the domain mycustomdomain.site points to.
```

src/ar_domain.erl
```diff
@@ -0,0 +1,49 @@
...
+%%%===================================================================
+%%% Private functions.
+%%%===================================================================
+
+get_labeling_1(ApexDomain, CustomDomains, Hostname) ->
+	case lists:member(Hostname, CustomDomains) of
+		true ->
+			{custom, Hostname};
+		false ->
+			error({
```
martin-torhage: I think this will happen quite often. I would rather output an explaining error message as a 400, 404 and not cause an ERROR entry in the log. INFO would be ok.

src/ar_http_iface_h.erl
```diff
@@ -17,13 +18,28 @@ init(InitialReq, State) ->
 %%% Private functions.
 %%%===================================================================
 
-do_handle(Req) ->
+%% @doc Because cowboy 2.0 seemingly doesn't support accessing the request from
+%% multiple processes (see https://github.com/ninenines/cowboy/issues/1374) the
+%% following is the only way to introduce a asynchronous handler timeout.
+handler_timeout_after(Req, T) ->
```
martin-torhage: Is it maybe supposed to be called `handle_timeout_after`?

src/ar_http_iface_h.erl
```diff
@@ -17,13 +18,28 @@ init(InitialReq, State) ->
...
+	HandlerPid = self(),
+	spawn_link(fun() ->
+		timer:sleep(T),
+		ar:warn([
```
martin-torhage: It looks like this will be triggered for every request.

src/ar_http_iface_h.erl
```diff
@@ -17,13 +18,28 @@ init(InitialReq, State) ->
...
+			handler_timeout,
+			{method, cowboy_req:method(Req)},
+			{path, cowboy_req:path(Req)}
+		]),
+		exit(HandlerPid, kill)
```
martin-torhage: `kill` is not a good reason. Something with "timeout" would be more meaningful.

ldmberman: `kill` is actually the system atom required to kill the process which traps exits.

src/ar_http_iface_h.erl
```diff
@@ -17,13 +18,28 @@ init(InitialReq, State) ->
...
 %%%===================================================================
 
-do_handle(Req) ->
+%% @doc Because cowboy 2.0 seemingly doesn't support accessing the request from
+%% multiple processes (see https://github.com/ninenines/cowboy/issues/1374) the
+%% following is the only way to introduce a asynchronous handler timeout.
+handler_timeout_after(Req, T) ->
+	HandlerPid = self(),
+	spawn_link(fun() ->
+		timer:sleep(T),
```
martin-torhage: One timer per request living for 30s. This is wasting resources. The timer should at least be canceled when not needed anymore, i.e. when the request finishes.

src/ar_http_iface_h.erl
ldmberman: It depends on whether cowboy kills the handler process or not, but worth checking.

src/ar.erl
```diff
@@ -111,7 +113,7 @@ main("") ->
 		end,
 		[
 			{"peer (ip:port)", "Join a network on a peer (or set of peers)."},
-			{"start_hash_list (file)", "Start the node from a given block."},
+			{"start_hash_list (hash)", "Start the node from a given block."},
```
martin-torhage: This doc is not correct anymore. The hash is a hash of the block hash list. I think the argument name ("start_hash_list") is confusing as well, but ok.

src/ar.erl
```diff
@@ -183,7 +187,7 @@ parse(["disk_space", Size|Rest], O) ->
 parse(["load_mining_key", File|Rest], O)->
 	parse(Rest, O#opts { load_key = File });
 parse(["start_hash_list", IndepHash|Rest], O)->
-	parse(Rest, O#opts { start_hash_list = ar_util:decode(IndepHash) });
+	parse(Rest, O#opts { start_hash_list = IndepHash });
```
martin-torhage: `IndepHash` is a hash of the block hash list, so calling it `BHLHash` as later in the code would make sense. Also, the `ar_util:decode/1` should probably still be called here since this is the parsing function.

src/ar_util.erl
```diff
@@ -40,13 +40,26 @@ pick_random(Xs) ->
 encode(Bin) ->
 	base64url:encode(Bin).
 
-%% @doc Decode a URL safe base64 to binary.
+%% @doc Try to decode a URL safe base64 into a binary or throw an error when
+%% invalid.
 decode([]) -> [];
-decode(Bin) when is_list(Bin) ->
-	decode(list_to_binary(Bin));
+decode(List) when is_list(List) ->
```
martin-torhage: Tiny naming suggestion: Let's call it `String` instead of `List`.

src/ar_util.erl
```diff
@@ -40,13 +40,26 @@ pick_random(Xs) ->
 encode(Bin) ->
 	base64url:encode(Bin).
 
-%% @doc Decode a URL safe base64 to binary.
+%% @doc Try to decode a URL safe base64 into a binary or throw an error when
+%% invalid.
 decode([]) -> [];
```
martin-torhage: This is old code, I know, but my god how unexpected to get a string back (instead of a binary) if the argument is an empty string.

src/ar_util.erl
```diff
@@ -40,13 +40,26 @@ pick_random(Xs) ->
...
-decode(Bin) when is_list(Bin) ->
-	decode(list_to_binary(Bin));
+decode(List) when is_list(List) ->
+	decode(list_to_binary(List));
+decode(<<>>) -> <<>>;
```
martin-torhage: I see no need of adding this signature. `base64url:decode/1` can handle empty binary.

src/ar_gateway_h.erl
```diff
@@ -0,0 +1,207 @@
...
+handle_labeled_request_2(Domain, Label, TXID, Filename, BH, Req) ->
+	case ar_domain:derive_tx_label(TXID, BH) of
+		Label -> serve_tx(Filename, Req);
+		OtherLabel -> redirect_to_labeled_tx(Domain, OtherLabel, TXID, Req)
+	end.
+
+handle_custom_request(Domain, CustomDomain, Req) ->
+	case resolve_path(cowboy_req:path(Req)) of
+		root -> handle_custom_request_1(CustomDomain, Req);
+		{tx, TXID, Filename} -> derive_label_and_redirect(Domain, TXID, Req);
```
martin-torhage: `Filename` is not used

src/ar_http_iface_cowboy_handler.erl
martin-torhage: ✅

src/ar_http_iface_cowboy_handler.erl
martin-torhage: ✅

src/ar_util.erl
maxmellen: Are you suggesting we change that behaviour?

src/ar_util.erl
maxmellen: I'd rather not but if you feel strongly about this, can do.

src/ar_domain.erl
```diff
@@ -0,0 +1,44 @@
...
+derive_tx_label(TXID, BH) ->
+	Data = <<TXID/binary, BH/binary>>,
+	Digest = crypto:hash(sha256, Data),
+	binary:part(ar_base32:encode(Digest), {0, 12}).
+
+%%%===================================================================
+%%% Private functions.
+%%%===================================================================
+
+get_labeling_1(ApexDomain, CustomDomains, Hostname) ->
```
martin-torhage: `ApexDomain` is not used. Should we also change the function name somehow by removing the argument?

src/ar_http_iface_h.erl
```diff
@@ -8,22 +8,43 @@
...
+		receive
+			cancel -> ok
+		after
+			T ->
+				ar:warn([
+					handler_timeout,
+					{method, cowboy_req:method(Req)},
+					{path, cowboy_req:path(Req)}
+				]),
+				exit(HandlerPid, kill)
```
martin-torhage: It seems like the HTTP request hangs when this happens. I would expect Cowboy to either return a 500 or at least cut the connection. I don't think this is ok.

PR #146, opened 2019-05-03
Added optional redirecting of browsers to other nodes when data is...

samcamwilliams:
...not available locally, using a Pareto-distributed random node selection scheme.

### DIFF
--- a/src/ar_http_iface_cowboy_handler.erl
+++ b/src/ar_http_iface_cowboy_handler.erl
@@ -185,12 +185,13 @@ do_handle(<<"POST">>, [<<"arql">>], Req) ->
 
 %% @doc Return the data field of the transaction specified via the transaction ID (hash) served as HTML.
 %% GET request to endpoint /tx/{hash}/data.html
-do_handle(<<"GET">>, [<<"tx">>, Hash, << "data.", _/binary >>], Req) ->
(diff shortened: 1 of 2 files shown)


src/ar_util.erl
```diff
@@ -36,6 +36,22 @@ pick_random(List, N) ->
 pick_random(Xs) ->
 	lists:nth(rand:uniform(length(Xs)), Xs).
 
+%% @doc Pick a random element, on a `1/(1+rank)` distribution.
```
ldmberman: I would clarify here ".. where rank 1 is assigned to the best peer, rank 2 to the second best peer, etc".

src/ar_http_iface_cowboy_handler.erl
```diff
@@ -185,12 +185,13 @@ do_handle(<<"POST">>, [<<"arql">>], Req) ->
 
 %% @doc Return the data field of the transaction specified via the transaction ID (hash) served as HTML.
 %% GET request to endpoint /tx/{hash}/data.html
-do_handle(<<"GET">>, [<<"tx">>, Hash, << "data.", _/binary >>], Req) ->
+do_handle(<<"GET">>, Addr = [<<"tx">>, Hash, << "data.", _/binary >>], Req) ->
```
ldmberman: Starting from #144 I think it should be only the functionality of the gateway server so it probably makes sense to wait until it's merged and move it there.

PR #147, opened 2019-05-14
Refactor ar_node_utils:calculate_proportion/3.

llaisdy:
This refactor does not affect behaviour (same outputs for same inputs),
but the new version does not do work that is not used.

### DIFF
--- a/src/ar_node_utils.erl
+++ b/src/ar_node_utils.erl
@@ -135,42 +135,15 @@ calculate_reward_pool(OldPool, TXs, _RewardAddr, Proportion) ->
 %% @doc Calculates the portion of the rewardpool that the miner is entitled
 %% to for mining a block with a given recall. The proportion is based on the
 %% size of the recall block and the average data stored within the weave.
-calculate_proportion(RecallSize, WeaveSize, Height) when (Height == 0)->
-	% Genesis block.
-	calculate_proportion(
-		RecallSize,
-		WeaveSize,
-		1
-	);
-calculate_proportion(RecallSize, WeaveSize, Height) when (WeaveSize == 0)->
-	% No data stored in the weave.
-	calculate_proportion(
-		RecallSize,
-		1,
-		Height
-	);
-calculate_proportion(RecallSize, WeaveSize, Height) when RecallSize >= (WeaveSize/Height) ->
-	% Recall size is larger than the average data stored per block.
-	XRaw = ((Height * RecallSize) / WeaveSize) - 1,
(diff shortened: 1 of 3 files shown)


llaisdy:
I'll re-add the explanatory comments in the function branches.

src/ar_node_utils.erl
```diff
@@ -135,42 +135,15 @@ calculate_reward_pool(OldPool, TXs, _RewardAddr, Proportion) ->
...
-		(math:pow(2, X) / (math:pow(2, X) + 2))
-	);
-calculate_proportion(RecallSize, WeaveSize, Height) when RecallSize == 0 ->
-	% Recall block has no data txs, hence size of zero.
-	calculate_proportion(
-		1,
-		WeaveSize,
-		Height
-	);
+calculate_proportion(0, W, H) -> calculate_proportion(1, W, H);
```
ldmberman: Why do we replace 0 recall size and weave size with 1? Shouldn't we in this case just return a minimal proportion right away?

Also, instead of replacing height 0 with 1 we can just divide by `H + 1` - I think, it is clearer this way.

src/ar_node_utils.erl
```diff
@@ -135,42 +135,15 @@ calculate_reward_pool(OldPool, TXs, _RewardAddr, Proportion) ->
...
+calculate_proportion(R, 0, H) -> calculate_proportion(R, 1, H);
+calculate_proportion(R, W, 0) -> calculate_proportion(R, W, 1);
+calculate_proportion(R, W, H) when R < (W/H) -> 0.1;
 calculate_proportion(RecallSize, WeaveSize, Height) ->
```
ldmberman: It might be a convenient occasion to improve this function by doing either/all of:

- add some clarifications of the formula, why it is like it is;
- give the function a better name, perhaps, `recall_block_reward_proportion`?
- replace 0.1 with a macro;
- add a test.

src/ar_node_utils.erl
llaisdy: Disagree with your second point.

Agree with your first point.  Those first three branches could all return 0.1.  That will preserve behaviour (I think).

src/ar_node_utils.erl
llaisdy: re 3 & 4: good idea. Will replace 0.1 with a macro and add eunit tests.
re 1: agree.
re 2: depends on 1.

llaisdy:
Now using macro, and with tests.

PR #148, opened 2019-05-17
Add support for configuration files

maxmellen:
This PR adds support for configuration files when starting up a node.

The configuration file is specified with the `config_file` cli argument like so:
```shell
./arweave-server config_file <file>
```

Any cli arguments passed in addition to `config_file` will take precedence over the configuration file.

The syntax of the configuration file aims to be as close to the cli args as possible. Here is an example from the test fixtures.
```json
{
  "peers": [
    "188.166.200.45",
    "188.166.192.169",
    "163.47.11.64",
    "159.203.158.108",
    "159.203.49.13",
    "139.59.51.59",
    "138.197.232.192",
    "46.101.67.172"
  ],
  "start_hash_list": "ZFTD3ne7_U3BONHi8O-QpBv0ZQTAXJ2eFIih8dod0f8",
  "mine": true,
  "port": 1985,
  "data_dir": "some_data_dir",
  "polling": true,
  "clean": true,
  "no_auto_join": true,
  "init": true,
  "diff": 42,
  "mining_addr": "LKC84RnISouGUw4uMQGCpPS9yDC-tIoqM2UVbUIt-Sw",
  "max_miners": 43,
  "new_mining_key": true,
  "load_mining_key": "some_key_file",
  "content_policies": [
    "some_content_policy_1",
    "some_content_policy_2"
  ],
  "transaction_blacklists": [
    "some_blacklist_1",
    "some_blacklist_2"
  ],
  "disk_space": 44,
  "benchmark": true,
  "auto_update": "V_zW1A2HeqWFBpwKCjyd8V6eI8S3ia8GOVA6g7YXAdU",
  "internal_api_secret": "some_very_very_long_secret",
  "enable": [
    "feature_1",
    "feature_2"
  ],
  "disable": [
    "feature_3",
    "feature_4"
  ],
  "gateway": {
    "port": 443,
    "domain": "gateway.localhost"
  },
  "custom_domains": [
    "domain1.example",
    "domain2.example"
  ]
}
```


src/ar.erl
```diff
@@ -151,67 +122,81 @@ main("") ->
...
+	Args = lists:reverse(Skipped),
+	parse_cli_args(Args, Opts).
+
+read_config_from_file(Path) ->
+	case file:read_file(Path) of
+		{ok, FileData} -> ar_config:parse(FileData);
+		{error, _} -> {error, file_unreadable, Path}
+	end.
+
+parse_cli_args([], O) -> O;
```
martin-torhage: Since we have a new `ar_config` module (which is good thing), why don't we have this long function there? It would make sense to me.

src/ar_config.hrl
```diff
@@ -0,0 +1,29 @@
+%% Start options with default values.
+-record(opts, {
```
martin-torhage: Seems like the term `config` is very well established by this branch. Wouldn't it make sense to rename this record to something with `config`? Maybe just `config`.

src/ar_config.erl
```diff
@@ -0,0 +1,171 @@
...
+-export([parse/1]).
+-include("ar.hrl").
+-include("ar_config.hrl").
+
+%%%===================================================================
+%%% Public interface.
+%%%===================================================================
+
+parse(Config) when is_binary(Config) ->
+	case ar_serialize:json_decode(Config) of
```
martin-torhage: The jiffy data structure jiffy uses is just legacy. Using lists and maps would make more sense. Now the jiffy stuff is mixed into the `parse_options` functions. How do you feel about translating the `ar_serialize:json_decode/1` return to maps and lists first?

src/ar_config.erl
```diff
@@ -0,0 +1,171 @@
...
+
+parse_options({KVPairs}) when is_list(KVPairs) ->
+	parse_options(KVPairs, #opts{});
+parse_options(JsonValue) ->
+	{error, root_not_object, JsonValue}.
+
+parse_options([{_, null} | Rest], Opts) ->
+	parse_options(Rest, Opts);
+parse_options([{<<"config_file">>, _} | _], _) ->
+	{error, config_file_set};
```
martin-torhage: I can imagine a user getting confused when creating a config file for the first time if he get the error "config_file_set". Can we explain this a little bit better? E.g. something like "config_file in the config file is not allowed"

src/ar_config.erl
```diff
@@ -0,0 +1,171 @@
...
+parse_options([{<<"peers">>, Peers} | Rest], Opts) when is_list(Peers) ->
+	ParsedPeers = lists:map(fun(Peer) -> ar_util:parse_peer(Peer) end, Peers),
+	parse_options(Rest, Opts#opts { peers = ParsedPeers });
+parse_options([{<<"peers">>, Peers} | _], _) ->
+	{error, {bad_type, peers, array}, Peers};
+parse_options([{<<"start_hash_list">>, Hash} | Rest], Opts) when is_binary(Hash) ->
+	parse_options(Rest, Opts#opts { start_hash_list = ar_util:decode(Hash) });
+parse_options([{<<"start_hash_list">>, Hash} | _], _) ->
+	{error, {bad_type, start_hash_list, string}, Hash};
```
martin-torhage: What's the point checking for some errors when the `ar_util:decode/1` will crash if the value is not in a valid format? Also, being this specific explaining that it's `bad_type` and something about `string` (which btw is not true - it can be anything but a binary) is not really needed.

I'm suggesting this:
Create a function that parses one argument (e.g. "start_hash_list" and the hash). It can return an error tuple for specific error handling. If it crashes, then `parse/1` can return something like `{error, {bad_config, "start_hash_list"}}`

martin-torhage:
I finished the review and added comments inline. Most of the feedback is for issues we already have in master. Nothing is a show stopper, but since this PR is not blocking anything else, I'm thinking this is a good time to just quickly improve it and be done with it.

src/ar_config.erl
maxmellen: My thinking while writing this function was that it's caller should turn the atoms into user-friendly messages. Although we could do this now, if someone ran into an error as things are now, we could easily understand what's wrong.

src/ar_config.erl
maxmellen: So the types I'm referring to here (such as `string`) are the expected JSON types. My thinking was again that the caller of this function could turn the error tuple into a readable error message such as "start_hash_list should be a string".

maxmellen:
I'll see if it's not too much of a pain to change `json_decode` to return a map and I'll add some code to give more helpful error messages.

PR #150, opened 2019-06-04
Make request limit per IP configurable and increase the default

ldmberman:
The change is suggested by the nodes usage.

### DIFF
--- a/src/ar.erl
+++ b/src/ar.erl
@@ -118,7 +118,8 @@ main("") ->
 			{"enable (feature)", "Enable a specific (normally disabled) feature. For example, subfield_queries."},
 			{"disable (feature)", "Disable a specific (normally enabled) feature. For example, api_compat mode."},
 			{"gateway (port) (domain)", "Run a gateway on the specified port and domain"},
-			{"custom_domain (domain)", "Add a domain to the list of supported custom domains."}
+			{"custom_domain (domain)", "Add a domain to the list of supported custom domains."},
+			{"requests_per_minute_limit (number)", "Limit the maximum allowed number of HTTP requests per IP address per minute. Default is 900."}
 		]
 	),
 	erlang:halt();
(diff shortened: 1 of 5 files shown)


src/ar.erl
```diff
@@ -118,7 +118,8 @@ main("") ->
 			{"enable (feature)", "Enable a specific (normally disabled) feature. For example, subfield_queries."},
 			{"disable (feature)", "Disable a specific (normally enabled) feature. For example, api_compat mode."},
 			{"gateway (port) (domain)", "Run a gateway on the specified port and domain"},
-			{"custom_domain (domain)", "Add a domain to the list of supported custom domains."}
+			{"custom_domain (domain)", "Add a domain to the list of supported custom domains."},
+			{"accept_requests_per_minute_limit (number)", "Limit the number of accepted requests per minute per IP. Default is 2000."}
```
martin-torhage: I'm suggesting calling this for "requests_per_minute_limit" instead and changing the description to:

Limit the number of HTTP requests allowed per IP address per minute. Default is 2000.

src/ar.hrl
```diff
@@ -141,8 +141,8 @@
 %% @doc Never have more than this number of peers (new peers excluded).
 -define(MAXIMUM_PEERS, 20).
 
-%% Maximum number of requests allowed by an IP in any 30 second period.
--define(MAX_REQUESTS, 450).
+%% Maximum allowed number of accepted requests per minute per IP.
+-define(DEFAULT_ACCEPT_REQUESTS_PER_MINUTE_LIMIT, 2000).
```
martin-torhage: I'm not comfortable with more than doubling this limit. Where did 2000 come from? Already 450 sounds really high to me.

src/ar.erl
ldmberman: I wanted to distinguish it this way from the outbound requests, but the context is hopefully clear enough so I think it's fine to go with your suggestion.

src/ar.hrl
ldmberman: The important bit here is to make it configurable so that we can use it in our gateway so we agreed to not change the default for now, will fix it up.

PR #152, opened 2019-07-05
Allow access to all regular node endpoints from gateway

maxmellen:
After merging the new gateway middleware, the routing table for our endpoints will look something like this.

| Endpoint                                       | Action                                                  |
| ---------------------------------------------- | ------------------------------------------------------- |
| `http://1.2.3.4:1984/`                         | Old info endpoint accessed normally                     |
| `http://1.2.3.4:1984/{txid}`                   | Tx content served as per old endpoint                   |
| `http://1.2.3.4:1984/*`                        | Old API endpoints accessible normally                   |
| `http://gateway.example/`                      | Redirect to `https://gateway.example/`                  |
| `http://gateway.example/{txid}`                | Redirect to `https://{txlabel}.gateway.example/{txid}`  |
| `http://gateway.example/*`                     | Redirect to `https://gateway.example/*`                 |
| `http://custom.domain.example/`                | Redirects to `https://custom.domain.example/`           |
| `http://custom.domain.example/{txid}`          | Redirects to `https://{txlabel}.gateway.example/{txid}` |
| `http://custom.domain.example/*`               | Redirects to `https://custom.domain.example/*`          |
| `https://gateway.example/`                     | Bad request (to be implemented)                         |
| `https://gateway.example/{txid}`               | Redirect to `https://{txlabel}.gateway.example/{txid}`  |
| `https://gateway.example/*`                    | Old API endpoints accessible normally                   |
| `https://{txlabel}.gateway.example/`           | Bad request (to be implemented)                         |
| `https://{txlabel}.gateway.example/{txid}`     | Tx content served by the gateway middleware             |
| `https://{txlabel}.gateway.example/*`          | Old API endpoints accessible normally                   |
| `https://custom.domain.example/`               | Custom domain tx served by the gateway middleware       |
| `https://custom.domain.example/{txid}`         | Redirect to `https://{txlabel}.gateway.example/{txid}`  |
| `https://custom.domain.example/*`              | Old API endpoints accessible normally                   |

These examples use the apex domain name `gateway.example` and a custom domain name `custom.domain.example`.


src/ar.hrl
```diff
@@ -1,3 +1,6 @@
+-ifndef(AR_HRL).
+-define(AR_HRL, true).
```
ldmberman: What is it defined for?

src/ar.erl
```diff
@@ -236,13 +236,22 @@ start(
...
-		custom_domains = GatewayCustomDomains,
+		gateway_domain = GatewayDomain,
+		gateway_custom_domains = GatewayCustomDomains,
 		requests_per_minute_limit = RequestsPerMinuteLimit
 	}) ->
 	%% Start the logging system.
 	error_logger:logfile({open, Filename = generate_logfile_name()}),
 	error_logger:tty(false),
+	%% Verify port collisions when gateway enabled
+	GatewayOpts = gateway_opts(GatewayDomain, GatewayCustomDomains),
```
ldmberman: Looks like `gateway_opts` are not used anywhere else, could we perhaps just check if `GatewayDomain` is set here?

src/ar.hrl
ldmberman: Ah, I see, you are avoiding redefinitions under some occasions.

ldmberman:
`ar_config_tests: parse_test_` fails for me for some reason.

src/ar.erl
maxmellen: I'm not sure I get your feedback, the `gateway_opts` helper function is just used once here to avoid inlining more login into the already huge start function, although I could inline it if you insist, but the `GatewayOpts` variable is also used later below

maxmellen:
@ldmberman Fixed `parse_test_`

src/ar.erl
ldmberman: May be it's cleaner if we keep it like this, let's keep it.

src/ar_gateway_middleware.erl
```diff
@@ -94,31 +95,47 @@ handle_labeled_request(Domain, Label, Req) ->
...
+	case cowboy_req:scheme(Req) of
+		<<"https">> -> not_found(Req);
+		<<"http">> -> other_request(Req, Env)
+	end.
+
 resolve_path(Path) ->
 	case Path of
 		<<"/">> -> root;
-		<<"/", Hash/binary>> -> resolve_path_1(Hash);
+		<<"/", Hash:43/bytes>> -> resolve_path_1(Hash);
```
ldmberman: There could be a valid path, like `/whatever/whateverelse/...`, which could happen to be 43 bytes long. Should we instead return `other` if we fail to recognize a TX ID?

ldmberman:
I tried to run a polling gateway session - I am getting this when I am trying to access localhost:1984:

```Error in process <0.834.0> with exit value:
{badarg,[{binary,part,[<<>>,{-1,1}],[]},
         {ar_http_iface_middleware,remove_trailing_slash,1,
                                   [{file,"src/ar_http_iface_middleware.erl"},
                                    {line,737}]},
         {ar_http_iface_middleware,parse_path,1,
                                   [{file,"src/ar_http_iface_middleware.erl"},
                                    {line,733}]},
         {ar_http_iface_middleware,handle,3,
                                   [{file,"src/ar_http_iface_middleware.erl"},
                                    {line,64}]},
         {ar_http_iface_middleware,'-execute/2-fun-0-',2,
                                   [{file,"src/ar_http_iface_middleware.erl"},
                                    {line,14}]}]}
```

src/ar_gateway_middleware.erl
maxmellen: Good catch! Fixed in 75e950b

maxmellen:
@ldmberman Fixed accessing `/` on the normal (1984) port in 2cc35ea

ldmberman:
It does not redirect me to `https://{txlabel}.gateway.localhost/{txid}` if there is a trailing slash, like `https://gateway.localhost/{txid}/`.

ldmberman:
It's interesting how it reacts to spam now. By running `for i in {1..1000}; do curl https://gateway.localhost/info --insecure & done` I could reproduce a typical `Too many requests`, but before it started to occur, curl reported a whole bunch of `Could not resolve host: gateway.localhost`,
`LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to gateway.localhost:443 `, and
`Failed to connect to gateway.localhost port 443: Connection reset by peer`. Not sure if it's bad though.

src/ar_http_iface_middleware.erl
```diff
@@ -1,23 +1,25 @@
...
+-export([execute/2, read_complete_body/2]).
 -include("ar.hrl").
 -define(HANDLER_TIMEOUT, 55000).
 
 %%%===================================================================
 %%% Cowboy handler callbacks.
 %%%===================================================================
 
-init(InitialReq, State) ->
+execute(Req, #{ handler := ar_http_iface_handler }) ->
```
ldmberman: We've discussed it might be helpful to add a comment about this part.

maxmellen: Comment added in bad9fd6a9aa74e9487b139558b1e219b92389a07

maxmellen:
@ldmberman So I've reverted the logic over how we look at paths to only consider them as a list of segments in 03c550c48ef823ee3e60c60cf2ef8178eeae7c96 and the behaviour now seems to be consistent across the gateway middleware and the node api middleware.

PR #155, opened 2019-07-12
 Order ArQL results by block height

maxmellen:
This PR also takes care of a few extra things:

- Returning an empty JSON array rather than an empty response when no results are found
- Adds an index on the `tx` field of the `arql_tag` mnesia table
- Removes the `ar_tx_search` server in favor of mnesia non-_dirty_ operations

### DIFF
--- a/src/ar_http_iface_middleware.erl
+++ b/src/ar_http_iface_middleware.erl
@@ -205,8 +205,11 @@ handle(<<"POST">>, [<<"arql">>], Req, Pid) ->
 		{ok, QueryJson, ReadReq} ->
 			case ar_serialize:json_struct_to_query(QueryJson) of
 				{ok, Query} ->
-					TXs = ar_util:unique(ar_parser:eval(Query)),
-					Body = ar_serialize:jsonify(ar_serialize:hash_list_to_json_struct(TXs)),
+					TXIDs = ar_util:unique(ar_parser:eval(Query)),
+					TXBlockHeightPairs = lists:map(fun ar_tx_search:get_tx_block_height/1, TXIDs),
+					SortedPairs = lists:sort(fun(A, B) -> A >= B end, TXBlockHeightPairs),
+					SortedTXs = [TXID || {_, TXID} <- SortedPairs],
+					Body = ar_serialize:jsonify(ar_serialize:hash_list_to_json_struct(SortedTXs)),
 					{200, #{}, Body, ReadReq};
 				{error, _} ->
 					{400, #{}, <<"Invalid ARQL query.">>, ReadReq}
--- a/src/ar_http_iface_middleware.erl
+++ b/src/ar_http_iface_middleware.erl
@@ -14,10 +14,12 @@
 %% middlewares. It uses the `handler` env value set by cowboy_router
 %% to determine whether or not it should run, otherwise it lets
 %% the cowboy_handler middleware run prometheus_cowboy2_handler.
-execute(Req, #{ handler := ar_http_iface_handler }) ->
+execute(Req, #{ handler := ar_http_iface_handler } = Env) ->
 	Pid = self(),
+	Req1 = with_pid_req_field(Req, Pid),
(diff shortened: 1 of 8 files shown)


src/ar_tx_search.erl
```diff
@@ -34,71 +35,79 @@
 %% @doc Start a search node.
 start() ->
 	initDB(),
-	PID = spawn(fun server/0),
-	erlang:register(http_search_node, PID),
-	{ok, PID}.
+	ok.
```
ldmberman: The module documentation (in the beginning of the file) needs an update.

Specifically, these parts should be removed:

```
%%% Access to the DB is wrapped in a server, which limits the number of
%%% concurrent requests to 1.
```

and

```
%%% Warning! Writes returns early but are serialized together with reads. If a
%%% write fails it will be logged, but the queue of reads and writes will
%%% continue.
```

src/ar_http_iface_middleware.erl
```diff
@@ -211,19 +205,11 @@ handle(<<"POST">>, [<<"arql">>], Req, Pid) ->
...
-								#{},
-								ar_serialize:jsonify(
-									ar_serialize:hash_list_to_json_struct(Set)
-								),
-								ReadReq
-							}
-					end;
+					TXIDs = ar_util:unique(ar_parser:eval(Query)),
+					TXBlockHeightPairs = lists:map(fun ar_tx_search:get_tx_block_height/1, TXIDs),
+					SortedTXs = [TXID || {_, TXID} <- lists:sort(TXBlockHeightPairs)],
```
ldmberman: I wonder if it makes more sense from the UX perspective to sort in the descending order so that the newest items are displayed on top, like we do in `GET /wallet/[Addr]/txs`, what do you think?

ldmberman:
We should use `ar_semaphore` in the ArQL-based endpoints now when we are dropping the access serialisation in `ar_tx_search`. I think, we should have a command argument with a reasonable conservative default for miners that can be overwritten on gateways.

maxmellen:
> We should use ar_semaphore in the ArQL-based endpoints now when we are dropping the access serialisation in ar_tx_search. I think, we should have a command argument with a reasonable conservative default for miners that can be overwritten on gateways.

@ldmberman What would you set the `ar_semaphore` capacity to? I think it's fine if the limit also applies to the gateway for now.

src/ar_http_iface_middleware.erl
maxmellen: Makes sense. We should definitely be consistent and I think we'll tend to search for recent data so I'll update the function

ldmberman:
It's hard to predict the numbers we are going to get, but clearly, the miners should not be cheaply spammed with requests while gateways, on the other hand, should be able to serve as many requests as possible. I would feel comfortable with something around 10/1000 (miners/gateways), although I might be overestimating the danger. What's your intuition?

src/ar_http_iface_server.erl
```diff
@@ -65,7 +69,9 @@ split_path(Path) ->
 do_start(Port, GatewayOpts) ->
 	Dispatch = cowboy_router:compile(?HTTP_IFACE_ROUTES),
 	HttpIfaceEnv = #{ dispatch => Dispatch },
-	ok = ar_semaphore:start_link(hash_list_semaphore, 1),
+	ok = ar_semaphore:start_link(hash_list_semaphore, ?MAX_PARALLEL_HASH_LIST_REQUESTS),
+	ok = ar_semaphore:start_link(arql_semaphore, ?MAX_PARALLEL_ARQL_REQUESTS),
```
ldmberman: A small wish, could we assign the semaphore atoms to their corresponding HTTP environments here? Then we do not have to match the `gateway` key in `with_arql_semaphore_req_field`, what is basically resolving the type of server again inside the request handler. Also, there would be no concern that atoms have to match here and there.

maxmellen: Done in 9026590

PR #156, opened 2019-07-15
Add scripting interface for app_block_tx_export

martin-torhage:
- [ ] Check with Sam it's ok to enable distribution like this.


martin-torhage:
Not ready yet

PR #157, opened 2019-07-17
Script block and transaction CSV file export

martin-torhage:
This goes together with https://github.com/ArweaveTeam/wallet-hunt-scripts/pull/1

### DIFF
--- a/apps/arweave/src/apps/app_block_tx_export.erl
+++ b/apps/arweave/src/apps/app_block_tx_export.erl
@@ -1,6 +1,6 @@
 -module(app_block_tx_export).
--export([export_blocks/1, export_blocks/3]).
--export([export_transactions/1, export_transactions/3]).
+-export([export_blocks/1, export_blocks/2]).
+-export([export_transactions/1, export_transactions/2]).
 -include("../ar.hrl").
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -29,66 +29,85 @@ DEFAULT_PEER_OPTS := \
...
+	-setcookie arweave-testing
+
 test_all: test test_apps test_ipfs
 
-test: build_test
-	@erl $(ERL_TEST_OPTS) -noshell -sname slave -setcookie test -run ar main port 1983 data_dir data_test_slave &
-	@erl $(ERL_TEST_OPTS) -noshell -sname master -setcookie test -run ar test_with_coverage -s init stop
+.PHONY: test
+test:
+	@${ERL_DIST} $(ERL_TEST_OPTS) -noshell -name test_slave@127.0.0.1 -setcookie test -run ar main port 1983 data_dir data_test_slave &
```
ldmberman: We need to remove the old cookie now.

arweave-kyle:
@ldmberman let's make it so that the user can configure the directory to output files to, or make it so the default behaviour is to export to an `exports` directory so we don't clog up the main directory?

PR #160, opened 2019-07-24
Fix Mnesia startup after bad interrupt and hostname change

ldmberman:
If a hostname changes and a transaction gets interrupted by a node shutdown, Mnesia refuses to pick up the state assuming a distributed mode with another hostname responsivle for resolving the current state. Since we are only using one Mnesia node all the time, we can force it to load the table.

### DIFF
--- a/src/ar_tx_search.erl
+++ b/src/ar_tx_search.erl
@@ -146,8 +146,17 @@ ensure_table_exists() ->
 		{atomic, ok} -> ok;
 		{aborted, {already_exists, arql_tag}} -> ok
 	end,
-	ok = mnesia:wait_for_tables([arql_tag], 5000),
-	ok.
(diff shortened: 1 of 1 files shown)


src/ar_tx_search.erl
```diff
@@ -146,8 +146,17 @@ ensure_table_exists() ->
 		{atomic, ok} -> ok;
 		{aborted, {already_exists, arql_tag}} -> ok
 	end,
-	ok = mnesia:wait_for_tables([arql_tag], 5000),
-	ok.
+	case mnesia:wait_for_tables([arql_tag], 5000) of
```
maxmellen: Why should we call this function. I think we should just replace it with `force_table_load`. Otherwise we just introduce useless overhead

ldmberman:
Closing at it needs some further investigation.

PR #162, opened 2019-08-14
Add support for optional intermediate CA certificates

ldmberman:
Some tools do not recognize some root CAs so the need for specifying a certificate of an older CA arises.


src/ar_http_iface_server.erl
```diff
@@ -140,5 +140,22 @@ derive_sni_hosts(Domains) ->
...
 			{keyfile, "priv/tls/" ++ SDomain ++ "/key.pem"}
-		]}
+		] ++ cacertfile_when_present(SDomain)}
 	end, Domains).
+
+cacertfile_when_present() ->
+	cacertfile_when_present(apex_domain).
+
+cacertfile_when_present(Domain) ->
+	File = case Domain of
```
maxmellen: Can we maybe rename this to `Name` or `Filename`?

PR #165, opened 2019-08-27
GraphQL Interface to ArQL

maxmellen:
This PR introduces a new interface to query data in the weave using GraphQL.

This PR continues to use the same backing database as the current implementation of ArQL, meaning that this PR only presents negligible performance implications compared to the current implementation.

The main issues addressed by this new interface are:
- Use an existing and well defined query language to let apps fetch data from the weave 
- Reduce the number of roundtrips required to get an exploitable dataset to use in an app on the permaweb
- Allow querying through user-defined relationships between transactions


ldmberman:
The IPFS server (`apps/app_ipfs_daemon_server.erl`) has a handler that reads the request body so needs updating.

PR #167, opened 2019-08-28
Configurable Webhooks

maxmellen:
This PR adds the ability to call configurable web hooks when the node receives new blocks or new transactions.

This is done by adding a `webhooks` field in the config file with the a list of webhook configuration objects with the following fields:
- `events`: an array or either `"transaction"`, `"block"` or both, specifying which events should be communicated to the webhook
- `url`: the URL where the webhook should be called
- `headers`: an object with HTTP header keys and values

For example:
```jsonc
{
  //...
  "webhooks": [
    {
      "events": ["transaction", "block"],
      "url": "https://example.com/hook",
      "headers": {
        "Authorization": "Bearer 123456"
      }
    }
  ]
}
```

### DIFF
--- a/src/adt_simple.erl
+++ b/src/adt_simple.erl
@@ -87,30 +87,25 @@ server(S = #state { gossip = GS }) ->
 					);
 				% New block and confirmed txs callback.
 				{NewGS, {new_block, _, _, B, _, _}} ->
-					ar:d({adt, recvd_new_block}),
-					FullTXs =
-						lists:map(
-							fun(T) -> ar_storage:read_tx(T) end,
-							B#block.txs
+					S1 =
+						apply_callback(
+							S#state { gossip = NewGS },
+							new_block,
+							B
 						),
-					NewS =
+					S2 =
 						lists:foldl(
-							fun(TX, NextS) ->
+							fun(TXID, NextS) ->
 								apply_callback(
 									NextS,
 									confirmed_transaction,
-									TX
+									ar_storage:read_tx(TXID)
 								)
 							end,
-							S#state { gossip = NewGS },
-							FullTXs
+							S1,
+							B#block.txs
 						),
-					server(
-						apply_callback(
-							NewS,
-							new_block,
-							B#block { txs = FullTXs })
-						)
+					server(S2)
 			end;
 		stop -> ok;
 		OtherMsg ->
--- a/src/ar.erl
+++ b/src/ar.erl
@@ -254,7 +254,8 @@ start(
 		gateway_custom_domains = GatewayCustomDomains,
 		requests_per_minute_limit = RequestsPerMinuteLimit,
 		ipfs_pin = IPFSPin,
-		ipfs_import = IPFSImport
+		ipfs_import = IPFSImport,
+		webhooks = WebhookConfigs
 	}) ->
 	%% Start the logging system.
 	error_logger:logfile({open, Filename = generate_logfile_name()}),
@@ -439,6 +440,7 @@ start(
 		false -> ok;
 		true  -> app_ipfs_daemon_server:start()
 	end,
+	ar_node:add_peers(Node, ar_webhook:start(WebhookConfigs)),
 	case Pause of
 		false ->
 			ok;
--- a/src/ar.hrl
+++ b/src/ar.hrl
@@ -248,7 +248,7 @@
 %% @doc Adjust the difficulty up to the limit.
 -define(DIFF_ADJUSTMENT_UP_LIMIT(Diff), (Diff * 4)).
 
-%% @doc A block on the weave.
+%% @doc A full block or block shadow (see more on txs field).
 -record(block, {
 	nonce = <<>>, % The nonce used to satisfy the mining problem when mined
 	previous_block = <<>>, % indep_hash of the previous block in the weave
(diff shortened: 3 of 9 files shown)


src/ar.erl
```diff
@@ -439,6 +440,7 @@ start(
 		false -> ok;
 		true  -> app_ipfs_daemon_server:start()
 	end,
+	ar_node:add_peers(Node, app_webhook:start(WebhookConfigs)),
```
ldmberman: I'd say we should make it the core `ar_webhook` module for now as we can't move it out because it depends on the node config and needs to be hooked up here.

src/apps/app_webhook.erl
```diff
@@ -0,0 +1,126 @@
...
+
+new_block(Configs, Block) ->
+	call_webhooks(Block, Configs),
+	Configs.
+
+new_transaction(Configs, Tx) ->
+	call_webhooks(Tx, Configs),
+	Configs.
+
+call_webhooks(Entity, Configs) ->
```
ldmberman: What do you think about

```
call_webhooks({transaction, TX}, Configs),
```

and later

```
call_webhooks({Event, Entity}, Configs) ->
...
case lists:member(Event, Events) of
...
```
?

src/apps/app_webhook.erl
```diff
@@ -0,0 +1,126 @@
...
+				ok = fusco:disconnect(Client),
+				ar:info([successful_webhook_call, {
+					{type, entity_type(Entity)},
+					{id, entity_id(Entity)},
+					{url, Url},
+					{headers, Headers},
+					{response, Result}
+				}]),
+				Url;
+			UnsuccessfulResult ->
```
ldmberman: I think, it only makes sense to retry `{error, connection_closed | connect_timeout | timeout}` because in other cases it's very unlikely a retry would help. May be also `5**` status codes. At the very least, I would not retry on exception.

src/apps/app_webhook.erl
```diff
@@ -0,0 +1,126 @@
...
+		{headers, Headers},
+		{number_of_tries, ?NUMBER_OF_TRIES},
+		{wait_between_tries, ?WAIT_BETWEEN_TRIES}
+	]),
+	ok.
+
+entity_type(#block {}) -> block;
+entity_type(#tx {}) -> transaction.
+
+entity_id(#block { indep_hash = Id }) -> ar_util:encode(Id);
```
ldmberman: A style note: `ID`, `TX`, `URL`, `JSON`.

src/apps/app_webhook.erl
```diff
@@ -0,0 +1,126 @@
...
+entity_id(#tx { id = Id }) -> ar_util:encode(Id).
+
+to_json(#block {} = Block, _) ->
+	{JsonKVPairs} = ar_serialize:block_to_json_struct(Block),
+	JsonStruct = {lists:keydelete(wallet_list, 1, JsonKVPairs)},
```
ldmberman: `wallet_list`s in the serialized blocks are hashes now, not sure if it's worth deleting them.

src/adt_simple.erl
```diff
@@ -86,31 +86,26 @@ server(S = #state { gossip = GS }) ->
 						)
 					);
 				% New block and confirmed txs callback.
-				{NewGS, {new_block, _, _, B, _, _}} ->
-					ar:d({adt, recvd_new_block}),
-					FullTXs =
-						lists:map(
-							fun(T) -> ar_storage:read_tx(T) end,
-							B#block.txs
+				{NewGS, {new_block, _, _, BSD, _, _}} ->
```
ldmberman: In some other places we are following the convention `B` - block shadow, `FullB` - full block.

src/adt_simple.erl
```diff
@@ -86,31 +86,26 @@ server(S = #state { gossip = GS }) ->
 						)
 					);
 				% New block and confirmed txs callback.
-				{NewGS, {new_block, _, _, B, _, _}} ->
```
ldmberman: An experiment would tell for sure, but from reading the code I conclude blocks accepted from peers are not sent to the internal gossip network - only the blocks mined by the node and the blocks received in the polling mode are.

If it's true, it can be fixed by calling `ar_gossip:send(GS, {new_block, ...`  somewhere [here](https://github.com/ArweaveTeam/arweave/blob/master/src/ar_node_utils.erl#L439), like it is done [here](https://github.com/ArweaveTeam/arweave/blob/master/src/ar_node_worker.erl#L599).

src/apps/app_webhook.erl
maxmellen: Addressed in 02d5ac19 👍

src/apps/app_webhook.erl
maxmellen: Addressed in 115b5874

src/adt_simple.erl
maxmellen: I'm pretty sure it's the opposite that happens: the callback does get called when the node receives a block, but not if it asks for one by polling

src/ar.erl
maxmellen: Addressed in d950db75

src/adt_simple.erl
maxmellen: Changed it back in 87cd4a02
I find `BSD` more explicit though, let's chat about it

src/apps/app_webhook.erl
ldmberman: Ah, no, a hash is computed when it's stored, here we do get the full list so nevermind.

src/apps/app_webhook.erl
maxmellen: Let's chat about it. Maybe this is good enough for now, and we can improve this later. I also think it would be nice to rework the whole queuing logic at some point so it cycles through the outgoing messages on failure rather than getting stuck on one

src/adt_simple.erl
ldmberman: Ah, yes, it should be coming from [the bridge](https://github.com/ArweaveTeam/arweave/blob/master/src/ar_bridge.erl#L222). So as soon as it passes edge verification (timestamp, PoW) it is sent here. I guess we could live with it, but the module documentation could make it clearer, what do you think?

src/apps/app_webhook.erl
ldmberman: Yes, the queueing logic could be improved.

What are the downsides of changing it so that it only retries connection and server errors now?

PR #168, opened 2019-09-04
Keep detailed track of transaction processing

ldmberman:


src/ar_tx_db.erl
```diff
@@ -15,36 +20,58 @@ start() ->
...
-%% @doc Retreive a term from the meta db.
 get_error_codes(TXID) ->
-	case ets:lookup(?MODULE, TXID) of
-		[{_, ErrorCodes}] -> {ok, ErrorCodes};
-		[] -> not_found
-	end.
+	Log = get_tx_log(TXID),
+	search_for_validation_error(Log).
+
+search_for_validation_error([{{validation_error, ErrorCodes}, _} | _] ) ->
```
maxmellen: If this is an ETS table, couldn't you use a match or a select to go through the records? Or is it not worth the optimisation?

ldmberman: `match` makes sense 👍

ldmberman: On second thought, matching is helpful if the key is specified in the pattern. In this case the event list is fetched by TXID first, and further searching inside the list is not something `ets:match` can help with.

Issue #68, opened 2018-10-16
Mining config generator should elegantly handle missing data


samcamwilliams:
The mining config generator web page ( https://github.com/ArweaveTeam/arweave/blob/feature/config-generator/src/apps/config_generator.html ) currently does not handle missing input very well.

For example, if you do not provide a trusted peer, the output command will include `./arweave-server peer mining_addr` which is not a valid argument set.

100 AR for anyone that can add the appropriate error state handling!

ldmberman:
Addressed by https://github.com/ArweaveTeam/arweave/pull/171.

cedriking:
@mkg20001 100 AR sent!
Transaction ID: `oXSHOjIIcos-8wc9Sv5d1Y2Jhfaiqi0Cp3tjkWRXjwo`

Thank you!

PR #171, opened 2019-09-10
Config Generator changes to fix #67 and #68

### LINKED ISSUE #67: Mining config generator needs styling
The Arweave mining configuration generator (found here: https://github.com/ArweaveTeam/arweave/blob/feature/config-generator/src/apps/config_generator.html ) is helpful, but is currently poorly styled.

There is up to a 200 AR bounty available for anyone that would like to give this a nicer UI!

mkg20001:
screenshot

### DIFF
--- a/src/apps/config_generator.html
+++ b/src/apps/config_generator.html
@@ -15,7 +15,7 @@
         err: 'Must be a valid IP address'
       },
       ipList: {
-        val: (v) => !v.split(' ').map(v => v.match(/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/)).filter(v => !v).length,
+        val: (v) => !v.split(' ').map(v => v.match(/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}:?[0-9]{0,5}$/)).filter(v => !v).length,
         err: 'Must be a valid IP address or list of IP addresses'
       },
       address: {
--- a/src/apps/config_generator.html
+++ b/src/apps/config_generator.html
@@ -5,6 +5,7 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <title>Arweave Miner Configuration Generator</title>
+
   <script>
     const id = document.getElementById.bind(document)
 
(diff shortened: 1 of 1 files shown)


mkg20001:
With error handling

screenshot

ldmberman:
Hey @mkg20001!

Thank you for submitting the improvement, it looks much nicer now! Also, sorry for the late reply.

I would ask to change the version in `Last updated for:` to `Release 28` - we do not use the `Arweave.N.X.X` scheme at the moment. I can change it myself - let me know if you prefer it.

I have a feature request - the support for multiple peers, like `peer 188.166.200.45 peer 188.166.192.169`. It is a very common use case; for example, we recommend using multiple peers in the mining guide. Please, tell me if you are willing to do it in this PR - if not, I will merge it right away as it already provides value.

mkg20001:
screenshot

I've added the changes

Btw, it seems like there's a bounty for both of those issues. Does this pr qualify?
If so could you send me the tokens to `qY6XZdOFKAUArsjIl9B_YcnExaJMT0mBFOPVCLSlXhM`

ldmberman:
Hi @mkg20001,

Thank you for the prompt response!

I asked the team about the bounty. I will get back to you regarding it soon.

Specifying multiple peers works nicely! Would it be difficult to allow an IP to contain the port, like `127.0.0.1:1983` - nodes often run on custom ports.

I've also just noticed there used to be some options in the "Feature flags" section, but now it's empty - could those have been removed accidentally?

mkg20001:
> I've also just noticed there used to be some options in the "Feature flags" section, but now it's empty - could those have been removed accidentally?

These options seem to no longer exist in the latest arweave, that's why I removed them.

> Specifying multiple peers works nicely! Would it be difficult to allow an IP to contain the port, like 127.0.0.1:1983 - nodes often run on custom ports.

It could be added. A quick'n'dirty validation would be `:?[0-9]{4,5}?`

ldmberman:
> These options seem to no longer exist in the latest arweave, that's why I removed them.

They exist as part of the flexible enable/disable mechanism. You can specify an arbitrary string after `enable` or `disable`, but it will only take effect if the miner actually expects it. A better approach might be to convert a space separated input to a sequence of `enable X enable Y enable Z` (and another one for `disable`), like you did for peers. What do you think about it?

> It could be added. A quick'n'dirty validation would be :?[0-9]{4,5}?

👍 looks good enough

mkg20001:
> They exist as part of the flexible enable/disable mechanism. You can specify an arbitrary string after enable or disable, but it will only take effect if the miner actually expects it. A better approach might be to convert a space separated input to a sequence of enable X enable Y enable Z (and another one for disable), like you did for peers. What do you think about it?

Unsure what you mean

ldmberman:
I mean, UI may look like:
```
Enable features          feature1 feature2
Disable featuers         feature3 feature4
```
what is converted into
```
./arweave-server enable feature1 enable feature2 disable feature3 disable feature4
```

mkg20001:
Ahh so you mean a generic field which can take multiple strings and add them to the command? Doable

ldmberman:
@mkg20001 awesome, thank you for the update!

ldmberman:
@mkg20001 I edited the text a tiny bit after passing it through Grammarly, here is the [merged commit](https://github.com/ArweaveTeam/arweave/commit/a05b6577a3187260d2ed929ca58860fc045c8b59).

A reward should be sent to you soon!

PR #173, opened 2019-09-18
Use a bigger safety buffer when estimating TX prices

ldmberman:
The buffer is the same to that used before the switch to the linear difficulty. A more optimistic estimation does not perform well under the present difficulty fluctuations.

### DIFF
--- a/src/ar_difficulty.erl
+++ b/src/ar_difficulty.erl
@@ -1,6 +1,6 @@
 -module(ar_difficulty).
 
--export([next_cumulative_diff/3]).
+-export([next_cumulative_diff/3, twice_smaller_diff/1]).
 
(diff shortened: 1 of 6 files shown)


llaisdy:
Looks good by eye.

Is `MaxDiff - (MaxDiff - CurrentDiff) * 2` equivalent to old-style `Diff - 2`?

Minor quibble: I might export `twice_smaller_diff` (maybe even from `ar_difficulty.erl`) and use it in `ar_http_iface_middleware.erl` as well as in the tests module.

ldmberman:
@llaisdy equivalent to old-style `Diff - 1`, yes. I added a fixup, also removed the confusing `Height` argument from `ar_mine:max_diifficulty`, which does not depend on it.

PR #177, opened 2019-09-24
Implement a TX propagation priority queue

ldmberman:
Network bandwidth is a limited resource therefore miners
need a way to prioritise certain transactions over the other.
The commit introduces a priority queue where the transactions
with the higher utility are propagated first. The utility is
computed as reward / size - a proxy for the cost of a unit of
traffic. Scaling the utility this way creates a market for the
limited bandwidth resource.

The commit also changes the order in which transactions
are included into blocks - higher utility transactions are
included first. In this case, the utility function proxies
the cost per unit of storage.


apps/arweave/src/ar_tx_queue.erl
```diff
@@ -0,0 +1,284 @@
...
+		true ->
+			{NewQ, TX, DroppedSize} = drop_lowest(Q),
+			maybe_drop(NewQ, Size - DroppedSize, MaxSize, [TX | DroppedTXs]);
+		false ->
+			{Q, Size, lists:filter(fun(TX) -> TX /= none end, DroppedTXs)}
+	end.
+
+%% @doc Drop lowest scoring item(s) from priority queue.
+%% NOTE relies on the internal implementation of priority_queue.
+drop_lowest(Q = {pqueue, []}) ->
```
maxmellen: Since we need to look at this library's internals to be able to drop the lowest element, I wonder if we can use erlang's `gb_sets` with `{Priority, Item}` tuples to emulate a priority queue and not use this library at all.

apps/arweave/src/ar_tx_queue.erl
```diff
@@ -0,0 +1,284 @@
...
+					[] ->
+						erlang:send(?MODULE, {emitter_finished, TX}),
+						State#state { tx_queue = NewQ, size = Size - TXSize };
+					_ ->
+						%% Send transactions to the "max_propagation_peers" best peers,
+						%% ?TX_PROPAGATION_PARALLELIZATION peers at a time. The resulting
+						%% maximum for outbound connections is ?TX_PROPAGATION_PARALLELIZATION * num_emitters.
+						lists:foreach(
+							fun(_) ->
+								erlang:send(?MODULE, {emit_tx_to_peer, TX})
```
maxmellen: What about `self() ! {emit_tx_to_peer, TX}` here?
I find the `!` operator more concise than `erland:send/2` and I prefer using `self()` inside a gen_server, since it makes the internals independent of whether or not the gen_server has a registered name.

apps/arweave/src/ar_tx_queue.erl
```diff
@@ -0,0 +1,284 @@
...
+	case DroppedTXs of
+		[] -> noop;
+		_ -> ar_bridge:drop_waiting_txs(whereis(http_bridge_node), DroppedTXs)
+	end,
+	{noreply, State#state { tx_queue = NewQ, size = NewSize }};
+
+handle_cast(_Request, State) ->
+	{noreply, State}.
+
+handle_info(emitter_go, State = #state { paused = true }) ->
```
maxmellen: Why don't we use casts instead of arbitrary messages for these operations?

apps/arweave/src/ar_tx_queue.erl
```diff
@@ -0,0 +1,284 @@
...
+handle_info({emit_tx_to_peer, TX}, State = #state { emit_map = EmitMap }) ->
+	TXID = TX#tx.id,
+	case EmitMap of
+		#{ TXID := #{ peers := [] } } ->
+			{noreply, State};
+		#{ TXID := TXIDMap = #{ peers := [Peer | Peers] } } ->
+			spawn(
+				fun() ->
+					ar_http_iface_client:send_new_tx(Peer, TX),
+					erlang:send(?MODULE, {emitted_tx_to_peer, TX})
```
maxmellen: Same comment, let's either use `!` or casts.

apps/arweave/src/ar_tx_queue.erl
```diff
@@ -0,0 +1,284 @@
...
+	end,
+	{noreply, State#state { tx_queue = NewQ, size = NewSize }};
+
+handle_cast(_Request, State) ->
+	{noreply, State}.
+
+handle_info(emitter_go, State = #state { paused = true }) ->
+	erlang:send_after(?EMITTER_START_WAIT, ?MODULE, emitter_go),
+	{noreply, State};
+handle_info(emitter_go, State = #state { emitters_running = EmittersRunning, max_emitters = MaxEmitters }) when EmittersRunning > MaxEmitters ->
```
maxmellen: I don't understand this clause. Let's chat about it :)

apps/arweave/src/ar_tx_queue.erl
```diff
@@ -238,35 +236,17 @@ maybe_drop(Q, Size, MaxSize) ->
...
-			]),
-			{{pqueue, lists:reverse([{LP, QO} | Rest])}, TX, Size}
-	end;
-drop_lowest(X) ->
-	{X, none, 0}.
-
 show_queue(Q) ->
-	[{ar_util:encode(TX#tx.id), TX#tx.reward, Size}
-	 || {_, {TX, Size}} <- priority_queue:to_list(Q)].
+	lists:reverse([
```
maxmellen: I don't know if we call `show_queue` often but if we wanted to avoid `reverse`ing the list, we could use `foldl`:
```erlang
show_queue(Q) ->
    lists:foldl(fun({_, {TX, Size}}, Acc) ->
        [{ar_util:encode(TX#tx.id), TX#tx.reward, Size} | Acc]
    end, [], gb_sets:to_list(Q)).
```
But if show_queue isn't called so often, I actually find the reverse more readable.

PR #178, opened 2019-09-24
Path manifest support for gateways

maxmellen:


src/ar_gateway_middleware.erl
```diff
@@ -132,21 +139,29 @@ handle_unknown_request(Req, Env) ->
...
+resolve_path_1(Hash, SubPathSegments) ->
 	case get_tx_from_hash(Hash) of
-		{ok, TXID, Filename} -> {tx, TXID, Filename};
-		invalid -> other;
-		not_found -> not_found
+		{ok, TXID, Filename} ->
+			SubPath = list_to_binary(lists:join(<<"/">>, SubPathSegments)),
+			{tx, TXID, Filename, SubPath};
+		invalid ->
+			other;
```
ldmberman: The callers seem to always expect a `{other, _}` tuple now.

src/ar_gateway_middleware.erl
```diff
@@ -237,11 +253,122 @@ get_tx_from_hash_1(TXID) ->
...
+		end,
+	case {ContentType, Path} of
+		{?MANIFEST_CONTENT_TYPE, <<>>} ->
+			serve_manifest(TX, Req);
+		{?MANIFEST_CONTENT_TYPE, _} ->
+			serve_manifest_path(TX, Path, Fallback, Req, Env);
+		{_, <<>>} ->
+			serve_plain_tx(TX, ContentType, Req);
+		_ ->
+			bad_request(Req)
```
ldmberman: Shouldn't we also consider the fallback in this case? Specifically, a custom domain request to a custom path, which we pass on to other middleware, may end up here, right?

src/ar_gateway_middleware.erl
```diff
@@ -237,11 +253,122 @@ get_tx_from_hash_1(TXID) ->
...
+			serve_manifest_listing(ar_util:encode(TX#tx.id), maps:keys(Paths), Req);
+		{error, _} ->
+			misdirected_request(Req)
+	end.
+
+serve_manifest_index(Index, Req) ->
+	Hostname = cowboy_req:host(Req),
+	Path = cowboy_req:path(Req),
+	NewPath = filename:join(Path, Index),
+	Location = <<
```
ldmberman: Is there a way to serve index without changing the path?

src/ar_gateway_middleware.erl
```diff
@@ -237,11 +253,122 @@ get_tx_from_hash_1(TXID) ->
...
+
+html_escape(Html) ->
+	lists:foldl(fun({Char, Escape}, NextHtml) ->
+		binary:replace(NextHtml, Char, Escape, [global])
+	end, Html, [
+		{<<"&">>, <<"&amp;">>},
+		{<<"<">>, <<"&lt;">>},
+		{<<">">>, <<"&gt;">>},
+		{<<"\"">>, <<"&quot;">>},
+		{<<"'">>, <<"&apos;">>}
```
ldmberman: I suppose, we need to escape `"` in addition to it `'` (also usually recommended in the XSS prevention guides).

test/ar_gateway_middleware_tests.erl
```diff
@@ -0,0 +1,577 @@
...
+		{"return 400 on subpath of non-manifest", fun return_400_on_subpath_of_non_manifest/0},
+		{"serve manifest subpath on custom domain", fun serve_manifest_subpath_on_custom_domain/0},
+		{"forward non-existent subpaths to node api", fun forward_non_existent_subpaths_to_node_api/0}
+	]}.
+
+setup() ->
+	meck:new(ar_storage),
+	meck:new(ar_tx_search),
+	meck:new(ar_domain),
+	meck:expect(
```
ldmberman: You can do `meck:new(ar_domain, [passthrough])`, then you do not have to redefine methods you do not want to mock.

test/ar_gateway_middleware_tests.erl
```diff
@@ -0,0 +1,577 @@
...
+	54, 5, 77, 26, 132, 131, 130, 249, 240, 18, 210, 37, 237, 159, 110, 108
+>>).
+
+-define(MANIFEST_DOG_TXID, <<
+	158, 115, 11, 58, 243, 223, 99, 147, 125, 173, 58, 194, 61, 208, 0, 46,
+	32, 73, 227, 78, 1, 176, 210, 161, 173, 22, 237, 195, 181, 19, 68, 151
+>>).
+
+execute_test_() ->
+	{foreach, fun setup/0, fun teardown/1, [
```
ldmberman: These tests are awesome, great job!

test/ar_gateway_middleware_tests.erl
```diff
@@ -0,0 +1,577 @@
+-module(ar_gateway_middleware_tests).
```
ldmberman: I added a fixup up adding this module to the list in `ar.erl` to make `make test` pick up these tests.

src/ar_gateway_middleware.erl
maxmellen: Good catch 👍  Fixed in eb5b42ee

src/ar_gateway_middleware.erl
maxmellen: Already done 😅
https://github.com/ArweaveTeam/arweave/blob/557cd2e9f64fe6af46ddd25598691b6524192356/src/ar_gateway_middleware.erl#L325

src/ar_gateway_middleware.erl
maxmellen: Good catch! Addressed in d118d06e

src/ar_gateway_middleware.erl
maxmellen: We should be able to add this in the future without breaking backwards compatibility, so it should be fine to move forward with this current implementation 🙂

ldmberman: 👍

nionis:
hello guys, do we have an ETA when this will be available in production?

ldmberman:
@nionis hi, I think, next week.

PR #188, opened 2019-10-25
Feature/efficient used space estimation

ldmberman:


src/ar_storage.erl
```diff
@@ -385,45 +391,22 @@ lookup_tx_filename(ID) ->
...
-		Directory,
-		"",
-		false,
-		fun(File, Acc) -> Acc + filelib:file_size(File) end,
-		0
-	).
+%% @doc Calculate the used space in bytes on the data directory disk.
+calculate_used_space() ->
+	{_, Size, UsedPercentage} = get_data_dir_disk_data(),
+	math:ceil(Size * UsedPercentage / 100 * 1024).
```
maxmellen: Why are we multiplying by 1024 here?

src/ar_storage.erl
```diff
@@ -385,45 +391,22 @@ lookup_tx_filename(ID) ->
...
 
-%% @doc Calculate the total amount of disk space available
-calculate_disk_space() ->
+get_data_dir_disk_data() ->
 	application:start(sasl),
 	application:start(os_mon),
 	DataDir = filename:absname(ar_meta_db:get(data_dir)),
-	[{_,Size,_}|_] = select_drive(disksup:get_disk_data(), DataDir),
-	Size*1024.
+	[DiskData | _] = select_drive(disksup:get_disk_data(), DataDir),
```
maxmellen: It's hard to understand why we only take the head of that list here.

src/ar_storage.erl
```diff
@@ -385,45 +391,22 @@ lookup_tx_filename(ID) ->
...
-	lists:foldl(
-		fun(Dir, Acc) ->
-			Acc + calculate_used_space(Dir)
-		end,
-		0,
-		Dirs
-	).
+%% @doc Calculate the available space in bytes on the data directory disk.
+calculate_disk_space() ->
+	{_, Size, _} = get_data_dir_disk_data(),
```
maxmellen: Why did we move the pattern match outside of `get_data_dir_disk_data`? I guess this tuple is related to `disksup:get_disk_data` so it would be nice to keep that pattern match as close as possible to that disksup call.

ldmberman: I use `get_data_dir_disk_data` in both `calculate_used_space` and `calculate_disk_space`, which depend on different items of this tuple, but neither of them needs the first element, what would you say to returning a tuple of two?

src/ar_storage.erl
ldmberman: `disksup` returns KBytes. Would you prefer to have this conversion inside `get_data_dir_disk_data`?

src/ar_storage.erl
ldmberman: I agree, `select_drive` is scary, but I would rather not change it here.

src/ar_storage.erl
maxmellen: In that case, maybe it's fine to keep it as it is. I get you do use the same `disk_data` naming.

src/ar_storage.erl
maxmellen: How about just renaming the variable to KByteSize then?

PR #189, opened 2019-10-29
Improvement/remove block index

ldmberman:


src/ar_storage.erl
```diff
@@ -49,52 +50,94 @@ ensure_directories() ->
...
+		end
+	).
 
-%% @doc Removes a saved block.
-delete_block(Hash) ->
-	ar_block_index:remove(Hash),
-	file:delete(block_filepath(Hash)).
+rename_block_files(HL) ->
+	BlocksDir = filename:join(ar_meta_db:get(data_dir), ?BLOCK_DIR),
+	case filelib:is_file(filename:join(BlocksDir, "blocks_renamed")) of
```
maxmellen: So do we track this migration with the existence of this `block_renamed` file? If so, we might wanna name the file something that indicates the new format, otherwise, we might have something like `blocks_renamed_again` in the future... Maybe something like `indep_hash_only_names`. I'm not sure if we'd wanna postfix some `_v1` afterthat, that might be overengineering it, but then again, [you can see how far I went with SQLite](https://github.com/ArweaveTeam/arweave-private/pull/148/files#diff-f1df8ff18cce849ef7902ab918e314a2R277) 🙃

src/ar_storage.erl
```diff
@@ -49,52 +50,94 @@ ensure_directories() ->
...
+								BlocksDir,
+								EncodedH ++ ".json"
+							),
+							file:rename(LegacyName, NewName),
+							Height - 1
+						end,
+						length(HL) - 1,
+						HL
+					),
+					write_file_atomic(filename:join(BlocksDir, "blocks_renamed"), <<>>),
```
maxmellen: Why do we need this `write_file_atomic` function?

src/ar_storage.erl
```diff
@@ -49,52 +50,94 @@ ensure_directories() ->
...
+	Filepath = block_filepath(Hash),
+	case filelib:is_file(Filepath) of
+		true ->
+			Filepath;
+		false ->
+			unavailable
 	end.
 
+%% @doc Remove blocks from disk. Used in tests.
+clear() ->
```
maxmellen: I'm not sure if wrapping this function in an `-ifdef(DEBUG)` would be paranoid

src/ar_storage.erl
```diff
@@ -178,25 +203,39 @@ write_encrypted_block(Hash, B) ->
...
+
+%% @doc Read the block from disk, given a hash or a height.
+read_block(unavailable, _HL) ->
+	unavailable;
+read_block(B, _HL) when is_record(B, block) ->
+	B;
+read_block(Blocks, HL) when is_list(Blocks) ->
+	lists:map(fun(B) -> read_block(B, HL) end, Blocks);
+read_block(Height, HL) when is_integer(Height) ->
+	case Height of
```
maxmellen: I know we're not big fans of the erlang if statement, but this is really asking to be one

src/ar_storage.erl
```diff
@@ -457,13 +440,9 @@ to_string(String) ->
 	String.
 
 block_filename(B) when is_record(B, block) ->
-	iolist_to_binary([
-		integer_to_list(B#block.height), "_", ar_util:encode(B#block.indep_hash), ".json"
-	]);
+	block_filename(B#block.indep_hash);
 block_filename(Hash) when is_binary(Hash) ->
-	iolist_to_binary(["*_", ar_util:encode(Hash), ".json"]);
```
maxmellen: I'm surprised this was even possible 😮

src/ar_storage.erl
ldmberman: With atomic writes, the files do not get corrupted if the machine is halted in the process. Also, it allowes two nodes to safely share the storage.

apps/arweave/src/ar_join.erl
```diff
@@ -25,7 +25,7 @@ start(Node, Peers, B) when is_atom(B) ->
 start(Node, RawPeers, NewB) ->
 	case whereis(join_server) of
 		undefined ->
-			ar_storage:rename_block_files(NewB#block.hash_list),
+			ar_storage_migration:run_migrations(NewB#block.hash_list),
```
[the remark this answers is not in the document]
ldmberman: Here we join the network, there we start from the given hash list.

apps/arweave/src/ar_storage_migration.erl
```diff
@@ -0,0 +1,57 @@
...
+			EncodedH = binary_to_list(ar_util:encode(H)),
+			LegacyName = filename:join(
+				BlocksDir,
+				integer_to_list(Height) ++ "_" ++ EncodedH ++ ".json"
+			),
+			NewName = filename:join(
+				BlocksDir,
+				EncodedH ++ ".json"
+			),
+			case file:rename(LegacyName, NewName) of
```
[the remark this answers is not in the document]
ldmberman: Yes, every migration function in this list needs to be "idempotent" - should eventually reach and stay at the completed state when invoked multiple times no matter what were the results of the previous runs.

ldmberman:
Closing since it was moved to `feature/fast-write` in a slightly changed form.

PR #191, opened 2019-11-07
Implement limits on ArQL queries

maxmellen:


priv/schema.graphql
```diff
@@ -1,16 +1,17 @@
 type Query {
     transaction(id: ID!): Transaction
-    transactions(from: [String!], to: [String!], tags: [TagInput!]): [Transaction!]!
-    countTransactions(from: [String!], to: [String!], tags: [TagInput!]): Int!
+    transactions(from: [String!], to: [String!], tags: [TagInput!], limit: Int, offset: Int): [Transaction!]!
+    countTransactions(from: [String!], to: [String!], tags: [TagInput!], limit: Int, offset: Int): Int!
```
ldmberman: What do you think about implementing count via SQL COUNT (not necessarily in this PR)? I think, the limit makes it counterintuitive and it does not seem to add much value to the `transaction` query in this form.

PR #192, opened 2019-11-07
Make polling mode more efficient

ldmberman:
Motivated by the possibility of the connectivity issues.

The improvement includes:

- Get rid of extra hash list, blocks, and txs downloads - fetch block shadows by height.
- Synchronize the polling server with the bridge to avoid double work.
- Check for new blocks every minute even if polling is not enabled explicitly, because it's very cheap now.

### DIFF
--- a/apps/arweave/src/ar.erl
+++ b/apps/arweave/src/ar.erl
@@ -105,8 +105,11 @@ show_help() ->
 			{"port", "The local port to use for mining. "
 						"This port must be accessible by remote peers."},
 			{"data_dir", "The directory for storing the weave and the wallets (when generated)."},
-			{"polling", "Poll peers for new blocks. Useful in environments where "
-						"port forwarding is not possible."},
+			{"polling", "Poll peers for new blocks every 10 seconds. "
+						"Useful in environments where "
+						"port forwarding is not possible. "
+						"When the flag is not set, the node still polls "
+						"if it does not receive blocks for a minute."},
 			{"clean", "Clear the block cache before starting."},
 			{"no_auto_join", "Do not automatically join the network of your peers."},
 			{"init", "Start a new blockweave."},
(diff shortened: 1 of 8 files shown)


src/ar.erl
```diff
@@ -436,13 +439,13 @@ start(
...
+	PollingArgs = [{trusted_peers, Peers}] ++ case Polling of
 		true ->
-			ar_meta_db:put(polling_mode, true),
-			ar_poller:start(Node, Peers);
+			[{polling_interval, 10 * 1000}];
 		false ->
-			do_nothing
+			[]
 	end,
+	{ok, _} = ar_poller_sup:start_link(PollingArgs),
```
maxmellen: How come we always start `ar_poller`? Especially nice it seems that it'll start polling anyway

src/ar_poller.erl
```diff
@@ -1,84 +1,99 @@
...
+
+handle_call(_Request, _From, State) ->
+	{noreply, State}.
+
+%%%===================================================================
+%%% Internal functions.
+%%%===================================================================
+
+start_polling() ->
+	gen_server:cast(?MODULE, schedule_polling).
```
maxmellen: So I would do `timer:apply_interval(Interval, gen_server, cast, [?MODULE, schedule_polling])` here and store the returned timer reference in the gen_server's state. It could then be canceled using that `cancel_timer` function using that reference

src/ar_poller.erl
```diff
@@ -1,84 +1,89 @@
...
+		_ ->
+			poll_block(maps:get(trusted_peers, State, []), Height + 1),
+			ok
+	end,
+	{noreply, State};
+handle_cast({set_height, Height}, State) ->
+	#{ polling_timer := PollingTimer, interval := Interval } = State,
+	erlang:cancel_timer(PollingTimer),
+	{ok, {interval, NewPollingTimer}} = start_polling(Interval),
+	{noreply, State#{ height => Height, polling_timer => NewPollingTimer }}.
```
llaisdy: Minor point about styling.  Isn't this more idiomatic?

```erlang
handle_cast(poll_block, #{ height := -1 } = State) ->
        %% Wait until the node joins the network or starts from a hash list.
        {noreply, State};
handle_cast(poll_block, #{ height := Height } = State) ->
        poll_block(maps:get(trusted_peers, State, []), Height + 1),
        {noreply, State};
handle_cast({set_height, Height},
                #{ polling_timer := PollingTimer, interval := Interval } = State) ->
        erlang:cancel_timer(PollingTimer),
        {ok, {interval, NewPollingTimer}} = start_polling(Interval),
        {noreply, State#{ height => Height, polling_timer => NewPollingTimer }}.
```

llaisdy:
Looks good & all tests pass.

PR #194, opened 2019-11-12
Return specific error codes for failed txs

ldmberman:
The network does not depend on this API, but we need to check if `arweave-js` and browser extensions need updating and talk to Upvest.

### DIFF
--- a/src/ar_tx.erl
+++ b/src/ar_tx.erl
@@ -77,7 +77,7 @@ sign(TX, PrivKey, PubKey) ->
 %% NB: The signature verification is the last step due to the potential of an attacker
 %% submitting an arbitrary length signature field and having it processed.
 -ifdef(DEBUG).
-verify(#tx { signature = <<>> }, _, _, _, _) -> true;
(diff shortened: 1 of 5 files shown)


src/ar_tx.erl
```diff
@@ -94,25 +94,23 @@ do_verify(TX, Diff, Height, Wallets, Timestamp) ->
...
+		{tx_id_not_valid,
 		 tx_verify_hash(TX)},
-		{"overspend",
+		{tx_insufficient_funds,
 		 validate_overspend(TX, ar_node_utils:apply_tx(Wallets, TX, Height))},
-		{"tx_signature_not_valid",
+		{tx_signature_not_valid,
 		 ar_wallet:verify(TX#tx.owner, signature_data_segment(TX), TX#tx.signature)}
-	],
+	] ++ tx_field_size_limit(TX, Height),
```
maxmellen: I'd maybe pluralize that function name since it now returns a list of limits

PR #195, opened 2019-11-14
Use Rebar3 to create Erlang releases

maxmellen:


apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -93,13 +93,14 @@ start_http_gateway_listener(off) ->
...
 start_https_gateway_listener({on, Domain, CustomDomains}) ->
+	PrivDir = code:priv_dir(arweave),
 	Dispatch = cowboy_router:compile([{'_', ?HTTP_IFACE_ROUTES}]),
 	HttpIfaceEnv = #{ dispatch => Dispatch },
 	SniHosts = derive_sni_hosts(CustomDomains),
 	TransportOpts = [
 		{port, 443},
-		{certfile, "priv/tls/cert.pem"},
-		{keyfile, "priv/tls/key.pem"},
+		{certfile, filename:join([PrivDir, "tls/cert.pem"])},
```
ldmberman: We need to update the gateway setup guide with new paths.

ldmberman:
Noticed we have `src/av/.av_detect.erl.swp`, let's remove it.

apps/arweave/src/rsa_pss.erl
```diff
@@ -31,10 +31,10 @@
 
 -spec sign(Message, DigestType, PrivateKey) -> Signature
 	when
-		Message    :: binary() | {digest, binary()},
```
ldmberman: Was this change intentional?

apps/arweave/src/arweave.app.src
```diff
@@ -0,0 +1,19 @@
+{application, arweave, [
+	{description, "Arweave"},
+	{vsn, "1.9.0.0"},
+	{mod, {arweave_app, []}},
+	{registered, []},
+	{applications, [
+		kernel,
+		stdlib,
+		os_mon,
```
ldmberman: We need to update `application:start(sasl), application:start(os_mon)` in `ar_storage.erl`.

apps/arweave/src/ar_http_iface_server.erl
ldmberman: Updated.

apps/arweave/src/arweave_app.erl
```diff
@@ -0,0 +1,20 @@
+-module(arweave_app).
+-behaviour(application).
+
+-export([start/2, stop/1]).
+
+start(_StartType, _StartArgs) ->
```
ldmberman: If there is no good way not to have this empy behaviour, I would put a brief explanation here.

apps/arweave/src/arweave.app.src
ldmberman: Updated.

PR #198, opened 2019-11-29
Make max_connections configurable

maxmellen:
This PR allows configuring the maximum number of concurrent connections for both the node API listener and the gateway listener.

The current defaults are:
- `max_connections`: 1024
- `max_gateway_connections`: 128

The gateway listener on port 80 whose only purpose is to redirect to port 443 has a constant number of maximum connections currently set to 10.

The startup script will also warn if `ulimit -n` returns less than 10000.

### DIFF
--- a/apps/arweave/src/ar.erl
+++ b/apps/arweave/src/ar.erl
@@ -139,7 +139,9 @@ show_help() ->
 			{"gateway (domain)", "Run a gateway on the specified domain"},
 			{"custom_domain (domain)", "Add a domain to the list of supported custom domains."},
 			{"requests_per_minute_limit (number)", "Limit the maximum allowed number of HTTP requests per IP address per minute. Default is 900."},
-			{"max_propagation_peers (number)", "How many peers to propagate blocks and transactions to. Default is 65."}
+			{"max_propagation_peers (number)", "How many peers to propagate blocks and transactions to. Default is 65."},
+			{"max_connections", "The number of connections to be handled concurrently. Its purpose is to prevent your system from being overloaded and ensuring all the connections are handled optimally. Default is 1024."},
+			{"max_gateway_connections", "The number of gateway connections to be handled concurrently. Default is 128."}
 		]
 	),
 	erlang:halt().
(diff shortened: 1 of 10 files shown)


apps/arweave/src/ar.erl
```diff
@@ -442,9 +450,15 @@ start(
 	),
 	ok = start_graphql(),
 	%% Start the first node in the gossip network (with HTTP interface).
-	ok = ar_http_iface_server:start(Port, GatewayOpts, [
+	ok = ar_http_iface_server:start([
 		{http_entrypoint_node, Node},
-		{http_bridge_node, Bridge}
+		{http_bridge_node, Bridge},
+		{port, Port},
+		{gateway, GatewayOpts},
```
ldmberman: Looks like we do not need this flag now, as well as the `gateway_opts` function.

Also, when checking the port above, we could just just do `is_binary(GatewayDomain)`?

apps/arweave/src/ar.erl
```diff
@@ -139,7 +139,9 @@ show_help() ->
 			{"gateway (domain)", "Run a gateway on the specified domain"},
 			{"custom_domain (domain)", "Add a domain to the list of supported custom domains."},
 			{"requests_per_minute_limit (number)", "Limit the maximum allowed number of HTTP requests per IP address per minute. Default is 900."},
-			{"max_propagation_peers (number)", "How many peers to propagate blocks and transactions to. Default is 65."}
+			{"max_propagation_peers (number)", "How many peers to propagate blocks and transactions to. Default is 65."},
+			{"max_connections", "The number of connections to be handled concurrently. Its purpose is to prevent your system from being overloaded and ensuring all the connections are handled optimally."},
```
ldmberman: I would specify the defaults here.

apps/arweave/src/ar.erl
maxmellen: Good catch! Though I removed it 😅 e8f9c451

apps/arweave/src/ar.erl
maxmellen: Done 👍 375e94a0

ldmberman:
@maxmellen should we do something about this warning?

```
Setting Ranch options together with socket options is deprecated. Please use the new map syntax that allows specifying socket options separately from other options.
```

PR #201, opened 2019-12-02
Deduct validated txs from IP blacklist counter

ldmberman:
Exclude successful requests with valid transactions from the
IP-based throttling, to avoid connectivity issues at the times
of excessive transaction volumes.

### DIFF
--- a/apps/arweave/src/ar_blacklist_middleware.erl
+++ b/apps/arweave/src/ar_blacklist_middleware.erl
@@ -6,6 +6,7 @@
 -export([start/0]).
 -export([reset/0, reset_rate_limit/1]).
 -export([ban_peer/2, is_peer_banned/1, cleanup_ban/0]).
+-export([decrement_ip_addr/1]).
 
 -include("ar.hrl").
 -include_lib("eunit/include/eunit.hrl").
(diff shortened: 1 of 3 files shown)


apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -273,7 +273,8 @@ handle(<<"POST">>, [<<"wallet">>], Req) ->
 handle(<<"POST">>, [<<"tx">>], Req) ->
 	TXJSON = ar_http_req:body(Req),
 	TX = ar_serialize:json_struct_to_tx(TXJSON),
-	case handle_post_tx(TX) of
+	{PeerIP, _Port} = cowboy_req:peer(Req),
```
jespern: Is there a way to move this up to the top of the function, so we don't unnecessarily read the body and attempt to parse it as JSON before performing the next actions?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -861,7 +863,11 @@ verify_mempool_txs_size(MempoolTXs, TX) ->
 			valid
 	end.
 
-handle_post_tx_accepted(TX) ->
+handle_post_tx_accepted(PeerIP, TX) ->
+	%% Exclude successful requests with valid transactions from the
+	%% IP-based throttling, to avoid connectivity issues at the times
+	%% of excessive transaction volumes.
+	ar_blacklist_middleware:increment_ip_addr(PeerIP, -1),
```
jespern: Prefer having a `ar_blacklist_middleware:decrement_ip_addr` method instead that does this.

apps/arweave/src/ar_blacklist_middleware.erl
```diff
@@ -74,10 +75,10 @@ reset_rate_limit(IpAddr) ->
 	ets:delete(?MODULE, {rate_limit, IpAddr}),
 	ok.
 
-increment_ip_addr(IpAddr) ->
+increment_ip_addr(IpAddr, Increment) ->
```
jespern: Maybe have a `increment_ip_addr(IpAddr) -> increment_ip_addr(IpAddr, 1);` as well so we don't update the code where the arg is `1`?

apps/arweave/src/ar_http_iface_middleware.erl
ldmberman: The body is currently parsed in a separate middleware. There is a WIP [branch](https://github.com/ArweaveTeam/arweave/pull/187) changing it, one of the prerequisites for "fast write".

PR #212, opened 2020-02-19
Optimize transaction data overhead on HTTP GET /tx/[TXHash] operations

Ayanda-D:
Not all operations on transactions require it's data to be present.
This change introduces a configurable HTTP header option for
specifying and controlling the type of formatting of the acquired
transaction record. Specifying `"x-tx-format"` header will dictate
whether or not, the retrieved transaction contains the data, or not.
Values can either be `"1"` or `"2"`, indicating whether transaction
data is included in the retrieved record, or not, respectively.

**NOTE:** Some gossip operations should also be further extended to
exclude data overheads as more `ansyc` coordination operations
are introduced in the network. These will be looked at case-by-case.

### DIFF
--- a/apps/arweave/src/ar_http_iface_tests.erl
+++ b/apps/arweave/src/ar_http_iface_tests.erl
@@ -279,6 +279,31 @@ get_non_existent_block_test() ->
 	{ok, {{<<"404">>, _}, _, _, _, _}}
 		= ar_httpc:request(<<"GET">>, {127, 0, 0, 1, 1984}, "/block/hash/abcd/hash_list").
 
+%% @doc Ensure transactions of format 2 type can be retrieved over the HTTP
+%% interface with no populated data, while retaining info on all other fields
+get_format_2_tx_with_no_data_test() ->
+	ar_storage:clear(),
+	[B0] = ar_weave:init(),
+	Node = ar_node:start([], [B0]),
+	ar_http_iface_server:reregister(Node),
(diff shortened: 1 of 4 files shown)


apps/arweave/src/ar_http_iface_tests.erl
```diff
@@ -279,6 +279,31 @@ get_non_existent_block_test() ->
 	{ok, {{<<"404">>, _}, _, _, _, _}}
 		= ar_httpc:request(<<"GET">>, {127, 0, 0, 1, 1984}, "/block/hash/abcd/hash_list").
 
+%% @doc Ensure transactions of format 2 type can be retrieved over the HTTP
+%% interface with no populated data, while retaining info on all other fields
+get_format_2_tx_with_no_data_test() ->
+	ar_storage:clear(),
+	[B0] = ar_weave:init(),
+	Node = ar_node:start([], [B0]),
```
ldmberman: Although there are still a few tests around which use this clunky setup, we have some helpers to make it better. You can do `{Node, _} = ar_test_node:start(B0)` in place of lines 287-291.

Also, sleeps can be avoided by waiting for a transaction - `ar_test_node:wait_until_txs(Node, [TX])` and a block - `ar_test_node:wait_until_height(Node, 1)`.

ldmberman:
@Ayanda-D I cherry-picked you commits into the branch, thank you!

PR #213, opened 2020-02-21
Apply transaction overhead removal on HTTP interface block acquisition calls

Ayanda-D:
This change applies the removal of transaction data overhead on HTTP 
operations acquiring blocks, which would've previously populated all 
transactions with data. All `ar_http_iface_middleware` block acquisitions
calls now specify set the`with_tx_header` flag, for removal/avoidance
of transaction data overhead.

The following block utilities are updated, but retain API compatibility
for pre 2.0 dependant operations (e.g. in tests):
- `ar_node:get_block/3`
- `ar_node:get_current_block/2`
- `ar_http_iface_client:get_block/3`
- `ar_http_iface_client:get_current_block/2`
- `ar_http_iface_client:get_full_block/4`

Transaction data filters, namely `no_tx`, `with_tx_header` and `with_tx_data`
are also introduced for attaining this control. The `ar_downloader` will acquire
all transaction data through an explicit `ar_node_utils:get_full_block/4` call,
specifying the `with_tx_data` filter.

**NOTE:** This is partly related to/a continuation of, #212.


apps/arweave/src/ar_tx.erl
```diff
@@ -8,6 +8,7 @@
 -export([generate_chunk_tree/1, generate_chunk_tree/2, generate_chunk_id/1]).
 -export([chunk_binary/2]).
 -export([chunks_to_size_tagged_chunks/1, sized_chunks_to_sized_chunk_ids/1]).
+-export([tx_format_to_http_header/1, tx_http_format/1]).
```
vkatsuba: Looks like `tx_http_format/1` is used inside of current module only - no need to export

apps/arweave/src/ar_tx.erl
```diff
@@ -431,6 +432,13 @@ chunks_to_size_tagged_chunks(Chunks) ->
 sized_chunks_to_sized_chunk_ids(SizedChunks) ->
 	[{ar_tx:generate_chunk_id(Chunk), Size} || {Chunk, Size} <- SizedChunks].
 
+tx_format_to_http_header(TXFormat) ->
+	[{<<"x-tx-format">>, tx_http_format(TXFormat)}].
+
+tx_http_format(?WITH_TX_DATA) -> ?TX_WITH_DATA_HTTP_FORMAT;
```
vkatsuba: Looks like, this function can be rewritten as:
```erlang
tx_http_format(?WITH_TX_HEADER) -> ?TX_WITHOUT_DATA_HTTP_FORMAT;
tx_http_format(_) -> ?TX_WITH_DATA_HTTP_FORMAT.
```

apps/arweave/src/ar_tx.erl
Ayanda-D: True. However, `tx_http_format/1` should be a usable utility for translating from an internal tx-format to the external http format (for internal purposes as well). It's not always the case that this function will only be used for creating an HTTP header via `tx_format_to_http_header/1`.

apps/arweave/src/ar_tx.erl
Ayanda-D: Indeed. But we're only making this explicit so that the intent is clear that we're handling what we know and what we expect in each clause. This is intentional. A handle all clause definitely works, but will also unnecessarily hide & abstract the tx-format protocol which we're expecting.

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -123,9 +124,11 @@ add_peer(Peer) ->
 
 %% @doc Get a peers current, top block.
 get_current_block(Peer) ->
-	get_current_block(Peer, get_block_index(Peer)).
-get_current_block(Peer, BI) ->
-	case get_full_block([Peer], hd(BI), BI) of
+	get_current_block(Peer, ?WITH_TX_DATA).
```
ldmberman: Could we get rid of this method along with the part of `ar_node:get_current_block/1` that calls it? As far as I can see, `ar_http_iface_client:get_current_block/..` is only used by `ar_node:get_current_block`, which in turn is only used once in `ar_http_iface_tests` (when an IP is passed, not a PID).

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -134,7 +137,9 @@ get_current_block(Peer, BI) ->
 
 %% @doc Retreive a block by height or hash from a remote peer.
 get_block(Peer, ID, BI) ->
-	case get_full_block([Peer], ID, BI) of
+	get_block(Peer, ID, BI, ?WITH_TX_DATA).
+get_block(Peer, ID, BI, TXFormat) ->
```
ldmberman: It looks like this one is also only used in `ar_http_iface_tests`, should we replace `get_full_block` with it? This would require to change all the places that call `get_full_block` then, but it might be worth it.

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -544,8 +554,12 @@ handle_block_response(Peer, _Peers, {ok, {{<<"200">>, _}, _, Body, _, _}}, block
 		Handled ->
 			Handled
 	end;
-handle_block_response(Peer, Peers, {ok, {{<<"200">>, _}, _, Body, _, _}}, {full_block, BI}) ->
-	case catch reconstruct_full_block(Peer, Peers, Body, BI) of
+handle_block_response(_, _, Response, _) ->
```
ldmberman: Where is the new `handle_block_response/4` used?

Ayanda-D: here https://github.com/ArweaveTeam/arweave/pull/213/files/4f1d52ab92356b126a45d0acfd229eb889532196#diff-bbfa4040e5f577553ef8de6f57e628a9R573-R574 - if response is not 200, ignore format and fall back on `handle_block_response/4`

apps/arweave/src/ar_http_iface_client.erl
Ayanda-D: I'm sure we can. I'll update

apps/arweave/src/ar_http_iface_client.erl
ldmberman: Ah, I see, thanks.

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -315,31 +322,34 @@ get_encrypted_full_block(Peer, Hash) when is_binary(Hash) ->
 		)
 	).
 
-get_txs(Peers, MempoolTXs, B) ->
+get_txs(_Peers, _MempoolTXs, _B, ?NO_TX) -> {ok, []};
```
ldmberman: This would override the list of txids in the block header with an empty list.

Ayanda-D: Ok. So a caller specifying `?NO_TX` will mean retain TXIDs, and if acquired block already has full TXs, retain & return them as well, but just don't fetch TXs in the current call. I'll update.

ldmberman: Correct, except for the acquired block is not supposed to ever have full transactions here.

ldmberman:
Cherry-picked into `feature/fast-write`, thank you!

PR #216, opened 2020-03-02
Asynchronous active/passive downloading of v2 blocks on peer nodes

Ayanda-D:
This change introduces capabilities for peer nodes with blocks 
comprising of format-2 transactions (with stripped of data) to
download full transactions and re-fill it’s local block copy as well
as persist it. Downloads may be **active** or **passive**, controlled by
the `active_download` configuration flag, which when set to `true`
means **active** mode and when `false`, downloads are **passive**.

Downloads of full blocks & transactions can be triggered in 3 ways:

1. When a node joins the network, through `ar_join`
2. Actively, when `active_download` = `true`
3. Passively (e.g. through a custom CLI command)

When a new block has been downloaded on a peer node, along its
full transactions, a gossip message `{drop_downloaded_txs, DTXS}`,
where `DTXS` are the downloaded transactions, emitted for cleaning
up the `txs_pending_downloads` collection in the local node’s state.
This is to allow for `on-demand` scheduling of block downloads, to be
performed only when necessary.

Related: #212, #213


apps/arweave/src/ar_node_state.erl
```diff
@@ -38,7 +38,8 @@ start() ->
 		{cumulative_diff, 0},               % Sum of the difficulty squared along the current weave
 		{hash_list_merkle, <<>>},           % The Merkle root of the current hash list
 		{block_txs_pairs, []},              % List of {BH, TXIDs} pairs for last ?MAX_TX_ANCHOR_DEPTH blocks
-		{mempool_size, 0}                   % Memory pool size
+		{mempool_size, 0},                  % Memory pool size
+		{txs_pending_download, sets:new()}  % set of txs pending download (only TXIDs)
```
ldmberman: Not sure if I understand the purpose of this set. It should be sufficient to go from the most recent block to the oldest and download the missing transactions from them, we do not really need them until they are mined.

apps/arweave/src/ar_downloader.erl
```diff
@@ -48,18 +48,29 @@ fill_to_capacity(Peers, ToWrite, BI) ->
...
-				[] ->
-					%% All of the TXs are found locally. Continue...
-					fill_to_capacity(
-						Peers,
-						lists:delete(RandHash, ToWrite),
-						NewBI
-					);
-				_ ->
+		B = #block{txs = TXS0} ->
+			{TXS, TXIDs} =
```
ldmberman: `ar_storage:read_block` always returns a block without transactions, just transaction identifiers. The `[TX || TX <- ar_storage:read_tx(B#block.txs), not is_record(TX, tx)]` was misleading here.

apps/arweave/src/ar_config.hrl
```diff
@@ -44,7 +44,8 @@
 	webhooks = [],
 	max_connections = 1024,
 	max_gateway_connections = 128,
-	max_option_depth = 8
+	max_option_depth = 8,
+	active_download = false
```
ldmberman: Miners need to download as many transactions as they can since it makes mining more efficient so I would make it true by default. In general, it does make sense to have it configurable because there are situations when you want to run a node which would participate in the gossip but not store data.

apps/arweave/src/ar_node_state.erl
Ayanda-D: ok, purpose is for controlling & determining whether the downloader can be scheduled for execution here https://github.com/ArweaveTeam/arweave/pull/216/commits/72e1d2f38cdc460137431ddc9916a347adf43413#diff-47fb75d8dee1a57e1c5e1a6dfae18c09R149-R150 or not - with its disk activity posing as being potentially "expensive", and with only format-2 TX (IDs) qualifying to be in this set. If empty, we don't even attempt to interact & trigger the downloader https://github.com/ArweaveTeam/arweave/pull/216/commits/72e1d2f38cdc460137431ddc9916a347adf43413#diff-47fb75d8dee1a57e1c5e1a6dfae18c09R738

ldmberman:
Some changes were incorporated in feature/fast-write, thank you!

PR #219, opened 2020-03-10
Add Travis-CI

vkatsuba:
Travis-CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: erlang
+addons:
+    apt:
+      packages:
+      - sqlite3
+install:
+    - epmd -daemon
+otp_release:
+    - 22.2
+# blocklist
+branches:
+    except:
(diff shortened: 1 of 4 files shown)


ldmberman:
I would suggest a slightly different way of determining the build outcome - we can exit with a non-zero code if the tests fail. To kill the background beam, the exit might be trapped. I've tried this locally and it worked fine:

Bildschirmfoto 2020-03-11 um 09 31 14

What do you think?

vkatsuba:
> I would suggest a slightly different way of determining the build outcome - we can exit with a non-zero code if the tests fail. To kill the background beam, the exit might be trapped. I've tried this locally and it worked fine:
> 
> Bildschirmfoto 2020-03-11 um 09 31 14
> 
> What do you think?

Done. Yes, this is a good point, thanks.

PR #224, opened 2020-04-03
Reduce the state passed on to mining workers

ldmberman:
Each mining worker currently consumes around 85 MB of heap memory as of the current weave height
of ~400k. On big machines with >20 miners it contributes gigabytes to the memory footprint.
The commit reduces this amount substantially.


ldmberman:
Closing, merged into `2.0.0.1-rc`.

PR #228, opened 2020-04-09
Storing blocks shadows in memory

vkatsuba:


vkatsuba:
Closed by bad working speed performance. The one of reasons is the lists of blocks in memory can be a big for sending from process to process.

PR #229, opened 2020-04-10
Fix a bug where nodes sync v2 data too slowly

ldmberman:


mattdf:
this looks good to me, i can't see any issues with the surrounding code.

if i'm reading the fix correctly, this code path was mistakenly adding backoff any time a tx has an empty data field right?

ldmberman:
yes, retries with backoff are needed still, I think, but after we do not find the data _anywhere_, not just on one peer

PR #230, opened 2020-04-10
Use a >20x faster Base64 encoding/decoding lib

ldmberman:


vkatsuba:
> actually, the new lib throws exceptions for invalid b64 code, unlike the old one that always returns an array no matter what
> 
> a catch needs to be added to keep the same behaviour as the old lib for ar_util:decode

Looks like the old library was used `base64:decode/1` and if we try decode `<<"h">>` will be exceptions eg:
```erlang
1> ar_util:decode(<<"h">>). 
** exception error: no function clause matching 
                    base64:decode_binary(<<>>,<<>>,33) (base64.erl, line 362)
2> base64:decode(<<"h">>).
** exception error: no function clause matching 
                    base64:decode_binary(<<>>,<<>>,33) (base64.erl, line 362)
```
I not sure that we need to catch this exception, if something wrong with binary we need to detect it. Here used the same solution approach as in previous library - let the process die

mattdf:
The old code uses `base64url:decode`, not `base64:decode`

vkatsuba:
> The old code uses `base64url:decode`, not `base64:decode`

Correct, but this lib use `base64:decode` - https://github.com/dvv/base64url/blob/master/src/base64url.erl#L23

mattdf:
Ah, I see, I was running the wrong version of the lib. Then yes the behaviour should be the same.

mattdf:
Should we also be porting forward the `urldecode_digit` function? I don't think the new c lib does those replacements, does it?

vkatsuba:
> Should we also be porting forward the `urldecode_digit` function? I don't think the new c lib does those replacements, does it?

Actually the fork was changed https://github.com/ArweaveTeam/b64fast/commit/508a61c9af4f6a8bd3ff105ceae2c0f0cf38c6d4, this mean that `urldecode_digit` was moved in `C` code

mattdf:
Okay I found a difference, old lib:

```
ar_util:decode(<<"PDw/Pz8+Pg==">>).    
<<"<<???>>">>
```

New lib:

```
ar_util:decode(<<"PDw/Pz8+Pg==">>).    
<<"<<ÿ??ÿ>">>
```

The new lib mangles decoding if there are non-base64 URL chars in the b64 string, whereas the new lib replaces them properly. We still need to do a first-pass replacement in the erlang or the C code, instead of just updating the decoding table.

mattdf:
Actually, just updating the decoding table as I stated here would fix it: https://github.com/ArweaveTeam/b64fast/commit/508a61c9af4f6a8bd3ff105ceae2c0f0cf38c6d4#r38445714

mattdf:
The differences are more drastic in the new lib actually. It just decodes any non-b64 character to `0xFF`, unlike the old lib that would just throw an exception if you gave it non-b64 chars:

```
ar_util:decode(<<"PDw:Pz8.Pg==">>).
<<"<<ÿ??ÿ>">>
```
Probably best to add a filter that throws if there are completely invalid b64 chars fed to the decode function.

ldmberman:
Hey @mattdf, good catch!

I think, updating the decoding table is sufficient - yes, we still need to decode `/` and `+` as it used to do, not just decode `_` as if it was `/` and `-` as if it was `+`. Updating the decoding table then gives me the matching outputs on your `<"PDw/Pz8+Pg==">>` example.

PR #232, opened 2020-04-15
Add support for DNS peers

mattdf:
simple PR that adds domain name support to the "peer" argument, I've also made the top level function use safe_parse_peer so that the node doesn't die if one of the peers fails to resolve

### DIFF
--- a/apps/arweave/src/ar.erl
+++ b/apps/arweave/src/ar.erl
@@ -178,7 +178,13 @@ parse_cli_args(["init"|Rest], C) ->
 parse_cli_args(["mine"|Rest], C) ->
 	parse_cli_args(Rest, C#config { mine = true });
 parse_cli_args(["peer", Peer|Rest], C = #config { peers = Ps }) ->
-	parse_cli_args(Rest, C#config { peers = [ar_util:parse_peer(Peer)|Ps] });
+	case ar_util:safe_parse_peer(Peer) of
+		{ok, ValidPeer} ->
+			parse_cli_args(Rest, C#config { peers = [ValidPeer|Ps] });
+		{error, _} ->
+			io:format("Peer ~p invalid ~n", [Peer]),
+			parse_cli_args(Rest, C)
+	end;
 parse_cli_args(["content_policy", File|Rest], C = #config { content_policy_files = Files }) ->
 	parse_cli_args(Rest, C#config { content_policy_files = [File|Files] });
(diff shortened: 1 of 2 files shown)


apps/arweave/src/ar_util.erl
```diff
@@ -124,11 +124,12 @@ parse_peer("") -> throw(empty_peer_string);
...
 parse_peer(Str) when is_list(Str) ->
-	case io_lib:fread("~d.~d.~d.~d", Str) of
-		{ok, [A, B, C, D], PortStr} ->
+    [Addr, PortStr] = parse_port_split(Str),
+    case inet:getaddr(Addr, inet) of
+		{ok, {A, B, C, D}} ->
 			{A, B, C, D, parse_port(PortStr)};
-		{error, _} ->
-			throw({invalid_peer_string, Str})
+		{error, Posix} ->
```
ldmberman, line 131: Why "Posix"?

ldmberman:
There is a failure in the build:

```
**error:{assertEqual,[{module,ar},
              {line,729},
              {expression,"Value"},
              {expected,[{ok,{5,6,7,8,9}},{ok,{1,2,3,4,1984}}]},
              {value,[{5,6,7,8,9},{1,2,3,4,1984}]}]}
  output:<<"">>
```

apps/arweave/src/ar_util.erl
mattdf: it's a posix return error code, i just took the name from the erlang lib description

apps/arweave/src/ar.erl
```diff
@@ -178,7 +178,7 @@ parse_cli_args(["init"|Rest], C) ->
 parse_cli_args(["mine"|Rest], C) ->
 	parse_cli_args(Rest, C#config { mine = true });
 parse_cli_args(["peer", Peer|Rest], C = #config { peers = Ps }) ->
-	parse_cli_args(Rest, C#config { peers = [ar_util:parse_peer(Peer)|Ps] });
+	parse_cli_args(Rest, C#config { peers = [ar_util:safe_parse_peer(Peer)|Ps] });
```
vkatsuba, line 181: `ar_util:safe_parse_peer/1` will return `{ok, Peer}` instead `Peer` - looks like by this reason you have a crash in tests

apps/arweave/src/ar_util.erl
vkatsuba: `posix()` is a spec for names(atoms) of POSIX error codes. More better name can be `Reason` or you can skip this `Reason` as was before

PR #234, opened 2020-04-26
Fix a bug where an empty node may fail to join the network

ldmberman:


ldmberman:
[arweave-2.0.0.3.tar.gz](https://github.com/ArweaveTeam/arweave/files/4535244/arweave-2.0.0.3.tar.gz)

PR #237, opened 2020-05-04
WIP: Implement data sync server initialization and processing

Ayanda-D:
This PR introduces the processing logic of the `ar_data_sync`.
The `ar_data_sync` is a gen-server process responsible for
tracking and synchronizing specific chunks on the weave as
desired (through specifying desire bytes on the weave range).
The data sync initialization and processing callbacks are
implemented as part of this PR.

Also included, is a pluggable KV store, `ar_kv` with a default
`ar_kv_rocksdb` backend engine. More backend engines could
be introduced in future, e.g. `ar_kv_mnesia`, etc

### DIFF
--- a/apps/arweave/src/ar_block.erl
+++ b/apps/arweave/src/ar_block.erl
@@ -11,6 +11,7 @@
 -export([generate_block_data_segment/1, generate_block_data_segment/3, generate_block_data_segment_base/1]).
 -export([generate_hash_list_for_block/2]).
 -export([generate_tx_root_for_block/1, generate_size_tagged_list_from_txs/1]).
+-export([generate_size_tagged_tx_ids/1]).
 -export([generate_tx_tree/1, generate_tx_tree/2]).
 -export([compute_hash_list_merkle/2]).
 
@@ -82,6 +83,24 @@ generate_size_tagged_list_from_txs(TXs) ->
 				end,
 				{0, []},
 				lists:sort(TXs)
+			)
+		)
+	).
+
+generate_size_tagged_tx_ids(TXs) ->
+	lists:reverse(
+		element(
+			2,
+			lists:foldl(
+				fun(TX = #tx{id = TXID}, {Pos, List}) ->
+						End = Pos + get_tx_data_size(TX),
+						{End, [{TXID, End} | List]};
+				   (TXID, {Pos, List}) when is_binary(TXID) ->
+						End = Pos + get_tx_data_size(TXID),
+						{End, [{TXID, End} | List]}
+				end,
+				{0, []},
+				TXs
 			)
 		)
 	).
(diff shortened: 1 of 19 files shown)


apps/arweave/src/ar_block.erl
```diff
@@ -86,6 +87,24 @@ generate_size_tagged_list_from_txs(TXs) ->
...
+generate_size_tagged_tx_ids(TXs) ->
+	lists:reverse(
+		element(
+			2,
+			lists:foldl(
+				fun(TX = #tx{id = TXID}, {Pos, List}) ->
+						End = Pos + get_tx_data_size(TX),
+						{End, [{TXID, End} | List]};
+				   (TXID, {Pos, List}) when is_binary(TXID) ->
+						End = Pos + get_tx_data_size(TXID),
```
domhaobaobao, line 99: can we use a different name for functions that handle TXID?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -101,16 +105,216 @@ sync_random_chunk() ->
...
 	%% block index. Store the new state.
-	State = #state{},
-	{ok, State}.
+	gen_server:cast(?MODULE, {init, BI}),
+	{ok, #state{
+			db = ?INIT_KV("kv-data-sync"),
+			tx_root_index = ?INIT_KV("kv-tx-root-index"),
+			block_offset_index = ?INIT_KV("kv-block-offset-index"),
+			tx_index = init_tx_index(MaintainTXIndex),
+			maintain_tx_index = MaintainTXIndex}}.
```
domhaobaobao, line 114: is any of the above intensive operations to be placed in the above gen_server cast?

apps/arweave/src/ar_block.erl
```diff
@@ -86,6 +87,24 @@ generate_size_tagged_list_from_txs(TXs) ->
 		)
 	).
 
+generate_size_tagged_tx_ids(TXs) ->
+	lists:reverse(
+		element(
+			2,
+			lists:foldl(
+				fun(TX = #tx{id = TXID}, {Pos, List}) ->
+						End = Pos + get_tx_data_size(TX),
```
ldmberman, line 96: You can just do `#tx.data_size` like it's done in `generate_size_tagged_list_from_txs`.

apps/arweave/src/ar_config.hrl
```diff
@@ -43,7 +43,8 @@
 	webhooks = [],
 	max_connections = 1024,
 	max_gateway_connections = 128,
-	max_poa_option_depth = 8
+	max_poa_option_depth = 8,
+	kv_engine = ar_kv_rocksdb
```
ldmberman, line 47: We also update `ar_config.erl` every time we add a new command flag because a flag can be specified in a JSON config file, but I would rather not introduce this flag at all for now, since we only support one kv backend.

apps/arweave/src/ar_data_sync.erl
```diff
@@ -101,16 +105,216 @@ sync_random_chunk() ->
...
+				chunk_ids_by_tx_id = ChunkIDsByTXID,
+				chunk_ids_by_tx_root = ChunkIDsByTXRoot,
+				tx_roots_with_proofs_by_chunk_id = TXRootsPOAByChunkID,
+				block_offset_by_tx_root = BlockOffsetByTXRoot},
+		maintain_tx_index  = MaintainTXIndex,
+		chunks_index = ChunksIndex} = State,
+
+	{Reply, FinalState} =
+		case map:get(TXRoot, BlockOffsetByTXRoot) of
+			true ->
```
domhaobaobao, line 204: this should not match to true

apps/arweave/src/ar_data_sync.erl
Ayanda-D: these just acquire the kv-store handle, so they should be fine. the intensive
operation we foresee will be in `load_state_from_disk/1`. that will be executed
non-blocking to the process `init`. 👍

PR #259, opened 2020-08-13
Simplify the node state management

ldmberman:
The main motivation was to decrease the overhead imposed on the transaction processing and caused
by an excess of memory copying between processes. See the detailed list of changes below.

- The node state is moved from the ETS table owned by ar_node_state to ar_node_worker's and
  ar_node's process stacks. ar_node_worker manages state updates, ar_node maintains a read-only
  copy. Reading data from the state (e.g. when serving the list of pending transactions via HTTP
  API) now involves only a single copy of the memory from the ar_node stack to the handler
  process. Previously, three copies were made - ETS -> ar_node_state, ar_node_state -> ar_node,
  and ar_node -> handler process. As a bonus, this fixes the small memory leak caused by the
  memory fragmentation as ETS tables are not defragmented.
- Some HTTP handlers and the corresponding ar_node functions are changed to minimize the amount
  of data passed from ar_node to the request handler.
- The tasks of accepting an alternative fork and the mined block are prioritized over processing
  the transaction gossip and updating transaction statuses.
- Unused functions are removed from ar_node and ar_node_worker.


ldmberman:
[Merged](https://github.com/ArweaveTeam/arweave/commit/355f4098c661663dbb0e24e0f50a3e2e88d5b376).

PR #264, opened 2020-09-14
Timestamp commits using Gitstamp

artob:
As per https://github.com/ArweaveTeam/arweave-js/pull/54. Sam says we want [Gitstamp](https://gitstamp.dev) for all repositories, so here goes 😃


ldmberman:
Thank you!

PR #275, opened 2020-10-29
fix sub-query by block height

goomario:
```
Error in process <0.12824.0> on node 'arweave@127.0.0.1' with exit value:
{badarg,
    [{erlang,binary_to_integer,[557419],[]},
     {ar_http_iface_middleware,find_block,2,
...
```

### DIFF
--- a/apps/arweave/src/ar_http_iface_middleware.erl
+++ b/apps/arweave/src/ar_http_iface_middleware.erl
@@ -1697,7 +1697,7 @@ search_in_block_index(H, BI) ->
 %% @doc Find a block, given a type and a specifier.
 find_block(<<"height">>, RawHeight) ->
 	BI = ar_node:get_block_index(whereis(http_entrypoint_node)),
-	ar_storage:read_block(binary_to_integer(RawHeight), BI);
+	ar_storage:read_block(RawHeight, BI);
 find_block(<<"hash">>, ID) ->
 	ar_storage:read_block(ID).
 


apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -1697,7 +1697,7 @@ search_in_block_index(H, BI) ->
 %% @doc Find a block, given a type and a specifier.
 find_block(<<"height">>, RawHeight) ->
 	BI = ar_node:get_block_index(whereis(http_entrypoint_node)),
-	ar_storage:read_block(binary_to_integer(RawHeight), BI);
+	ar_storage:read_block(RawHeight, BI);
```
ldmberman, line 1700: Hey @goomario, thank you for looking into it.

This way if you ask for /height/2, it will search for a block with the hash `<<"2">>`, we need to do smth like [this](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_http_iface_middleware.erl#L1668) here instead.

goomario: Maybe use this:
```
find_block(<<"height">>, RawHeight) when is_integer(RawHeight) ->
	BI = ar_node:get_block_index(whereis(http_entrypoint_node)),
	ar_storage:read_block(RawHeight, BI);
find_block(<<"height">>, RawHeight) ->
	BI = ar_node:get_block_index(whereis(http_entrypoint_node)),
	ar_storage:read_block(binary_to_integer(RawHeight), BI);
```

goomario: It's work fine, I tested.

ldmberman: `find_block(<<"height">>, RawHeight) when is_integer(RawHeight) ->` - this should never match as RawHeight is a binary.

PR #280, opened 2020-11-10
Improvements

halturin:
This PR is intended to make code close to OTP design

 - reworked application bootstrapping (most of the code moved close to the usage place)
 - rewritten ar_node and ar_bridge (now they are OTP designed gen_servers)
 - introduced the main supervisor ar_sup
 - introduced ar_watchdog
 - tests have been rewritten to suppor OTP behavior of ar_node/ar_bridge


halturin:
Going to create another one with PR to the master

PR #281, opened 2020-11-27
Improvement/preparation for succinct proofs of random access consensus

ldmberman:


bin/benchmark
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
halturin, line 1: Dont you want to use `/usr/bin/env bash`?

scripts/benchmark
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
halturin, line 1: The same. It would be more flexible to use `/usr/bin/env bash` for that

PR #284, opened 2020-12-31
Docker + Travis-CI Integration

TheLoneRonin:
### Description

- Adds a Dockerfile that builds and compiles Arweave.

- Updates Travis-CI to build Arweave with Ubuntu Focal (20.04)

- Provides Documentation in `README.md` to build, compile and develop Arweave with Docker.

- The following is a sample build running the Travis build matrix: https://travis-ci.org/github/TheLoneRonin/arweave/builds/752315807

### Notes

It may be a good idea to add other badges to the top of the repository. However, we may want other badges too ie: license, AR version etc.


ldmberman:
Hi @TheLoneRonin, thank you for the update.

We are using Travis CI in fact. It does not work for you because you specified OTP 23, which is not compatible with [the current versions of the libraries](https://github.com/ArweaveTeam/arweave/pull/250). OTP 22 should work.

As per the setup, I believe it's important to configure external storage (a volume/bind mount). Only using the container's writable layer has very limited use, even for tests.

@tudorfil9 do you have any thoughts or developments to share?

tudorfil9:
I agree, there should be a separate storage option. The container layer is very small. 

Have a look at this Arweave Docker Image from DockerHub. 

https://hub.docker.com/layers/116646671/tudorfil9/arweave-node/latest/images/sha256-9df66b9b4f705eda5a1f27908ca21c4a4e08d13d33e8b52d252d77896ce19901?context=explore

TheLoneRonin:
@ldmberman, @tudorfil9,

OTP 23.0 compiles Arweave on a fresh install of Ubuntu 20.04.

I will update this PR with 22.0 and see if Travis works.

---

In terms of the storage option. Agree absolutely I should document how to mount volumes.

I think there should be two documentation options for volumes.

1. How to use a volume it using the `--mount` flag.

2. How to mount an existing volume.

Furthermore, I should add snippets on how to mount it to a different drive... ie: using `lsblk` to find

And of course, link to existing Docker documentation for reference purposes.

TheLoneRonin:
@ldmberman, @tudorfil9, now I remember why I changed it from OTP 22.2...

This bizarre error came up: https://travis-ci.org/github/TheLoneRonin/arweave/jobs/752828446

tudorfil9:
Hey @TheLoneRonin it looks like your key for otp version might not be right. 

Can you check your available versions and change the otp_release key to 22.2.4 for example ?

[](https://docs.travis-ci.com/user/languages/erlang/#choosing-otp-releases-to-test-against)

TheLoneRonin:
@tudorfil9, yea, Travis is acting very weird. Not sure the best approach to resolving this.

Then ontop of that we'd have to tackle the `-lerl_interface` issue too once this is resolved.

https://travis-ci.org/github/ArweaveTeam/arweave/jobs/752840013

Issue #283, opened 2020-12-30
Transaction doesn't appear in pending list


APshenkin:
I build transaction to send some funds between accounts

```
{
    "format": 2,
    "id": "PrTlOjxQYJZmOJjEwcdz6j_-LaxaUmNa3THmnobSElM",
    "last_tx": "rxchIjzzdHzhBi6iJ7XjVURdr14hv_VxxMdfIb8-EdtNzfUB9mEZ_GY2d6AVCKwR",
    "owner": "ILmrC5N0JSvpgEGi72lTdDBa00bM+ERJ0cLwpDoVfAhoAHnYzoqSafSGQtl1Uo5+mjweX224MotXKJkdFB1SjycHE9Nwton7Afgg3RQ0gOVCGJ5DZqsiAsHOjG8Ow9FZZMsRfW6nIfmr12GTZUgQpgXMiDO6PynA76foZCbisyUagWqWqu00g1ShkgUUhBG5687bHqxZwvAAnauyacyA2BxFILugx/UTQ20BK9SghbJTMntjJbU3KsTPVvO1lSrcRZDUFaEcIEtcCnwfGBRHxmE4jIp0XoDZZi4nIPAQLZDpfLdX1WlTR2o7xhJjS8JN8c7P8C6FQHHFl07YzT46HnLggRZGYSijqX0gpZ9/bIi1J+fM8ysWqYt6DkLXohwur10NyXUtvk1aCPjvGiXFtlaO9CRRSGRRNIKehcHifq5TJHMgAt/x+ztfu0HiChRk25w/cf+3jDb4nZZ5g8cXNuFg87SDYva3KZPrax87OslOi1T7z6N66wEJtStyuZZGq1g8KtkfubEJb0qSidYsNEMV+6DCUmS1PCc20LQ/c0SH9oATsFuOlcq74LazLMQ85xWYkBUo+Pd6JFZemjcpMhExSfvskv4Ukfv6OnlYkkFVABARzezbBfJkTNMO/0uAjN2RT2cRtW0pDx820dM+juRjCZ6zWMBru25CtBOMO8c=",
    "tags": [],
    "target": "lLKrX4uT5bEewbWmSA7cEjEz0L1oqOahdTCc_5p7TDk",
    "quantity": "1000",
    "data_size": "0",
    "data_root": "",
    "reward": "16461774",
[...]

PR #286, opened 2021-01-04
Fix public bit size in RSA PSS signature check

onesk:
Adjusts the PSS padding to correct length in case modulus bit length is less than 4096 - the byte size should be rounded up, not down.

A test case for this is evidenced in #283. Arweave-js would succesfully verify said signature via crypto.subtle, but Erlang implementation fails to do so. Not only it rejects the perfectly valid signature, it also crashes the verification process.


ldmberman:
Hi @onesk, great catch, thank you very much! 👏 We need to include it in the next hard fork.

ldmberman:
Added [here](https://github.com/ArweaveTeam/arweave/commit/ad5d0ac8dda520fa515ead7de2417fb45862ad4d), thank you!

PR #290, opened 2021-01-26
Update README.md

alexander-morris:


README.md
```diff
@@ -26,6 +26,8 @@ For more information, refer to the [mining guide](https://docs.arweave.org/info/
 - CMake
 - SQLite3 headers (libsqlite3-dev on Ubuntu)
 
+Install all of these in one line on Ubuntu with `sudo apt install erlang build-essential clang cmake libsqlite3-dev`
```
samcamwilliams, line 29: How about something like?

You can install all of these dependencies on Ubuntu using the following command:
```
sudo apt install erlang build-essential clang cmake libsqlite3-dev
```

PR #309, opened 2021-07-06
Support alternative transaction signing methods

CDDelta:
This PR adds support for signing transactions with the secp256k1 ECDSA and ed25519 EdDSA schemes. This change is entirely backwards compatible with the existing signing method, the process for signing transactions with RSA remains the same.

Transactions can now include an optional `signature_type` field which can be `PS256_65537`, `ES256K` or `Ed25519` to specify their signing scheme. Transactions using the new schemes should include the `signature_type` field as the last leaf used to generate the Merkle root used as the transaction signature message. This should not be done for the pre-existing scheme `PS256_65537`.

Transactions signed with secp256k1 ECDSA must specify a compressed `owner` (public key). Uncompressed `owner`s are rejected to avoid confusion with a single EC wallet having two possible addresses.

`signature_type` field values correspond to those defined in the [IANA Jose assignment database](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms) and [IETF RFC8410](https://datatracker.ietf.org/doc/html/rfc8410#section-8) with an extension for `PS256` to also define the RSA public exponent.


apps/arweave/src/ar_tx.erl
```diff
@@ -236,12 +246,14 @@ signature_data_segment_v1(T) ->
...
-	NewTX = TX#tx{ owner = PubKey },
+sign(TX, PrivKey, {_, Owner}, SignatureDataSegment) ->
+	NewTX = TX#tx{ owner = Owner },
 	Sig = ar_wallet:sign(PrivKey, SignatureDataSegment),
 	ID = crypto:hash(?HASH_ALG, <<Sig/binary>>),
 	NewTX#tx {
-		signature = Sig, id = ID
+		id = ID,
+		signature = Sig
+		% signature_type = KeyType
```
CDDelta, line 256: We need to uncomment this line to allow node operators to send transactions signed with ECDSA or EdDSA via the `/unsigned_tx` endpoint but that would require that we introduce a height parameter to this function.

What if we just enabled this after the fork?

apps/arweave/src/ar_tx.erl
```diff
@@ -440,21 +454,42 @@ ends_with_digit(Data) ->
 	LastByte = binary:last(Data),
 	LastByte >= 48 andalso LastByte =< 57.
 
+verify_signature_type(#tx { format = Format, signature_type = TXSigType }, Height) ->
+	case Height < ar_fork:height_2_5() of
+		true -> Format > 1;
```
CDDelta, line 459: Should `ar_serialize:json_struct_to_tx/1` return a dummy signature type so we can catch invalid `signature_type` values here?

CDDelta (commented):
Have some questions here.

apps/arweave/src/ar_tx.erl
ldmberman: signature_type=undefined essentially means RSA now so I would set the signature type right away on deserialization (or as a #tx record default if the tx is created internally).

Regarding `Format > 1;`, we are not dropping support for v1 transactions in 2.5, we just need to assert the type is one of the three supported types.

apps/arweave/src/ar_tx.erl
ldmberman: Yeah, I do not think (m)any people are using this endpoint. Anyway, it won't be difficult to add support for other signatures later at any moment.

apps/arweave/src/ar_tx.erl
CDDelta: > signature_type=undefined essentially means RSA now so I would set the signature type right away on deserialization (or as a #tx record default if the tx is created internally).
I don't believe we can do this as we need to know whether or not `signature_type` was defined when the tx was signed so we can check its signature appropriately.

> Regarding Format > 1;, we are not dropping support for v1 transactions in 2.5
Oops, yeah. Will fix that.

> we just need to assert the type is one of the three supported types.
From what I can tell, right now when someone sends in a tx with an invalid `signature_type` we'll get an error in `ar_serialize:json_struct_to_tx/1` as none of the case conditions will match. That was why I was wondering if it would make sense to return a dummy value like an `invalid_signature` atom when deserializing so `ar_tx:verify_signature_type/2` can validate things appropriately. What do you think?

ldmberman: We need to ensure compatibility with the existing clients; therefore, if there is no `signature_type` field we need to consider it an RSA-signed transaction.

Hm, if we detect an invalid signature in `json_struct_to_tx` already, why pass it on? Obviously, `ar_tx:verify` must in any case fail when facing an invalid signature type, but it can do it miserably (e.g., because `case SignatureType of` does not have a clause for the specified unsupported type).

CDDelta: > We need to ensure compatibility with the existing clients; therefore, if there is no signature_type field we need to consider it an RSA-signed transaction.

Yeah, definitely. Currently, I am doing this by checking whether or not `signature_type` is defined and using `?DEFAULT_KEY_TYPE` when verifying if it isn't. Here's an [example](https://github.com/ArweaveTeam/arweave/pull/309/files#diff-a7235baf70043c993ba2c2a7dcaab49f31756d54297f0fb6703f98d9cf43fcb0R465).

I do this as opposed to just defaulting `signature_type` on txs because `signature_data_segment_v2/1` needs a way to know whether or not `signature_type` was defined in the original tx to generate the right signature data [here](https://github.com/ArweaveTeam/arweave/pull/309/files#diff-a7235baf70043c993ba2c2a7dcaab49f31756d54297f0fb6703f98d9cf43fcb0R219).

As an alternative to this we can maybe include a boolean on the internal tx representation on whether or not `signature_type` was included when deserializing but I don't think this is as elegant.

CDDelta: > Hm, if we detect an invalid signature in json_struct_to_tx already, why pass it on? Obviously, ar_tx:verify must in any case fail when facing an invalid signature type, but it can do it miserably (e.g., because case SignatureType of does not have a clause for the specified unsupported type).

If we do this users won't get an `invalid_signature_type` error from the HTTP API when they specify an invalid `signature_type` right? It isn't a big issue, I was just wondering if we should do this for dev API ergonomics sake.

ldmberman: > I do this as opposed to just defaulting signature_type on txs because signature_data_segment_v2/1 needs a way to know whether or not signature_type was defined in the original tx to generate the right signature data.

There is another alternative - do not include the signature type in the segment of RSA transactions. I like it more because otherwise we get two redundant ways of constructing and validating an RSA transaction; moreover, the "new" format (with the type explicitly specified) cannot be verified by the existing clients. 

> If we do this users won't get an invalid_signature_type error from the HTTP API when they specify an invalid signature_type right

Good point! We could update [the handler](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_http_iface_middleware.erl#L1932) accordingly and [record the error code](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_tx.erl#L312) in the serialization function, WDYT? The error cache is used by the `GET /tx/[id]` handler to display the failure reason for recently failed transactions.

CDDelta: > There is another alternative - do not include the signature type in the segment of RSA transactions. I like it more because otherwise we get two redundant ways of constructing and validating an RSA transaction;

Ahh, I like this method too since it removes the need for all the `undefined` checks.

> We could update the handler accordingly and record the error code in the serialization function, WDYT? The error cache is used by the GET /tx/[id] handler to display the failure reason for recently failed transactions.

Sounds good. Have a look at my recent commit, I think I got what you meant.

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -1929,6 +1929,15 @@ post_tx_parse_id(parse_json, {TXID, Req, Body}) ->
...
 			{error, invalid_json, Req};
+		{error, invalid_signature_type} ->
+            case TXID of
+                not_set ->
+                    noop;
+                _ ->
+                    ar_ignore_registry:remove_temporary(TXID),
+					ar_tx_db:put_error_codes(TXID, [<<"invalid_signature_type">>])
+            end,
+            {error, invalid_signature_type, Req};
```
ldmberman, line 1940: We need to add a clause [here](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_http_iface_middleware.erl#L449) as well, to convert the atom to the response binary.

apps/arweave/src/ar_wallet.erl
```diff
@@ -18,45 +17,86 @@
...
+		{?ECDSA_SIGN_ALG, secp256k1} ->
+			{{_, Priv, Pub}, _} = new(KeyType),
+			<<_:8, PubPoint/binary>> = Pub,
+			PubPointMid = byte_size(PubPoint) div 2,
+			<<X:PubPointMid/binary, Y:PubPointMid/binary>> = PubPoint,
+			Key =
+				ar_serialize:jsonify(
+					{
+						[
+							{kty, <<"EC">>},
```
ldmberman, line 73: Is it some common key storage format for ecdsa keys?

CDDelta: Are you referring to the formatting of X and Y here? It appears to be. The Erlang OTP always generates public keys [uncompressed with an 0x04 prefix](https://bitcoin.stackexchange.com/questions/19666).

Here's an example of a [secp256k1 ECDSA jwk](https://identity.foundation/EcdsaSecp256k1RecoverySignature2020/#privateKeyJwk).

I think it would be good to add some tests loading externally generated keyfiles and signing/verifying with them.

CDDelta: I thought about ECDSA public key compression a bit more and did some research.

It looks like Bitcoin supports both compressed and uncompressed EC public keys in its txs, see [here](https://en.bitcoin.it/wiki/Elliptic_Curve_Digital_Signature_Algorithm) and [here](https://bitcoin.stackexchange.com/questions/5952/why-does-bitcoin-support-both-compressed-and-uncompressed-keys-addresses). It will likely make more sense for us to only support compressed EC public keys in txs to save space and prevent confusion around the same EC wallets having two different addresses depending on client implementation.

Something that we can also take advantage of is the fact that EC public keys can be derived from their signatures. Ethereum appears to do this with their txs [here](https://cryptobook.nakov.com/digital-signatures/ecdsa-sign-verify-messages#ecdsa-public-key-recovery-from-signature). Not too sure on the trade-offs for implementing this since the changes to the node software might be substantial and this optimization is exclusive to ECDSA.

From what I can tell EdDSA signed txs don't require these considerations as its public keys are all already compressed.

I'll look into whether or not the Erlang OTP can verify ECDSA signatures with compressed public keys (it should be able to 🤷).

ldmberman: Interesting, thank you. I wonder how fast the signature verification is compared to RSA, also considering compressed keys.

> Not too sure on the trade-offs for implementing this though since the changes to the node software might be substantial and this optimization is exclusive to ECDSA. It will also mess with the GraphQL API that lets you query for txs with the tx owner.

It does not require a protocol upgrade, we can introduce it later. Strategically though, I think, the more important work is including chain headers into the dataset for preservation and creating bundling services for scaling upload.

apps/arweave/src/ar_wallet.erl
```diff
@@ -111,39 +201,71 @@ verify(Key, Data, Sig) ->
...
 %%%===================================================================
 %%% Tests.
 %%%===================================================================
 
-wallet_sign_verify_test() ->
-	TestData = <<"TEST DATA">>,
-	{Priv, Pub} = new(),
-	Signature = sign(Priv, TestData),
-	true = verify(Pub, TestData, Signature).
+wallet_sign_verify_test_() ->
```
CDDelta, line 237: Should these tests be copied over to the `ar_wallet_tests.erl` file?

CDDelta: Going off the style guide, I have moved these tests.

apps/arweave/src/ar_wallet.erl
CDDelta: I have updated the node to generate and load ECDSA public keys in a compressed form and made it so that it only accepts properly sized public keys for ECDSA and EdDSA.

CDDelta:
@ldmberman are there any tests that test out tx inclusion e2e ie. tx creation, `/tx` endpoint posting, block inclusion? Could you point me to them if there are? I think it would be good to have them updated with the new signature types as well.

ldmberman:
@CDDelta there are lots of e2e tests in the codebase, one thing I would do is make `ar_wallet:new/0` return an ECDSA or EDDSA wallet and see if the tests pass, but not necessarily commit the change. To name a few places, there are complex e2e tests in `ar_multiple_txs_per_wallet_tests.erl` and `ar_data_sync_tests.erl`. It might make sense to update these to use a mix of wallets. Also, perhaps, check `ar_node_tests`, e.g., `wallet_transaction_test_`.

CDDelta:
> @CDDelta there are lots of e2e tests in the codebase, one thing I would do is make `ar_wallet:new/0` return an ECDSA or EDDSA wallet and see if the tests pass, but not necessarily commit the change. To name a few places, there are complex e2e tests in `ar_multiple_txs_per_wallet_tests.erl` and `ar_data_sync_tests.erl`. It might make sense to update these to use a mix of wallets. Also, perhaps, check `ar_node_tests`, e.g., `wallet_transaction_test_`.

`ar_node_tests` was just what I was looking for. I believe everything is in line now. Could you have a look again? :)

DougAnderson444:
Hello Team, wondering what the status of this PR is? Would love to see ed25519 keys used to sign Arweave transactions.

Issue #349, opened 2022-02-20
when updated 2.5.1.0 ubuntu18.0 rocksdb failed to start


fadeAce:
built from source code v2.5.1.0
```
=CRASH REPORT==== 20-Feb-2022::09:10:12.761260 ===
  crasher:
    initial call: supervisor:kernel/1
    pid: <0.685.0>
    registered_name: []
    exception exit: {on_load_function_failed,rocksdb}
      in function  init:run_on_load_handlers/0
      in call from kernel:init/1 (kernel.erl, line 189)
      in call from supervisor:init/1 (supervisor.erl, line 330)
      in call from gen_server:init_it/2 (gen_server.erl, line 423)
      in call from gen_server:init_it/6 (gen_server.erl, line 390)
    ancestors: [kernel_sup,<0.659.0>]
    message_queue_len: 0
    messages: []
    links: [<0.661.0>]
    dictionary: []
    trap_exit: true
    status: running
    heap_size: 376
    stack_size: 28
    reductions: 253
  neighbours:
```

fadeAce:
but works fine swithed to v2.5.0.0

fadeAce:
I've solved this problem related to PR#352

PR #352, opened 2022-02-22
CMake requirements supplement to compile the upgrade "Upgrade erlang-…

fadeAce:
…rocksdb to 1.7.0; RocksDB 6.25.3"

If the CMake version is below v3.10.0, we could still build a rocksDB binary lib but meetting the error at bin/start of arweave issued #349


ldmberman:
@fadeAce thank you for the update!

PR #396, opened 2023-01-18
Changed rsync command for dd in defragmentation

ElFantasma:
This was tested on a test node: The rsync command seems to be a bit slower and if used with the `--sparse` flag the file is not defragmented. See [it's documentation](https://explainshell.com/explain?cmd=rsync+--sparse)

### DIFF
--- a/apps/arweave/src/ar_data_sync.erl
+++ b/apps/arweave/src/ar_data_sync.erl
@@ -437,6 +437,9 @@ handle_cast({join, RecentBI}, State) ->
 		{_, {_H, Offset, _TXRoot}} ->
 			PreviousWeaveSize = element(2, hd(CurrentBI)),
 			{ok, OrphanedDataRoots} = remove_orphaned_data(State, Offset, PreviousWeaveSize),
+			{ok, Config} = application:get_env(arweave, config),
+			[gen_server:cast(list_to_atom("ar_data_sync_" ++ ar_storage_module:id(Module)),
+					{cut, Offset}) || Module <- Config#config.storage_modules],
 			ar_chunk_storage:cut(Offset, StoreID),
 			ar_sync_record:cut(Offset, ?MODULE, StoreID),
 			ar_events:send(data_sync, {cut, Offset}),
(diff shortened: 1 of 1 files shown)


apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -518,7 +518,7 @@ defrag_files([Filepath | Rest]) ->
 	?LOG_DEBUG([{event, defragmenting_file}, {file, Filepath}]),
 	ar:console("Defragmenting ~s...~n", [Filepath]),
 	TmpFilepath = Filepath ++ ".tmp",
-	DefragCmd = io_lib:format("rsync --sparse --quiet ~ts ~ts", [Filepath, TmpFilepath]),
+	DefragCmd = io_lib:format("dd if=~ts of=~ts conv=notrunc bs=1M status=none", [Filepath, TmpFilepath]),
```
vird, line 521: According to this https://superuser.com/questions/1136357/why-didnt-dd-conv-sparse-save-space-as-i-expected
I think conv=sparse is necessary:
* conv=notrunc was needed when you read and write same file. Simple dd will just preserve (in theory) sparsity, but it should be verified
* user can accidently copy data with plain cp. dd with conv=sparse will fix that for user, so less space will be used and altual read will consume less bandwidth, because not reading zeroes.

Also I think at least bs=256k is necessary
* We have chunks 256k length each (with packing all chunks have equal size)
* Actually even lower bs needed in ideal case because all 3.6 TB blocks are shifted
https://arweave.net/block/height/1100753
"strict_data_split_threshold":"30607159107830"
30607159107830%(256*1024)
41206 bytes shift.

So even with bs=256k we still have 1 extra 256k block allocated for each continious region

ElFantasma: Ok, I believe I misunderstood part of the requirement: 
- We just need to copy the file from one place to another to have a better probability that all the chunks are placed in contiguous space in the disk. Is that it? This is regardless the sparsity of the file.

The part I think I misunderstood is that we don't want to un-sparse the file (ie. filling with zeroes the "holes" in it).

If that's the case I think the `rsync` version is OK and you can discard this PR. I can, though, make some performance tests on `rsync` vs. `dd` preserving the sparsity. But that will be a minor improvement.

vird: We have open discussion right now with Lev about do we need sparse files at all in 2.6 for storage modules, but yes our previous points were
* preserve sparsity
* make less HDD read head repositions (good impact on IOPS)
* (was not noted before) less bandwidth possible (do not read zeroes) both on OS side, but also on HDD side (less spins - more time to do something more usable)

PR #397, opened 2023-01-18
Change `current` field in mining performance report to be a time window since last report

AminArria:
The core of the change here is on how the performance report is tracked and what it shows

The tracking has changed to keep track of 4 things
- Timestamp from when the partition "started" to be mined 
- Counter since partition "started"
- Timestamp of last report
- Counter since last report

Given this changes the `avg` displayed is the same as before, but now `current` refers to the values since last time it was reported. If the reports are paused this will cause the first one to come out to be a sort of average for that duration of the pause


ldmberman:
Included the commit, thank you!

PR #399, opened 2023-01-24
Add estimated optimal performance of partition to mining report

AminArria:
Should we use the VDF speed from startup or should is there a metric we can lookup each time?

### DIFF
--- a/apps/arweave/src/ar_mining_server.erl
+++ b/apps/arweave/src/ar_mining_server.erl
@@ -222,7 +222,7 @@ handle_cast(report_performance, #state{ io_threads = IOThreads, session = Sessio
 					{total_avg_hps, TotalAvg * 4}, {total_current_mibps, TotalCurrent},
 					{total_current_hps, TotalCurrent * 4}]),
 			Str = io_lib:format("~nMining performance report:~nTotal avg: ~.2f MiB/s, "
-								" ~.2f h/s; current: ~.2f MiB/s, ~.2f h/s. VDF speed ~.2f~n",
+								" ~.2f h/s; current: ~.2f MiB/s, ~.2f h/s. VDF ~.2f~n/sec",
 					[TotalAvg, TotalAvg * 4, TotalCurrent, TotalCurrent * 4, VdfSpeed]),
 			prometheus_gauge:set(mining_rate, TotalCurrent),
 			IOList2 = [Str | [IOList | ["~n"]]],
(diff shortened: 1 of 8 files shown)


apps/arweave/src/ar_mining_server.erl
```diff
@@ -1079,6 +1084,13 @@ pick_hashing_thread(Threads) ->
 	{{value, Thread}, Threads2} = queue:out(Threads),
 	{Thread, queue:in(Thread, Threads2)}.
 
+optimal_performance(StoreID) ->
+	VdfSpeed = persistent_term:get(vdf_speed_ms) / 1000,
+	case prometheus_gauge:value(v2_index_data_size_by_packing, [StoreID, spora_2_6]) of
+		undefined -> 0;
+		StorageSize -> (100 / VdfSpeed) * (StorageSize / ?PARTITION_SIZE)
```
ldmberman, line 1091: We should replace 100 with 200 here because we've got 2 100 MiB recall ranges and thus in the best case every partition receives 200 MiB worth of chunks per second on average. Also, it is possible to set up a storage module smaller than a partition so makes sense to replace ?PARTITION_SIZE with the storage module size.

ldmberman:
> Should we use the VDF speed from startup or should is there a metric we can lookup each time?

Reusing the value computed on startup is fine (although in practice the speed may fluctuate so updating it after every computation is a bit better). A bigger issue here is we may be connected to a remote VDF server and our speed then won't be relevant. Could you take a look into [the part where we accept remotely computed steps](https://github.com/ArweaveTeam/arweave/blob/testnet/2.6/apps/arweave/src/ar_nonce_limiter.erl#L299) and try to measure the speed there as it goes? I'd say it is also makes sense to measure every step computed by us then too.

apps/arweave/src/ar_mining_server.erl
AminArria: Changed it look for the storage module size

AminArria:
@ldmberman decided to change how the VDF speed was calculated, now it is looking at the frequency of `computed_output` events it gets, which if understood correctly would cover the local and remote cases

ldmberman:
Yes, computing the frequency of receiving `computed_output`s works, thank you!

I added a few small fixes and merged the changes.  Reversed `VdfCount / VdfLapse` and removed the optimum and VDF speed from the output when they are not known (yet).

PR #414, opened 2023-03-20
Add metrics for assessing syncing and packing performance

JamesPiechota:
Metrics are used in this Grafana dashboard: https://arweave.grafana.net/d/cvSjnX-Vk/sync-and-pack?orgId=1&var-node=gateway-3-temp.arweave.net&from=now-3h&to=now

- Top panels track high-level metrics. They chart Packing Rate, Packing Latency, and Chunk Write Throughput against a collection of benchmarks.
- All the bottom (collapsed) panels provide more details that can be useful in investigating performance issues.

This PR also adds a lightweight process sampler. Every 500 ms it wakes up, checks the active processes, and increments a prometheus counter for each one. This provides a (rough) view into what processes might be dominating the CPU.

### DIFF
--- a/apps/arweave/src/ar_packing_server.erl
+++ b/apps/arweave/src/ar_packing_server.erl
@@ -429,11 +429,7 @@ record_buffer_size_metric() ->
 
 record_packing_benchmarks(PackingStateRef, MaxRate, PackingRate, Schedulers) ->
 	prometheus_gauge:set(packing_latency_benchmark,
-		[protocol, pack, spora_2_5], ?PACKING_LATENCY_MS),
-	prometheus_gauge:set(packing_latency_benchmark,
 		[protocol, pack, spora_2_6], ?PACKING_LATENCY_MS),
-	prometheus_gauge:set(packing_latency_benchmark,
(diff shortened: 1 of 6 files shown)


apps/arweave/src/ar_packing_server.erl
```diff
@@ -414,6 +427,85 @@ record_buffer_size_metric() ->
 			ok
 	end.
 
+record_packing_benchmarks(PackingStateRef, MaxRate, PackingRate, Schedulers) ->
+	prometheus_gauge:set(packing_latency_benchmark,
+		[protocol, pack, spora_2_5], ?PACKING_LATENCY_MS),
```
ldmberman, line 432: spora_2_5 packing involves a different [number of RX hashes](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/include/ar_consensus.hrl#L2). The latency does not linearly depend on the number of RX programs so we cannot simply prorate it unfortunately, without making experiments. The spora_2_5 packing is not involved in the new protocol, although some miners are still repacking their legacy storages, so may be we should just exclude spora_2_5 from these metrics.

PR #419, opened 2023-05-11
Improvement/vdf server pull interface

ldmberman:


apps/arweave/src/ar_metrics.erl
```diff
@@ -485,6 +485,13 @@ name_route([]) ->
 	"/";
 name_route([<<"vdf">>]) ->
 	"/vdf";
+name_route([<<"vdf">>, <<"init">>]) ->
```
JamesPiechota, line 488: I'm not finding the /vdf/init endpoint?

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -357,23 +356,30 @@ init([]) ->
...
 	case ar_util:safe_parse_peer(RawPeer) of
 		{ok, Peer} ->
-			resolve_remote_server_raw_peers(RawPeers, [Peer | Peers]);
+			case ets:lookup(ar_nonce_limiter, {vdf_server_raw_peer, RawPeer}) of
+				[] ->
+					ets:insert(ar_nonce_limiter, {{vdf_server_raw_peer, RawPeer}, Peer}),
+					ets:insert(ar_nonce_limiter, {{vdf_server_peer, Peer}});
+				[{_, Peer}] ->
+					ok;
+				[{_, PreviousPeer}] ->
```
JamesPiechota, line 373: Do you know when we'd expect this case to happen? Would it primarily be when a peer uses a domain name rather than IP address, and the IP underlying the domain name has changed?

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -546,7 +554,6 @@ handle_cast({validated_steps, Args}, State) ->
 					Session3#vdf_session.step_number),
 			Sessions2 = gb_sets:add_element({element(2, NextSessionKey),
 					element(1, NextSessionKey)}, Sessions),
-			ar_events:send(nonce_limiter, {validated_output, {SessionKey, Session2}}),
```
JamesPiechota, line 549: Without this, we'll basically wait for the next VDF step that the VDF server computes, at which point it will try to push both the current and previous sessions, right? i.e. we don't need to immediately push the validated session since we can wait ~1 second for the next step to be computed and push the sessions then

apps/arweave/src/ar_nonce_limiter_server_worker.erl
```diff
@@ -40,6 +69,17 @@ handle_cast(resolve_raw_peer, #state{ raw_peer = RawPeer } = State) ->
 	case ar_util:safe_parse_peer(RawPeer) of
 		{ok, Peer} ->
 			ar_util:cast_after(15000, self(), resolve_raw_peer),
+			case ets:lookup(ar_nonce_limiter, {vdf_client_raw_peer, RawPeer}) of
```
JamesPiechota, line 72: any way we can combine this code with the duplicate code in `resolve_remote_server_raw_peers`?

apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -0,0 +1,157 @@
...
+	{ok, Config} = application:get_env(arweave, config),
+	RawPeers = queue:from_list(Config#config.nonce_limiter_server_trusted_peers),
+	{ok, #state{ remote_servers = RawPeers }}.
+
+handle_call(Request, _From, State) ->
+	?LOG_WARNING("event: unhandled_call, request: ~p", [Request]),
+	{reply, ok, State}.
+
+handle_cast(pull, State) ->
+	ar_util:cast_after(800, ?MODULE, pull),
```
JamesPiechota, line 45: I'm not sure how much it matters, but with a fixed polling rate, the client will likely get some VDF steps pretty "late". 
i.e. if the server is generating a new step every second, the client will get:
- step 1, at 1.6 seconds (0.6 seconds late)
- step 2, at 2.4 seconds (0.4 seconds late)
- step 3, at 3.2 seconds (0.2 seconds late)
- step 4, at 4 seconds (ontime)
- step 5 at 5.6 seconds (0.6 seconds late)
- etc...

However since the lateness doesn't compound I'm not sure how material this is for miners

One fix might be to have the server include information in its response about when to next pull (e.g. based on an average of how fast its computing VDF and when it last computed the VDF)

apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -0,0 +1,157 @@
...
+									{noreply, State};
+								_ ->
+									{noreply, State2}
+							end;
+						#nonce_limiter_update_response{ step_number = StepNumber }
+								when StepNumber >= SessionStepNumber ->
+							%% We are ahead of this server; try another one, if there are any.
+							{noreply, State2};
+						_ ->
+							case fetch_and_apply_session(Peer) of
```
JamesPiechota, line 73: Is this the case where we received a partial session, but there's a gap in the step numbers? e.g. the update we received is at step 100, but our last seen step was 90?

apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -0,0 +1,157 @@
...
+					end;
+				not_found ->
+					%% The server might be restarting. Try another one, if there are any.
+					{noreply, State#state{ remote_servers = queue:in(RawPeer, Q2) }};
+				{error, Reason} ->
+					?LOG_WARNING([{event, failed_to_fetch_vdf_update},
+							{peer, ar_util:format_peer(Peer)},
+							{error, io_lib:format("~p", [Reason])}]),
+					%% Try another server, if there are any.
+					{noreply, State#state{ remote_servers = queue:in(RawPeer, Q2) }}
```
JamesPiechota, line 88: In this case we'll wait 0.8 seconds to try the next server, right? (i.e. for the next cast message to be received)

apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -0,0 +1,157 @@
...
+				_ ->
+					error
+			end
+	end.
+
+fetch_and_apply_session_and_previous_session(Peer) ->
+	case ar_http_iface_client:get_vdf_session(Peer) of
+		{ok, #nonce_limiter_update{ session = #vdf_session{
+				prev_session_key = PrevSessionKey } } = Update} ->
+			case ar_http_iface_client:get_previous_vdf_session(Peer) of
```
JamesPiechota, line 124: I *think* there's a risk of race condition here. If the server switches to a new session between calls to `get_vdf_session` and `get_previous_vdf_session` we may end up getting the same session twice.

apps/arweave/src/ar_nonce_limiter.erl
ldmberman: Yep, that's what I wanted to do for a start.

apps/arweave/src/ar_nonce_limiter_client.erl
JamesPiechota: ah yea, I see that now. I think there's still a chance of missing a session. Like in this case:

1. GET /vdf
- session_key = B
- prev_session_key = A
2. Session changes to C
3. GET /vdf/previous_session
- session_key = B
- prev_session_key = A
4. Hit the retry case since we were expecting session_key = A
5. GET /vdf
 - session_key = C
 - prev_session_key = B
 6. Get /vdf/previous_session
 - session_key = B
 - prev_session_key = A
 7,  ok

In this case we end up with C and B, and never process A (I think). Is that okay?

JamesPiechota: Gotcha! Thanks!

apps/arweave/src/ar_nonce_limiter.erl
JamesPiechota: Got it!

apps/arweave/src/ar_nonce_limiter_server_worker.erl
JamesPiechota: looks good!

JamesPiechota (approved):

PR #422, opened 2023-05-27
2.6.8 Hard Fork

JamesPiechota:
Old pricing logic:
- 2.6HF -> 2.6TransitionStart: pre-2.6 pricing
- 2.6TransitionStart to 2.6TransitionEnd: interpolate between pre-2.6 pricing and 2.6 pricing

New pricing logic:
- 2.6HF -> 2.6TransitionStart: pre-2.6 pricing
- 2.6TransitionStart to 2.6.8HF: interpolate between pre-2.6 pricing and 2.6 pricing
- 2.6.8HF to 2.6.8TransitionStart: fixed fee (0.858 AR)
- 2.6.8TransitionStart to 2.6.8TransitionEnd: interpolate between 0.858 AR fixed fee and 2.6 pricing


apps/arweave/src/ar_fork.erl
```diff
@@ -94,3 +94,11 @@ height_2_6() ->
 			1132210. % Targeting 2023-03-06 14:00 UTC
 	-endif.
 -endif.
+
+-ifdef(FORKS_RESET).
+height_2_6_8() ->
+	0.
+-else.
+height_2_6_8() ->
+	1189720. % Targeting 2023-05-30 16:00 UTC
```
ldmberman, line 103: I estimate ~1189560 computing the average block time over the 10000 blocks window (128 seconds).

apps/arweave/src/ar_tx.erl
```diff
@@ -559,32 +559,54 @@ get_tx_fee(Args) ->
...
+	StaticFee = (?STATIC_2_6_8_FEE_WINSTON div ?GiB) * (DataSize + ?TX_SIZE_BASE),
+
+	case Height of
+		H when H >= V2PricingHeight ->
+			get_tx_fee2(Args3);
+		H when H >= TransitionStart_2_6_8 ->
+			%% For 2.6.8 the fee at the start of the transition period is
+			%% the 2.6 transition fee in effect when the 2.6.8 fork occured.
+			get_transition_tx_fee(
+				StaticFee, %% StartFee
```
ldmberman, line 587: This way we do not incur an extra fee for new accounts. I suggest we use `get_tx_fee_pre_fork_2_6(Args2)` here and in the case below (or `get_tx_fee_pre_fork_2_6({DataSize, Rate, Height, Accounts, Addr, ?FORK_2_6_TIMESTAMP})` if we want it static).

apps/arweave/src/ar_tx.erl
```diff
@@ -559,32 +559,54 @@ get_tx_fee(Args) ->
...
+		H when H >= Fork_2_6_8 ->
+			StaticFee;
+		H when H >= TransitionStart_2_6 ->
+			get_transition_tx_fee(
+				get_tx_fee_pre_fork_2_6(Args2), %% StartFee
+				get_tx_fee2(Args3), %% EndFee
+				TransitionStart_2_6, 
+				TransitionEnd_2_6,
+				Height);
+		_ ->
```
ldmberman, line 601: We know `H >= TransitionStart_2_6` happened so we do not need this clause.

JamesPiechota: This last clause is for H < TransitionStart_2_6 - do we still ned this clause in order to allow validation of old blocks?

apps/arweave/src/ar_tx.erl
JamesPiechota: For the StaticFee we were aiming to pick a fee that is roughly what it would be when the 2.6.8 HardFork goes out - so would I use: `get_tx_fee_pre_fork_2_6({DataSize, Rate, Height, Accounts, Addr, ?FORK_2_6_8_TIMESTAMP}`

And add
```
-define(FORK_2_6_8_TIMESTAMP, 1685462400) %% timestamp for 2023-05-30 16:00 UTC
```

apps/arweave/src/ar_tx.erl
JamesPiechota: https://github.com/ArweaveTeam/arweave/pull/422/commits/15e10af482b0e9e58e95c43981386e182e7f2fdb

apps/arweave/src/ar_tx.erl
JamesPiechota: https://github.com/ArweaveTeam/arweave/pull/422/commits/15e10af482b0e9e58e95c43981386e182e7f2fdb

apps/arweave/src/ar_fork.erl
JamesPiechota: https://github.com/ArweaveTeam/arweave/pull/422/commits/15e10af482b0e9e58e95c43981386e182e7f2fdb

apps/arweave/src/ar_tx.erl
JamesPiechota: Oh wait and if we want it static, we also need to freeze the Height to ar_fork:height_2_6_8()

apps/arweave/src/ar_tx.erl
JamesPiechota: Removed that clause: https://github.com/ArweaveTeam/arweave/pull/422/commits/3d33e3fe750714dac7373eddf9ad76ae5bd8f9cc

apps/arweave/src/ar_tx.erl
JamesPiechota: Removed static fee and instead use `get_tx_fee_pre_fork_2_6(Args2)` - i.e. pricing as if the transition never started

https://github.com/ArweaveTeam/arweave/pull/422/commits/3d33e3fe750714dac7373eddf9ad76ae5bd8f9cc

PR #429, opened 2023-06-07
2 OOM fixes, some syncing optimizations, and a metrics cull

JamesPiechota:
This PR includes 3 main changes:
1. Fix two out-of-memory issues
2. Optimizations to the data sync process
3. Remove about 2,000 unused prometheus metrics from /metrics

## Out of memory issues
The two memory issues were both due to a backlog in the erlang process mailbox:
1. **VDF server out of memory**: when a configured `vdf_client_peer` was down or slow to respond (i.e. it took more than ~1 second to respond) the VDF server mailbox would gradually fill up with a queue of VDF updates.
    1. The fix is to have the VDF server catch up whenever it pushes a new update to a client. The VDF server will skip over updates that may have come in while it was waiting on a peer to respond to a previous push.
2. **Data sync out of memory**: when syncing local data from neighboring partitions the sync process could be overwhelmed with thousands of `read_range` messages causing a spike in memory footprint. This would only happen when a partition contained thousands of small chunks. In one particular example partition 31 contained thousands of 4-byte chunks. When trying to copy these 4-byte chunks to the neighboring partition 30, each chunk was packed up to the standard 256KiB size and embedded in an erlang message. These thousands of 256KiB messages would often spike the memory footprint above the node capacity.
    1. The fix is throttle how many chunks are copied at once in order to give the `ar_data_sync_storage_module` process time to process its queue.

## Optimization to the data sync process
1. Better load balancing across sync_jobs
2. A more balanced algorithm for determining when to cut a given peer's queue. Previously the algorithm was based on a hard-coded value - occasionally this would allow the available sync queue space to be dominated by a single slow peer, starving out the node's ability to sync from other (faster) peers. The new algorithm weights the available queue space based on a peers recent performance.
3. Similarly we now use a more dynamic metric for determining the optimal number of concurrent requests a node is allowed to make to a given peer. Previously this metric was tied to the ration of `packing_rate` to `sync_jobs`, which could yield variable sync performance depending on how those values were set.

## Remove prometheus metrics
The following metrics have been removed:
- erlang_vm_msacc_XXX
- erlang_vm_allocators
- erlang_vm_dist_XXX

Together they accounted for 2,000-3,000 individual metrics (roughy half the metrics advertised by the /metrics endpoint)


config/sys.config
```diff
@@ -11,6 +11,9 @@
 			{request_labels, [http_method, route, reason, status_class]},
 			{error_labels, [http_method, route, reason, error]},
 			{labels_module, ar_prometheus_cowboy_labels}
-		]}
+		]},
+		{vm_system_info_collector_metrics, []},
```
JamesPiechota, line 15: Omit the following metrics:
- `erlang_vm_msacc_XXX`
- `erlang_vm_allocators`
- `erlang_vm_dist_XXX`

Together they represent about 2,000 individual metrics that aren't widely used (if used at all)

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -816,9 +816,9 @@ repack(Start, End, NextCursor, RightBound, RequiredPacking, StoreID) ->
 												{register_packing_ref, Ref, PaddedOffset}),
 										ar_util:cast_after(300000, Server,
 												{expire_repack_request, Ref}),
-										ar_events:send(chunk, {repack_request, Ref,
+										ar_packing_server:request_repack(Ref,
```
JamesPiechota, line 819: Since only the `ar_packing_server` handles the `chunk, repack_request` and `chunk, unpack_request` events, we'll send the message straight to the `ar_packing_server` rather than broadcasting it to all `chunk` subscribers. This cuts down on the number of messages that can queue up in different processes' mailboxes (and occasionally lead to OOM exceptions)

apps/arweave/src/ar_data_sync_worker.erl
```diff
@@ -82,7 +89,9 @@ read_range({_Start, _End, _OriginStoreID, TargetStoreID, _SkipSmall} = Args) ->
 		false ->
 			case ar_data_sync:is_disk_space_sufficient(TargetStoreID) of
 				true ->
-					read_range2(Args);
+					?LOG_DEBUG([{event, read_range},
+						{size, (_End - _Start) / (1024*1024)}, {args, Args}]),
+					read_range2(?READ_RANGE_MESSAGES_PER_BATCH, Args);
```
JamesPiechota, line 94: Limit the number of `read_range` messages that will be sent at once. This avoids an OOM exception that can occur when handling ranges that contains thousands of small chunks (e.g. 4-byte chunks that each get packed up to 256KiB)

apps/arweave/src/ar_data_sync_worker_master.erl
```diff
@@ -49,28 +59,31 @@ is_syncing_enabled() ->
 %% @doc Returns true if we can accept new tasks. Will always return false if syncing is
 %% disabled (i.e. sync_jobs = 0).
 ready_for_work() ->
-	{ok, Config} = application:get_env(arweave, config),
-	TotalTaskCount = get_task_count(scheduled) + get_task_count(queued),
-	TotalTaskCount < (Config#config.sync_jobs * 50).
+	try
+		gen_server:call(?MODULE, ready_for_work, 1000)
```
JamesPiechota, line 63: Previously when this was implement via an ETS query it could lead to race conditions where the `ar_data_sync_worker_master` was in the middle of processing a message and therefore unable to update the ETS data - during that tim the `ar_data_sync` processes would send thousands of extra tasks. Now that the `read_for_work` is itself implemented as a message it synchronizes the request and forces the `ar_data_sync` processes to wait for the `ar_data_sync_worker_master` to finish processing one batch of tasks before sending more.

apps/arweave/src/ar_data_sync_worker_master.erl
```diff
@@ -167,84 +183,118 @@ dequeue_main_task(State) ->
...
+			{PeerTasks, State}
 	end.
 
+%% @doc the maximum number of tasks we can have in process - including stasks queued here as well
+%% as those scheduled on ar_data_sync_workers.
+max_tasks() ->
+	{ok, Config} = application:get_env(arweave, config),
+	Config#config.sync_jobs * 50.
+
+%% @doc The maximum number of tasks we can have queued for a given peer.
```
JamesPiechota, line 202: These changes update the algorithm for deciding when to cut a peer's queue of pending requests.

apps/arweave/src/ar_data_sync_worker_master.erl
```diff
@@ -254,99 +304,84 @@ schedule_read_range(Args, State) ->
...
+set_peer_tasks(PeerTasks, State) ->
+	State#state{ peer_tasks =
+		maps:put(PeerTasks#peer_tasks.peer, PeerTasks, State#state.peer_tasks)
+	}.
+
+get_worker(State) ->
+	AverageLoad = State#state.scheduled_task_count / State#state.worker_count,
+	cycle_workers(AverageLoad, State).
+
+cycle_workers(AverageLoad, #state{ workers = Workers, worker_loads = WorkerLoads} = State) ->
```
JamesPiechota, line 375: Better load balancing when sending work to `ar_data_sync_workers`. Keep cycling through the queue until we find a worker with less than average load.

apps/arweave/src/ar_nonce_limiter_server_worker.erl
```diff
@@ -56,12 +56,18 @@ handle_info({event, nonce_limiter, _Event}, #state{ peer = undefined } = State)
...
 	#state{ peer = Peer, pause_until = Timestamp } = State,
 	{SessionKey, Session, PrevSessionKey, PrevSession, Output, PartitionUpperBound} = Args,
+	CurrentStepNumber = ar_nonce_limiter:get_current_step_number(),
 	case os:system_time(second) < Timestamp of
 		true ->
 			{noreply, State};
 		false ->
-			{noreply, push_update(SessionKey, Session, PrevSessionKey, PrevSession, Output,
-					PartitionUpperBound, Peer, State)}
+			case Session#vdf_session.step_number < CurrentStepNumber of
```
JamesPiechota, line 64: Skip over any VDF updates that have occurred while the `server_worker` was waiting on the `vdf_client` to process that last update.

apps/arweave/src/ar_sup.erl
```diff
@@ -98,5 +98,9 @@ init([]) ->
 		?CHILD(ar_p3, worker),
 		?CHILD(ar_p3_db, worker)
 	],
-
-	{ok, {{one_for_one, 5, 10}, Children}}.
+	{ok, Config} = application:get_env(arweave, config),
```
JamesPiechota, line 101: Only run the `ar_process_sampler` when `debug=true`

apps/arweave/src/ar_chunk_storage.erl
ldmberman: I would go even further and reply via direct messages too (reply PID may be passed to the request_* functions)

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -52,6 +53,10 @@ put(Offset, Chunk, StoreID) ->
 			Reply
 	end.
 
+store_packed_chunk(Offset, ChunkArgs, StoreID) ->
```
JamesPiechota, line 56: The changes to this file have the the ar_data_sync process cast directly to the ar_chunk_storage process covering the same storage module. Previously the ar_chunk_storage listed to the `chunk, packed` ar_event.

Question: is this the correct handling? Previusly *every* ar_chunk_storage process would get notified for every packed chunk - no only the one ar_chunk_storage that covers the chunk's storage_module will be notified.

apps/arweave/src/ar_chunk_storage.erl
JamesPiechota: Done! https://github.com/ArweaveTeam/arweave/pull/429/commits/6937c0c5540055bfe5ba92eccdfdba2d14e16101

I have one question called out here: https://github.com/ArweaveTeam/arweave/pull/429/files#r1230125676

ldmberman (approved):

PR #433, opened 2023-06-14
Feature/merkle tree rebase support

ldmberman:


apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -670,8 +651,20 @@ pre_validate_poa(B, PrevB, PartitionUpperBound, H0, H1, Peer, Timestamp, ReadBod
 	RecallByte1 = RecallRange1Start + B#block.nonce * ?DATA_CHUNK_SIZE,
 	{BlockStart1, BlockEnd1, TXRoot1} = ar_block_index:get_block_bounds(RecallByte1),
 	BlockSize1 = BlockEnd1 - BlockStart1,
+	RebaseThreshold =
+		case B#block.height > ar_fork:height_2_7() of
```
JamesPiechota, line 655: `>=`? not sure it matters, but that seems to be what we use elsewhere

JamesPiechota: oh nevermind! I see.

apps/arweave/src/ar_data_sync.erl
```diff
@@ -1510,12 +1518,17 @@ validate_served_chunk(Args) ->
 		false ->
 			case ar_block_index:get_block_bounds(Offset - 1) of
 				{BlockStart, BlockEnd, TXRoot} ->
-					{_Strict, ValidateDataPathFun} =
-						case BlockStart >= ?STRICT_DATA_SPLIT_THRESHOLD of
+					ValidateDataPathFun =
```
JamesPiechota, line 1521: can this chunk be refactored with the other `ValidateDataPathFun` selection code?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -2429,30 +2442,34 @@ validate_proof2(Args) ->
...
-			R ->
-				{true, R}
+	Base = ar_merkle:validate_path_strict_borders(DataRoot, Offset, TXSize, DataPath),
+	Strict = ar_merkle:validate_path_strict_data_split(DataRoot, Offset, TXSize, DataPath),
+	Rebase = ar_merkle:validate_path2(DataRoot, Offset, TXSize, DataPath),
+	Result =
+		case {Base, Strict, Rebase} of
+			{false, false, false} ->
+				false;
+			{_, {_, _, _} = R, _} ->
```
JamesPiechota, line 2452: What do you think about renaming R to better indicate which of the validation results it comes from? e.g.
```
		case {Base, Strict, Rebase} of
			{false, false, false} ->
				false;
			{_, {_, _, _} = StrictResult, _} ->
				StrictResult;
			{_, _, {_, _, _} = RebaseResult} ->
				RebaseResult;
			{{_, _, _} = BaseResult, _, _} ->
				BaseResult
		end,
```

apps/arweave/src/ar_data_sync.erl
```diff
@@ -2931,20 +2949,50 @@ deregister_currently_processed_disk_pool_key(Key, State) ->
...
+
 process_disk_pool_chunk_offset(Iterator, TXRoot, TXPath, AbsoluteOffset, MayConclude, Args,
 		State) ->
 	#sync_data_state{ disk_pool_threshold = DiskPoolThreshold } = State,
-	{Offset, _, _, DataRoot, DataPathHash, _, _, PassedStrictValidation} = Args,
-	case AbsoluteOffset =< ?STRICT_DATA_SPLIT_THRESHOLD orelse PassedStrictValidation of
+	{Offset, _, _, DataRoot, DataPathHash, _, _,
+			PassedBase, PassedStrictValidation, PassedRebaseValidation} = Args,
+	PassedValidation =
+		case {AbsoluteOffset >= get_merkle_rebase_threshold(),
```
JamesPiechota, line 2966: Is there some way to document this logic? I have sample comment below... but I've been tryin to think of how to change the code so that a comment wouldn't be needed... but coming up empty. Here's a potentially comment:

```
%% Validation passes under the following conditions:
%% 1. If offset is past the rebase_threshold it must pass rebase validation
%% 2. If offset is between the data_split_threshold and rebase_threshold it must pass all 3 validations (base, strict, and rrebase) 
%% 3. If offset is before the data_split_threshold it must either pass all 3 validations, or *just* pass the base validation (and fail the other 2)
```

JamesPiechota: One minor change (not sure if it makes it easier to follow though) - basically if all checks pass, then the overall validation passes regardless of what the Offset is. (I think this is logically consistent with the current code)

```
	PassedValidation =
		case {AbsoluteOffset >= get_merkle_rebase_threshold(),
				AbsoluteOffset >= ?STRICT_DATA_SPLIT_THRESHOLD,
				PassedBase, PassedStrictValidation, PassedRebaseValidation} of
			{_, _, true, true, true} ->
				true;
                         {true, true, _, _, true} ->
				true;
			{false, false, true, false, false} ->
				true;
			_ ->
				false
		end,
```

apps/arweave/src/ar_data_sync.erl
```diff
@@ -3008,7 +3056,8 @@ process_disk_pool_matured_chunk_offset(Iterator, TXRoot, TXPath, AbsoluteOffset,
 	%% The other modules will either sync the chunk themselves or copy it over from the
 	%% other module the next time the node is restarted.
 	#sync_data_state{ chunk_data_db = ChunkDataDB, store_id = DefaultStoreID } = State,
-	{Offset, _, ChunkSize, DataRoot, DataPathHash, ChunkDataKey, Key, _StrictSplit} = Args,
+	{Offset, _, ChunkSize, DataRoot, DataPathHash, ChunkDataKey, Key, _BaseSplit, _StrictSplit,
```
JamesPiechota, line 3059: what do you think about renaming `_BaseSplit, _StrictSplit, _RebaseSupportSplit` to `_PassedBase, _PassedStrictValidation, _PassedRebaseValidation`? I think they are the same values, right? It might make it easier to track the values across function calls / messages.

apps/arweave/src/ar_merkle.erl
```diff
@@ -190,6 +196,112 @@ validate_path_strict_data_split(_ID, _Dest, _LeftBound, _RightBound, _Path, _Pat
 		_DataSize) ->
 	false.
 
+%% @doc Validate the given merkle path where any subtrees may have 0-based offset.
+validate_path2(ID, Dest, RightBound, _Path) when RightBound =< 0 ->
```
JamesPiechota, line 200: Instead of `validate_path2` what do you think about `validate_rebased_path` or `validate_path_with_offset` or something simiar

apps/arweave/src/ar_merkle.erl
```diff
@@ -190,6 +196,112 @@ validate_path_strict_data_split(_ID, _Dest, _LeftBound, _RightBound, _Path, _Pat
...
+		<< Data:?HASH_SIZE/binary, EndOffset:(?NOTE_SIZE*8) >>, PathSize, DataSize,
+		LeftBoundShift) ->
+	case EndOffset - LeftBound > ?DATA_CHUNK_SIZE
+			orelse RightBound - LeftBound > ?DATA_CHUNK_SIZE of
+		true ->
+			io:format("Bounds: E=~B L=~B R=~B~n", [EndOffset, LeftBound, RightBound]), % TODO
+			false;
+		false ->
+			ChunkSize = EndOffset - LeftBound,
+			ValidateSplit =
```
JamesPiechota, line 221: Does this value store whether or not the split is valid? What do you think about naming it `IsSplitValid` or something like that?

apps/arweave/src/ar_merkle.erl
```diff
@@ -190,6 +196,112 @@ validate_path_strict_data_split(_ID, _Dest, _LeftBound, _RightBound, _Path, _Pat
...
+								andalso DataSize - LeftBound > (?DATA_CHUNK_SIZE)
+								andalso DataSize - LeftBound < 2 * (?DATA_CHUNK_SIZE)
+				end,
+			case ValidateSplit of
+				false ->
+					io:format("Split: E=~B L=~B R=~B C=~B~n",
+							[EndOffset, LeftBound, RightBound, ChunkSize]), % TODO
+					false;
+				true ->
+					case hash([hash(Data), hash(note_to_binary(EndOffset))]) of
```
JamesPiechota, line 242: is it possible to refactor this with the non-rebased check [here](https://github.com/ArweaveTeam/arweave/pull/433/files#diff-da5e1b73bd29394af0e871a3dc4b2299cafb3a0cc8aeb5b072f362667865918aL53-L56)?

apps/arweave/src/ar_merkle.erl
```diff
@@ -190,6 +196,112 @@ validate_path_strict_data_split(_ID, _Dest, _LeftBound, _RightBound, _Path, _Pat
...
+validate_path2(ID, Dest, RightBound, Path) when Dest >= RightBound ->
+	validate_path2(ID, RightBound - 1, RightBound, Path);
+validate_path2(ID, Dest, RightBound, Path) when Dest < 0 ->
+	validate_path2(ID, 0, RightBound, Path);
+validate_path2(ID, Dest, RightBound, Path) ->
+	validate_path2(ID, Dest, 0, RightBound, Path, byte_size(Path), RightBound, 0).
+
+validate_path2(ID, _Dest, LeftBound, RightBound,
+		<< Data:?HASH_SIZE/binary, EndOffset:(?NOTE_SIZE*8) >>, PathSize, DataSize,
+		LeftBoundShift) ->
```
JamesPiechota, line 213: Is `LeftBoundShift` the 0-based offset?

apps/arweave/src/ar_merkle.erl
JamesPiechota: Actually... is there some way to combine both the `validate_path` and `validate_path2` functions into a single function? E.g. by setting the rebase to 0 or something when calling `validate_path`? Or maybe it's LeftBoundShift that needs to be set when validating a non-rebased tree?

apps/arweave/src/ar_merkle.erl
```diff
@@ -190,6 +196,112 @@ validate_path_strict_data_split(_ID, _Dest, _LeftBound, _RightBound, _Path, _Pat
...
+					[Dest, LeftBound, RightBound, LeftBoundShift, NextLeftBoundShift]), % TODO
+			validate_path2(Path, Dest2, NextLeftBound, NextRightBound, Rest,
+					PathSize, DataSize, NextLeftBoundShift);
+		_ ->
+			false
+	end;
+validate_path2(ID, Dest, LeftBound, RightBound,
+		<< L:?HASH_SIZE/binary, R:?HASH_SIZE/binary, Note:(?NOTE_SIZE*8), Rest/binary >>,
+		PathSize, DataSize, LeftBoundShift) ->
+	case hash([hash(L), hash(R), hash(note_to_binary(Note))]) of
```
JamesPiechota, line 283: can this be combined with the `validate_path` code [here](https://github.com/ArweaveTeam/arweave/pull/433/files#diff-da5e1b73bd29394af0e871a3dc4b2299cafb3a0cc8aeb5b072f362667865918aL59-L71)?

apps/arweave/src/ar_poa.erl
```diff
@@ -26,11 +26,16 @@ validate(Args) ->
 				RecallOffset - BlockStartOffset
 		end,
 	ValidateDataPathFun =
-		case BlockStartOffset >= StrictDataSplitThreshold of
+		case BlockStartOffset >= MerkleRebaseSupportThreshold of
```
JamesPiechota, line 29: What do you think about moving this code block and the one in [ar_data_sync.erl](https://github.com/ArweaveTeam/arweave/pull/433/files#diff-2934ddad2ed46ac194a6be7abde4be6d440bc7565e1128bba234fad9fd2bd5d2R929) into ar_merkle.erl?

ldmberman: Makes total sense, although I would move it to ar_poa and keep ar_merkle the place to define basic merkle tree functionality without the specifics of the protocol like rule-upgrading thresholds.

apps/arweave/src/ar_merkle.erl
```diff
@@ -208,81 +304,96 @@ extract_root(_) ->
...
-		Elements
-	).
+generate_tree([Element | Elements], Stack, Tree) when is_list(Element) ->
+	{SubRoot, SubTree} = generate_tree(Element),
+	SubTree2 = [mark_rebased(Node, SubRoot) || Node <- SubTree],
+	SubRootN = get(SubRoot, SubTree2),
+	generate_tree(Elements, queue:in(SubRootN, Stack), Tree ++ SubTree2);
+generate_tree([Element | Elements], Stack, Tree) ->
+	Leaf = generate_leaf(Element),
+	generate_tree(Elements, queue:in(Leaf, Stack), [Leaf | Tree]);
```
JamesPiechota, line 314: Why is SubTree2 added to the end of the Tree, but Leaf is added to the beginning?

apps/arweave/src/ar_merkle.erl
```diff
@@ -208,81 +304,96 @@ extract_root(_) ->
...
+	{N, Shift2} = generate_node(L, R, Shift),
+	[N | generate_row(Rest, Shift2)].
+
+generate_node(Left, empty, Shift) ->
+	{Left, Shift};
+generate_node(L, R, Shift) ->
+	LMax = L#node.max,
+	LMax2 = case L#node.is_rebased of true -> Shift + LMax; _ -> LMax end,
+	RMax = R#node.max,
+	RMax2 = case R#node.is_rebased of true -> LMax2 + RMax; _ -> RMax end,
```
JamesPiechota, line 360: Why is RMax shifted by LMax2 and not Shift?

apps/arweave/src/ar_merkle.erl
ldmberman: The tree structure here is not sensitive to the ordering so it makes a tiny bit more sense to prepend (as it is faster). SubTree is a list too though so it does not matter.

apps/arweave/src/ar_merkle.erl
ldmberman: Consider a tree:
```
                          Root
                  /                  \
        Branch1                      Branch2 <- we are here
       /         \                  /        \
Leaf1            Leaf2          Leaf3     Leaf4

Shift      = 2*ChunkSize
LMax       = 3*ChunkSize, when not rebased 
           =   ChunkSize, when rebased
LMax2      = 3*ChunkSize (always)
RMax       = 4*ChunkSize, when not rebased
           =   ChunkSize, when rebased
RMax2      = 4*ChunkSize (always)
```
The tags passed to ar_merkle:generate_tree when no tree is rebased:

```
[{Leaf1Hash, ChunkSize}, {Leaf2Hash, 2*ChunkSize}, {Leaf3Hash, 3*ChunkSize}, {Leaf4Hash, 4*ChunkSize}]
```

when the subtree with Leaf3 is rebased:

```
[{Leaf1Hash, ChunkSize}, {Leaf2Hash, 2*ChunkSize}, [{Leaf3Hash, ChunkSize}], {Leaf4Hash, 4*ChunkSize}]
```

when the subtree with Leaf4 is rebased:

```
[{Leaf1Hash, ChunkSize}, {Leaf2Hash, 2*ChunkSize}, {Leaf3Hash, 3*ChunkSize}, [{Leaf4Hash, ChunkSize}]]
```

JamesPiechota: Ah, okay. I'm still trying to keep the data formats straight. It sounds like LMax2 and RMax2 will always be the non-rebased values? Or maybe that's not the right term - they'll always be the "global" values? I.e. the values after all 0-offset subtrees have been rebased against a shared Root?

ldmberman: Yes!

JamesPiechota: Great - thanks!

JamesPiechota (approved):

PR #438, opened 2023-06-28
Fix a bug where VDF client domains are not re-resolved

ldmberman:


apps/arweave/src/ar_nonce_limiter_server_worker.erl
```diff
@@ -39,11 +43,12 @@ handle_call(Request, _From, State) ->
 handle_cast(resolve_raw_peer, #state{ raw_peer = RawPeer } = State) ->
 	case ar_peers:resolve_and_cache_peer(RawPeer, vdf_client_peer) of
 		{ok, Peer} ->
+			ar_util:cast_after(?RESOLVE_DOMAIN_NAME_FREQUENCY_MS, self(), resolve_raw_peer),
 			{noreply, State#state{ peer = Peer }};
 		{error, Reason} ->
 			?LOG_WARNING([{event, failed_to_resolve_vdf_client_peer},
 					{reason, io_lib:format("~p", [Reason])}]),
-			ar_util:cast_after(30000, self(), resolve_raw_peer),
+			ar_util:cast_after(10000, self(), resolve_raw_peer),
```
JamesPiechota, line 51: Should this also be `?RESOLVE_DOMAIN_NAME_FREQUENCY_MS`?

JamesPiechota (approved):

apps/arweave/src/ar_nonce_limiter_server_worker.erl
ldmberman: I think, it's likely to be a temporary error here (e.g., I had one recently where it failed but the next attempt succeeded) so I reduced it a bit, but it does not matter that much

JamesPiechota: Gotcha. Makes sense!

PR #443, opened 2023-07-07
Jp/unify ranking 20230707

JamesPiechota:
This change does 3 things:
1. Consolidates all peer ranking in `ar_peers`
2. Introduces an additional ranking algorithm (`lifetime` vs. `current`)
3. Fixes some small inefficiencies in the ranking and syncing code

### Consolidates all peer ranking in `ar_peers`

Earlier this year there was a bug in the data syncing logic which caused nodes to stall when syncing chunks from peers. The fix was to update `ar_data_sync_worker_master` to be more active in managing the `ar_data_sync_worker`s and which peers they were syncing from. To do this we introduced several operational metrics (e.g. average latency per request and number of concurrent requests per peer)  the `ar_data_sync_worker_master`. In effect this created a second "ranking" algorithm separate from the ranking maintained by `ar_peers`.

While this solved the syncing performance issue, it created a dichotomy that introduced inefficiencies and made some future feature work (e.g. syncing from P3 servers) difficult. This PR moves all of the operational metrics from `ar_data_sync_worker_master` into `ar_peers` so that they can be incorporated into the global ranking. This addresses the earlier inefficiencies (e.g. the peer list used by `ar_data_sync` to select peers for syncing now incorporates data about which peers have provided good throughput in the past) and will allow us to integrate P3 servers more efficiently.

###  Introduces an additional ranking algorithm (`lifetime` vs. `current`)

1. Old ranking is labeled `lifetime` since it considers all of the data received from a peer for all time
2. New ranking is labeled `current`. It uses an Exponential Moving Average to track each of the metrics used to calculate a peer's rating. 
3. Consider concurrent requests when calculating a peer's `current` rating (i.e. its recent estimated throughput)
4. Maintain an average failure rate rather than strict fail count. Failure rate impacts the estimated throughput. Scale both ratings by this value. Once a peer drops below an 80% success rating, it is (temporarily) removed from the peer set. It can be re-added the next time the node rebuilds its network map (roughly every 4 minutes)
5. For now the only place that the `current` rating is used is when collecting peers for data discovery (`ar_data_discovery:collect_peers()`), as that operation is very sensitive to available bandwidth and slow peers can have a material impact on how long it takes to sync the weave.

### Fixes some small inefficiencies in the ranking and syncing code
1. Replace `ar_events` broadcasts with direct casts to `ar_peers` when updating ratings. This reduces some of the interprocess message traffic
2. `ar_data_sync_worker_master` will continue to schedule tasks until all workers are utilized
3. `ar_data_sync_worker_master` rebalances peers every minute instead of on every completed request


apps/arweave/src/ar_blacklist_middleware.erl
```diff
@@ -45,7 +45,6 @@ start() ->
 ban_peer(Peer, TTLSeconds) ->
 	Key = {ban, peer_to_ip_addr(Peer)},
 	Expires = os:system_time(seconds) + TTLSeconds,
-	ar_events:send(peer, {banned, Peer}),
```
ldmberman, line 48: Did you have an intention to replace it with a direct ar_peers cast?

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -309,10 +313,14 @@ name_route([<<"hash_list">>, _From, _To]) ->
 name_route([<<"hash_list2">>, _From, _To]) ->
 	"/hash_list2/{from}/{to}";
 
+name_route([<<"block">>]) ->
```
ldmberman, line 316: One-element paths are handled on L226.

apps/arweave/src/ar_blacklist_middleware.erl
JamesPiechota: ar_peers now calls `ar_blacklist_middleware:ban_peer` directly: https://github.com/ArweaveTeam/arweave/pull/443/files#diff-de37132d068e49cb11eb29b5c880b8ef2454d93f79686171b6cff9392824a2c5R353

We only ban peers in response to certain types of invalid blocks. Previously ar_block_pre_validator would both call `ban_peer` and send a `block, rejected` event. Now it just sends the event, which ar_peers handles and decides whether or not to ban the peer.

apps/arweave/src/ar_http_iface_server.erl
JamesPiechota: ah, thanks! Reverted: https://github.com/ArweaveTeam/arweave/pull/443/commits/b32f7c4ee9411df9b93358ca80eb6a2dbfb0a99c

apps/arweave/src/ar_data_sync.erl
```diff
@@ -944,7 +944,6 @@ handle_cast({store_fetched_chunk, Peer, Time, TransferSize, Byte, Proof} = Cast,
 							ar_util:cast_after(1000, self(), Cast),
 							{noreply, State};
 						false ->
-							ar_events:send(peer, {served_chunk, Peer, Time, TransferSize}),
```
ldmberman, line 947: Where do we record chunk download statistics now?

JamesPiechota: Previously the only action we did in response to a `peer, served_chunk` event was to update the peer rating in `ar_peers` (that was the main reason I removed the event).

Now the rating is updated [here](https://github.com/ArweaveTeam/arweave/pull/443/files/b32f7c4ee9411df9b93358ca80eb6a2dbfb0a99c#diff-f2d3cef82c02c5e4782cac6bd44f38f590355bb8573c4c980198ee030ccc43bcR320)
 in `ar_data_sync_worker_master` whenever an `ar_data_sync_worker` completes syncing a chunk.
 
 Other statistics are also update in `ar_data_sync_worker_master` whenever an `ar_data_sync_worker` [completes syncing a range](https://github.com/ArweaveTeam/arweave/pull/443/files/b32f7c4ee9411df9b93358ca80eb6a2dbfb0a99c#diff-f2d3cef82c02c5e4782cac6bd44f38f590355bb8573c4c980198ee030ccc43bcR109)

apps/arweave/src/ar_peers.erl
```diff
@@ -552,59 +641,134 @@ rank_peers(ScoredPeers) ->
...
+		undefined -> TotalBytes;
+		_ -> TotalBytes + DataSize
+	end,
+	%% AverageBytes is the average number of bytes transferred during the AverageLatency time
+	%% period. In order to approximate the impact of multiple concurrent requests we multiply
+	%% DataSize by the Concurrency value. We do this *only* when updating the AverageBytes
+	%% value so that it doesn't distort the TotalBytes.
+	AverageBytes2 = case DataSize of
+		undefined -> AverageBytes;
+		_ -> calculate_ema(AverageBytes, (DataSize * Concurrency), ?THROUGHPUT_ALPHA)
```
ldmberman, line 693: What happens if we do not use `Concurrency` here? Do I understand it correctly that when we download from 2 peers at the same time, the contribution to both ratings is doubled?

apps/arweave/src/ar_peers.erl
```diff
@@ -34,23 +37,47 @@
 %% The maximum number of peers to return from get_peers/0.
 -define(MAX_PEERS, 1000).
 
-%% The number of failed requests in a row we tolerate before dropping the peer.
--define(TOLERATE_FAILURE_COUNT, 20).
+%% Minimum average_success we'll tolerate before dropping a peer.
+-define(MINIMUM_SUCCESS, 0.8).
+-define(THROUGHPUT_ALPHA, 0.05).
```
ldmberman, line 42: I suggest to comment it is used in the transfer throughput EMA, which is approximately the average over 1 / 0.05 = 20 data points.

apps/arweave/src/ar_peers.erl
```diff
@@ -34,23 +37,47 @@
...
+%% set so that roughly 20 consecutive failures will drop the average_success below 0.5
+-define(SUCCESS_ALPHA, 0.035).
```
ldmberman, line 44: It should be approximately the average over 1 / 0.035 = 28.5 data points so considering we had successes before a series of consecutive failures, 28.5 / 2 failures are sufficient to bring the average down to 50%. Does it sound correct?

apps/arweave/src/ar_peers.erl
```diff
@@ -552,59 +641,134 @@ rank_peers(ScoredPeers) ->
...
+	end,
+	%% AverageBytes is the average number of bytes transferred during the AverageLatency time
+	%% period. In order to approximate the impact of multiple concurrent requests we multiply
+	%% DataSize by the Concurrency value. We do this *only* when updating the AverageBytes
+	%% value so that it doesn't distort the TotalBytes.
+	AverageBytes2 = case DataSize of
+		undefined -> AverageBytes;
+		_ -> calculate_ema(AverageBytes, (DataSize * Concurrency), ?THROUGHPUT_ALPHA)
+	end,
+	TotalLatency2 = case LatencyMilliseconds of
```
ldmberman, line 695: What do you think about computing the ratio (throughput) EMA directly? I am thinking about the sensitivity to occasional throughput spikes, like a Gbit/s transfer after the usual 100 Mbit/s - we should probably value consistent 200 MBit/s transfers higher than 100 MBit/s with very occasional mixes of 1 GBit/s transfers. The harmonic mean will be least sensitive to such spikes (`1 / (Alpha / NewThroughput + (1 - Alpa) / OldThroughput)` should be the harmonic EMA alternative). However, this might be very theoretical as in practice the performance of a single peer won't vary much and "occasional" spikes still mean paying for bandwidth for a potential attackerso we might just note it down for later.

apps/arweave/src/ar_peers.erl
JamesPiechota: The concurrency value allows us to estimate the true throughput from that peer better. Since we are only considering one data point at a time if we ignore concurrency we underestimate the throughput. 

For example if we make two concurrent requests to a peer:
A. 1kb with a latency if 1second
B. 2kb with a latency if 1second

The true throughput is 3kb per second. 

However if ignore concurrency we'll average each data point and end up with 1.5kb per second. 

When using a concurrency value of 2 we compute the normal data average (1.5kb) and then multiply by 2 to approximate the fact that there are 2 streams of data concurrently and. And we end up with the "correct" 3kb per second

It is just an approximation - it won't always yield exactly the true throughput. But it is much closer than just considering each data point in isolation. This is primarily important for data syncing where we regularly have many concurrent requests going

apps/arweave/src/ar_peers.erl
JamesPiechota: The math is weird because of how ema is calculated - I'll try to find the equation again. But when I ran it 0.035 ended up being 20 failures to bring value from 1 to 0.5

apps/arweave/src/ar_peers.erl
JamesPiechota: You mean instead of averaging data and latency separately, we would calculate the throughput for a single data point and then average that?

I'm fine with that change although I'm not entirely sure what the impact will be (I'll have to think through it some more). Would we do the same for the lifetime rating? 

I guess that one would become: (sum of all throughout) / TotalTransfers

apps/arweave/src/ar_peers.erl
JamesPiechota: It turns out its roughly: the last 90 data points contribute 99% of the throughput average. I added a comment explaining the logic here: https://github.com/ArweaveTeam/arweave/pull/443/commits/fbfa3d54ecfc630d5b4534c17cf35d7e3e911aad

JamesPiechota: Perhaps that's too many? Would we prefer to only consider 20 data points?

JamesPiechota: Oops, in my previous explanation I forgot one part of the equation (a datapoint is immediately scaled by Alpha when it is added to the average). With the correct interpretation its roughly the last 31 points that contribute 99% of the average. Updated commit with the comment: https://github.com/ArweaveTeam/arweave/pull/443/commits/8937471296bfe63478c22df5baf505576ffe0654

ldmberman (approved):

PR #447, opened 2023-07-17
Add support for running a local weave

ldmberman:


README.md
```diff
@@ -80,6 +80,37 @@ You can join the public testnet now:
...
+The given address (if none is specified, one will be generated for you) will be assigned
+`1000_000_000_000` AR in the new weave.
+
+The network name will be `arweave.localnet`. You can not start the same node again with the
+init option unless you clean the data directory - you need to either restart with the
+`start_from_block_index` option or specify a peer from the same Arweave network via
+`peer <peer>`. Note that the state is only persisted every 50 blocks so if you
+restart the node without peers via `start_from_block_index` before reaching the height 50,
+it will go back to the genesis block.
+
```
JamesPiechota, line 113:
```suggestion
When POST'ing transactions to your localnet make sure to include the `X-Network: arweave.localnet` header. If the header is omitted, the mainnet network will be assumed and the request will fail.
```

JamesPiechota, line 113:
```suggestion
As with mainnet peers, each peer must be run in its own physical or virtual environment (e.g. on its own machine or in its own container or virtual machine). If you try to run two nodes within the same environment you will get an error like `Protocol 'inet_tcp': the name arweave@127.0.0.1 seems to be in use by another Erlang node`
```

JamesPiechota (approved):
LGTM! This is great!

I tested it out on two machines and everything works as expected. I was able to spin up two peers, mine blocks, and post data transactions. I added a couple notes to the Readme to try to head off some potentially common mistakes.

PR #449, opened 2023-08-04
feat(nix): add maxVDFValidationThreadCount option

hlolli:
An option some are interested to configure via nix configuration.

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -1,12 +1,23 @@
 { config, lib, pkgs, ... }:
 
 let
-  inherit (lib) literalExpression mkEnableOption mkIf mkOption types;
+  inherit (lib) literalExpression filterAttrs isAttrs isList mkEnableOption mkIf mkOption types;
   cfg = config.services.arweave;
   defaultUser = "arweave";
   arweavePkg = pkgs.callPackage ./arweave.nix { inherit pkgs; };
+  recursiveFilterNulls = set:
+    let
+      isNotNull = value: value != null;
+      filterNulls = attrs: filterAttrs (name: value: isNotNull value) attrs;
+    in
+      lib.mapAttrs (name: value:
+        if isAttrs value then recursiveFilterNulls (filterNulls value)
+        else if isList value then value
+        else if isNotNull value then value
+        else {})
+      set;
   generatedConfigFile =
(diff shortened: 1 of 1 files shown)


nix/module.nix
```diff
@@ -306,6 +307,12 @@ in
       description = "As semaphore, the max amount of parallel get sync record requests to perform.";
     };
 
+    maxVDFValidationThreadCount = mkOption {
+      type = types.int;
+      default = 10;
```
JamesPiechota, line 312: FWIW the default [used in ar.erl](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/include/ar_config.hrl#L88-L89) if the `max_vdf_validation_thread_count` is not provided is `schedulers_online / 2` - which translates roughly to half the available cores.

JamesPiechota: Not sure if this is relevant to your use case, though!

hlolli: That's true, we may want to add that as a comment for users to consider. But given the dynamic nature of "schedulers_online" it's not really easy or even possible to infer that from a static configuration. I'll add it as a comment!

hlolli: Well I just realized, that I can make it default to null to not interfere with the default dynamic behavior. Only when user passes a value it will be set, otherwise not. Then it's the same behavior.

JamesPiechota: Agree - comment works for me.

Unless there's an easy way to not include `max_vdf_validation_thread_count` in the generated config.json if it isn't provided? That would have the node use the `ar.erl` hardcoded default.

hlolli: yes it was easy to allow for nullability (kudos to chatgpt for the null filter function), I'm testing this now on a node. I can give heads up when I've confirmed that this works as intended.

JamesPiechota: hahah, I used chatgpt to explain the filter function to me! Oh man, I'm seeing a future where ChatGPT trolls us all by returning some gibberish and then pretending like the gibberish does exactly what it's supposed to.

hlolli: turned out chatgpt did some blunders, the function is fixed now as well as the nullability, which required the type `type = with types; nullOr int;`

PR #456, opened 2023-08-18
Ensure the genesis txs are copied into the data_dir on node launch.

### LINKED ISSUE #455: Genesis block TXs are not served
It appears that TXs in the genesis block are no longer being served. While not a huge issue, this is visible when trying to index the chain from genesis (e.g. using an ar.io node).

JamesPiechota:
resolves #455

### DIFF
--- a/apps/arweave/src/ar_weave.erl
+++ b/apps/arweave/src/ar_weave.erl
@@ -105,19 +105,24 @@ create_genesis_tx(Key) ->
 	ar_tx:sign_v1(UnsignedTX, Key).
 
 add_mainnet_v1_genesis_txs() ->
-	{ok, Files} = file:list_dir("data/genesis_txs"),
-	{ok, Config} = application:get_env(arweave, config),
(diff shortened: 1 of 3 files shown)


apps/arweave/src/ar_weave.erl
```diff
@@ -105,19 +105,24 @@ create_genesis_tx(Key) ->
...
+					SourcePath = "data/genesis_txs/" ++ F,
+					TargetPath = Config#config.data_dir ++ "/" ++ ?TX_DIR ++ "/" ++ F,
+					file:copy(SourcePath, TargetPath),
+					[ar_util:decode(hd(string:split(F, ".")))|Acc]
+				end,
+				[],
+				Files
+			);
+		false ->
+			[]  % Skip if the directory does not exist
```
vird, line 123: Maybe log warning here?

JamesPiechota: done!

PR #472, opened 2023-09-29
fix(http): provide a 404 not_found response for missing tx-data

hlolli:


apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -1555,7 +1555,7 @@ serve_format_2_html_data(Req, ContentType, TX) ->
 				{error, tx_data_too_big} ->
 					{400, #{}, jiffy:encode(#{ error => tx_data_too_big }), Req};
 				{error, not_found} ->
-					{200, #{ <<"content-type">> => ContentType }, <<>>, Req};
+					{404, #{}, <<"Not Found">>, Req};
```
JamesPiechota, line 1558: For consistency with the other `GET /tx` might be worth updating this line as well:
https://github.com/ArweaveTeam/arweave/pull/472/files#diff-0b7084531612365a22d52032a422eb664adc91b34cefb5f8d9499584072aa489R1526

Also, it looks like the `GET /tx/ID/data.html` endpoint uses this to report a 404:
https://github.com/ArweaveTeam/arweave/pull/472/files#diff-0b7084531612365a22d52032a422eb664adc91b34cefb5f8d9499584072aa489R334

`{404, #{}, sendfile("data/not_found.html"), Req};`

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -1523,7 +1523,7 @@ serve_tx_data(Req, #tx{ format = 2, id = ID } = TX) ->
 				{error, tx_data_too_big} ->
 					{400, #{}, jiffy:encode(#{ error => tx_data_too_big }), Req};
 				{error, not_found} ->
-					{200, #{}, <<>>, Req};
+					{404, #{}, sendfile("data/not_found.html"), Req};
```
JamesPiechota, line 1526: not sure what this one should return for data... I think it normally responds with JSON, so an HTML response probably isnt' right... maybe just `<<>>`

Oh looks like ` <<"Not Found.">>` is that that code path returns elsewhere: https://github.com/ArweaveTeam/arweave/pull/472/files#diff-0b7084531612365a22d52032a422eb664adc91b34cefb5f8d9499584072aa489R1247

JamesPiechota (approved):

PR #475, opened 2023-10-09
ci(tests): self-hosted runner

hlolli:


.github/workflows/test.yml
```diff
@@ -1,24 +1,113 @@
...
+            ar_retarget,
+            ar_serialize,
+            ar_storage,
+            ar_sync_buckets,
+            ar_tx,
+            ar_tx_db,
+            ar_unbalanced_merkle,
+            ar_util,
+            ar_weave,
+            # note we don't explicitly include *_tests.erl here
```
JamesPiechota, line 73: I think we may need some of the *_tests.erl file - not all of them share the same name as an existing module

JamesPiechota: and we have smoe that match an existing module, but that module itself doesn't have tests so isn't in the `core_test_mod` list

JamesPiechota: actually it looks like that's a lot of *_tests.erl files...

JamesPiechota: basically I think we need to include all the *_tests.erl file *except* for those that match the list above.

JamesPiechota (approved):

PR #478, opened 2023-10-16
feat: remove unused dependencies sqlite and erlang-graphql

hlolli:


nix/arweave.nix
```diff
@@ -451,16 +419,10 @@ let
         mkdir -p apps/arweave/lib/RandomX
         cp -rf ${randomx}/* apps/arweave/lib/RandomX
         cp -rf ${jiffy}/lib/erlang/lib/* apps/jiffy
-        cp -rf ${graphql}/lib/erlang/lib/* apps/graphql
       '';
 
       postPatch = ''
-        sed -i -e 's|-arch x86_64|-arch ${pkgs.stdenv.targetPlatform.linuxArch}|g' \
-          apps/arweave/c_src/Makefile \
```
JamesPiechota, line 459: do we not need `apps/arweave/c_src/Makefile` any more?

hlolli: good catch, we do, I should have just removed apps/ar_sqlite3/c_src/Makefile

hlolli: fixed

PR #481, opened 2023-10-21
Feature/coordinated mining v6

JamesPiechota:
This PR has 3 main parts:

1. Implement **Coordinated Mining** This PR builds on the work started in the `feature/coordinated-mining_v4` branch. The `v4` branch (and the versions before it), build out the coordinated mining system which I'll describe below. 
2. Address some reported **mining performance issues**
3. **Code refactoring and testing**

# Coordinated Mining

The coordinated mining system built out in earlier work was implemented by @vird and @ldmberman 

## Overview

When coordinated mining is configured multiple nodes can cooperate to find mining solutions for the same mining address without the risk of slashing which can occur when two nodes publish blocks with the same mining address. This allows multiple nodes which each store a disjoint subset of the weave to reap the hashrate benefits of more two-chunk solutions.

## Basic System

In a coordinated mining cluster there are 2 roles:
1. Exit Node
2. Miners

All nodes in the cluster share the same mining address. Each Miner generates H1 hashes for the partitions they store. Occasionally they will need an H2 for a packed partition they don't store. In this case, they can find another Miner in the coordinated mining cluster who does store the required partition packed with the required address, send them the H1 and ask them to calculate the H2. When a valid solution is found (either one- or two-chunk) the solution is sent to the Exit Node. Since the Exit Node is the only node in the coordinated mining cluster which publishes blocks, there's no risk of slashing. This point can be further enforced by ensuring only the Exit Node stores the mining address private key (and therefore only the Exit Node can sign blocks for that mining address)

Every node in the coordinated mining cluster is free to peer with any other nodes on the network as normal.

## Single-Miner One Chunk Flow

```mermaid
sequenceDiagram
participant Miner1
participant Miner2
participant Exit Node

Note over Miner1: Compute H0
Note over Miner1: Compute H1 > Difficulty
Miner1->>Exit Node: POST /coordinated_mining/publish
Note over Exit Node: Validate solution and gossip block
```

Note: The single-miner two chunk flow (where Miner1 stores both the H1 and H2 partitions) is very similar

## Coordinated Two Chunk Flow

```mermaid
sequenceDiagram
participant Miner1
participant Miner2
participant Exit Node

Note over Miner1: Compute H0
Note over Miner1: Compute H1 < Difficulty
Miner1->>Miner2: POST /coordinated_mining/h1
Note over Miner2: Compute H2 > Difficulty
Miner2->>Miner1: POST /coordinated_mining/h2
Miner1->>Exit Node: POST /coordinated_mining/publish
Note over Exit Node: Validate solution and gossip block
```

## Additional Details

- Miners use the new GET /coordinated_mining/partition_table to build and maintain a mapping of coordinated mining peers to the partitions they store. Each miner uses this mapping when deciding who (if anyone) to ask about an H2 solution
  - This feature was added for convenience as it enables you to update the partitions on one miner without needing to modify the other miners in the cluster. You can stop one of your miners, add more HDDs, adjust your start config with the additional partitions, and restart. The other miners in the cluster will use the `GET /coordinated_mining/partition_table` endpoint to update their internal mapping.
- While a multi-miner solution is being generated, the mining session is ignored. Before sharing a block the Exit Node will validate that the solution is still valid.
- The coordinated mining cluster is configured via these new configuration options:
  - `cm_exit_peer`: all miners should share the same `cm_exit_peer`, and this should be left unset on the Exit Node itself
  - `cm_peers`: all peers in a coordinated mining cluster (excluding the current miner, and the exit node)
  - `cm_poll_interval`: how often the partition table is rebuilt
  - `cm_stat_interval`: how often the coordinated mining performance data is generated
- All nodes in a coordinated mining cluster are not required to share their `h0` with each other. In this case they won't unlock additional two-chunk solutions, but they will prevent slashing for 1-chunk solutions if the miners in the cluster share the same mining address. 
- All nodes in a coordinated mining cluster are not required to share the same mining address. It is common to fully pack one full replica with one mining address first before starting a second replica with a second mining address. However, you have the flexibility to do as you wish. For instance if you chose to start packing both replicas at the same time there may be a period of time where nodes in the coordinated mining cluster do not all share the same mining addresses.

# Mining performance improvements

## Don't tie `miner_server_chunk_cache_size_limit` to free memory

Previously if the `miner_server_chunk_cache_size_limit` wasn't set by the user it would be automatically calculated by `ar_mining_server.erl`. The logic used tied the limit to `free_memory` - which could vary over time and cause mining performance to degrade. This PR ties the value to `total_memory` which is a consistent value.

# Code Refactoring

## Introduce #mining_candidate and #mining_solution records

The main intent of these [two records](https://github.com/ArweaveTeam/arweave/pull/448/files#diff-48fdb1b3bf04f2846c0faf9c02e76fe3a3510663e27f40b70ec5c818c09d9d1fR6) was to encapsulate the long argument tuples previously used in `ar_mining_server.erl` and avoid some of the hairier "argument ordering" bugs. 

## Move the IO threads into ar_mining_io.erl

With the extra logic needed for coordinated mining, `ar_mining_server.erl` was getting pretty complex. Removing the IO thread management helped.

## Unify the mining flow

The basic mining flow is: `compute h0` -> `read chunk 1` -> `compute h1` -> `read chunk 2` -> `compute h2` -> `validate solution` -> `publish solution`. Parts of the flow are parallelized, and the exit conditions vary - but the same basic flow exists for both coordinated and non-coordinated miners and both one- and two-chunk solutions.

Where possible, this PR attempts to reduce code duplication in the flow to increase the amount of shared code for the main scenarios:
- single-miner one-chunk solution
- single-miner two-chunk solution
- single-coordinated-miner one-chunk solution
- single-coordinated-miner two-chunk solution
- double-coordinated-miner two-chunk solution

## Tighten up the `chunk_cache_size`

The `chunk_cache_size` is used to track how much memory is used caching chunks during the mining process. The logic to update the size is complex, and gets even moreso with coordinated mining. Although this value has a strong impact on  mining performance (mining is paused while the cache is full to avoid trigging an out of memory exception), it is reset whenever a new block is found or applied. So a small discrepancy in the cache size tracking is unlikely to have a noticeable impact on performance.

However since it *does* have an impact, and since the coordinated mining logic increases the likelihood of a discrepancy. This PR introduces several tests to validate that the cache size is tracked correctly - and has several bug fixes for those edge cases where it was not.

# Testing

- Introduce 3 new nodes that get launched by the test framework to enable a full coordinated mining cluster
- Migrate all the naming from slave/master to main/peer1/peer2/peer3/peer4 to better capture that the nodes aren't configured in a master/slave relationship


apps/arweave/src/ar_coordination.erl
```diff
@@ -379,7 +399,7 @@ refresh([], State) ->
 	State;
 refresh([Peer | Peers], State) ->
 	NewState = case ar_http_iface_client:get_cm_partition_table(Peer) of
-		{ok, PartitionList} ->
+		{ok, PartitionList} when length(PartitionList) > 0 ->
```
vird, line 402: In this case user will see there is no connection to that node
Original code will make sure that user sees that there is connection to node, but no partitions there

JamesPiechota: Got it. Reason I added that clause was with the new code the node only returns partitions if it is configured to mine - so it may have storage_modules defined, but if mining is disabled, or if the node recently restarted and is still initializing, it will see no partitions.

I guess that' sfine though... if mining is disabled then that node shouldn't receive H1s anyways since it won't be able to compute the H2. And if it's in the middle of initializing, should be fine too since it will be polled again in 60s at which point it will have partitions.

Ok, I'll remove the `when` clause!

apps/arweave/src/ar_data_sync.erl
```diff
@@ -389,6 +389,10 @@ is_chunk_cache_full() ->
 			not_initialized
 	end.
 
+-ifdef(DEBUG).
+is_disk_space_sufficient(_StoreID) ->
+	true.
+-else.
```
hlolli, line 395: we run some nodes in debug mode, I wonder if this will cause the disk to fill up?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -544,7 +549,7 @@ init("default" = StoreID) ->
 				Free = proplists:get_value(free_memory, memsup:get_system_memory_data(),
 						2000000000),
 				Limit2 = min(1000, erlang:ceil(Free * 0.9 / 3 / 262144)),
-				Limit3 = Limit2 - Limit2 rem 100 + 100,
+				Limit3 = ar_util:ceil_int(Limit2, 100),
```
hlolli, line 552: is this making 1.1 -> 2 (ceil) or could math_min work better here?

JamesPiechota: This makes 110 -> 200 - i.e. it rounds up to the nearest multiple of 100

hlolli: ahhh

apps/arweave/src/ar_data_sync.erl
hlolli: ok, bad naming, but makes sense then

hlolli (approved):

PR #485, opened 2023-11-01
feat(ar_disksup): handle input/output errors better when detected

hlolli:
Usually this means the hardware connection failed or the disk needs to be checked with fsck. It's safer that the application crashes because writing to the disk will either fail or make the situation worse.

### DIFF
--- a/apps/arweave/src/ar_disksup.erl
+++ b/apps/arweave/src/ar_disksup.erl
@@ -220,12 +220,28 @@ disk_free_cmd({unix, darwin}, Df, DataDirPath, Port) ->
 disk_free_cmd({unix, _}, Df, DataDirPath, Port) ->
 	 my_cmd(Df ++ " -Pa -B1 " ++ DataDirPath ++ "/", Port).
 
+%% check for hardware errors in df output
+check_for_hardware_error(DfOutput, ThrowOnError) ->
+    case lists:member("Input/output error", DfOutput) of
+        true ->
+            ar:console("~nERROR: one or more of your disks are in corrupt/failing state.~n~p~n", [DfOutput]),
+            case ThrowOnError of
+                true ->
+                    erlang:error({input_output_error_detected, DfOutput});
+                _ ->
+                    true
+            end;
+        false ->
+            false
+    end.
+
 %% doc: iterates trough storage modules
 broadcast_disk_free({unix, _} = Os, Port) ->
 	Df = find_cmd("df"),
 	[DataDirPathData | StorageModulePaths] = get_storage_modules_paths(),
 	{DataDirID, DataDirPath} = DataDirPathData,
 	DataDirDfResult = disk_free_cmd(Os, Df, DataDirPath, Port),
+	check_for_hardware_error(DataDirDfResult, true),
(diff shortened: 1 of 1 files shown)


apps/arweave/src/ar_disksup.erl
```diff
@@ -220,12 +220,28 @@ disk_free_cmd({unix, darwin}, Df, DataDirPath, Port) ->
 disk_free_cmd({unix, _}, Df, DataDirPath, Port) ->
 	 my_cmd(Df ++ " -Pa -B1 " ++ DataDirPath ++ "/", Port).
 
+%% check for hardware errors in df output
+check_for_hardware_error(DfOutput, ThrowOnError) ->
+    case lists:member("Input/output error", DfOutput) of
+        true ->
+            ar:console("~nERROR: one or more of your disks are in corrupt/failing state.~n~p~n", [DfOutput]),
+            if
```
JamesPiechota, line 228: Because of how weirdly/confusingly erlang handles if statements (i.e `true` means `false` 🙃 ) The codebase usually uses case statement for everything. e.g.

```
case ThrowOnError of
    true ->
        erlang:error({input_output_error_detected, DfOutput});
    _ ->
        false
end
```

hlolli: fixed, true means, 1. it matches the error string we are on the lookout for and 2. true means HasError

hlolli: ah wait your comment is regarding the use of if and not case?

hlolli: it's now all case

apps/arweave/src/ar_disksup.erl
```diff
@@ -235,14 +251,21 @@ broadcast_disk_free({unix, _} = Os, Port) ->
...
 	HandleSmPath = fun({StoreID, StorageModulePath}) ->
-		Result = disk_free_cmd(Os, Df, StorageModulePath, Port),
-		[StorageModuleFs, Bytes, Percentage] = parse_df_2(Result),
-		IsDataDirDrive = string:equal(DataDirFs, StorageModuleFs),
-		ar_events:send(disksup, {
-			remaining_disk_space,
-			StoreID, IsDataDirDrive, Percentage, Bytes
-		})
-	end,
+  	Result = disk_free_cmd(Os, Df, StorageModulePath, Port),
```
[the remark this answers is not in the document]
hlolli: fixed

apps/arweave/src/ar_disksup.erl
```diff
@@ -236,13 +252,20 @@ broadcast_disk_free({unix, _} = Os, Port) ->
...
-		IsDataDirDrive = string:equal(DataDirFs, StorageModuleFs),
-		ar_events:send(disksup, {
-			remaining_disk_space,
-			StoreID, IsDataDirDrive, Percentage, Bytes
-		})
-	end,
+		HasDiskError = check_for_hardware_error(Result, false),
+		case HasDiskError of
+			true ->
+						ar:console("~nERROR: storage module ~p is offline.~n", [StorageModulePath]),
```
JamesPiechota, line 258: haha and now this one looks like it's crazy indented

JamesPiechota (approved):

PR #487, opened 2023-11-02
feat(nix): add cookie and improve logging experience

hlolli:


apps/arweave/src/ar.erl
```diff
@@ -541,6 +541,14 @@ start(Port) when is_integer(Port) ->
 	start(#config{ port = Port });
 start(Config) ->
 	%% Start the logging system.
+	case os:getenv("TERM") of
+		"dumb" ->
+			% Set logger to output all levels of logs to the console
+			% when running in a dumb terminal.
+			logger:add_handler(console, logger_std_h, #{level => all});
```
JamesPiechota, line 548: Do you know if this behaves the same as the debug logger when log volume gets too high? When log volume gets high the debug logger will start dropping logs. I think that is probably the behavior we want - when `debug` is on there can be a *lot* of logs.  I bring it up because if the logger doesn't drop logs, I think it could be possible to crash a node by overwhelming the logger

hlolli: answer given in pm, resolving :)

PR #498, opened 2023-11-17
Don't include full #vdf_session records in ar_nonce_limiter events

JamesPiechota:
Limit the data that gets sent with a `nonce_limiter` `computed_output` event. Only `{SessionKey, StepNumber, Output, UpperBound}` are sent now and no full Sesssions are sent. If a subscriber needs the full session data it can query it via the `SessionKey`. This can greatly reduce the memory footprint of a node during certain edge case situations.


apps/arweave/src/ar_mining_io.erl
```diff
@@ -54,7 +54,7 @@ init([]) ->
 					?LOG_DEBUG([{event, start_io_threads},
 							{bucket_size, BucketSize},
 							{bucket, Bucket},
-							{mining_address, ar_util:encode(Addr)}]),
+							{mining_address, ar_util:safe_encode(Addr)}]),
```
ldmberman, line 57: What is the case when we cannot encode the address here?

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -1189,17 +1187,13 @@ maybe_set_vdf_metrics(SessionKey, CurrentSessionKey, Session) ->
...
-trigger_computed_outputs(SessionKey, Session, PrevSessionKey, PrevSession, UpperBound,
-		[Step | Steps]) ->
-	ar_events:send(nonce_limiter, {computed_output, {SessionKey, Session, PrevSessionKey,
-			PrevSession, Step, UpperBound}}),
+send_events_for_external_update(SessionKey, Session, UpperBound, [Step | Steps]) ->
 	#vdf_session{ step_number = StepNumber, steps = [_ | PrevSteps] } = Session,
-	Session2 = Session#vdf_session{ step_number = StepNumber - 1, steps = PrevSteps },
-	trigger_computed_outputs(SessionKey, Session2, PrevSessionKey, PrevSession, UpperBound,
-			Steps).
+	ar_events:send(nonce_limiter, {computed_output, {SessionKey, StepNumber, Step, UpperBound}}),
```
ldmberman, line 1194: We do not seem to need the manipulations with Session#vdf_session.steps anymore, we only need to track the StepNumber, which might make sense to pass as another function argument.

apps/arweave/src/ar_nonce_limiter_server_worker.erl
```diff
@@ -93,88 +92,86 @@ terminate(_Reason, _State) ->
...
-				{true, true, true, false} ->
-					%% Client requested missing steps
-					push_session(SessionKey, Session, Peer, Format),
+push_update(SessionKey, StepNumber, Output,
+		PartitionUpperBound, Peer, Format, State) ->
+	Session = ar_nonce_limiter:get_session(SessionKey),
+	Update = ar_nonce_limiter_server:make_partial_nonce_limiter_update(
+		SessionKey, Session, StepNumber, Output, PartitionUpperBound),
+	case Update of
+		not_found -> ok;
```
ldmberman, line 101: We need to return `State` here.

JamesPiechota: Thanks!

apps/arweave/src/ar_nonce_limiter.erl
JamesPiechota: Ooh, good call. I love it! More code deletion!

JamesPiechota: Fixed: https://github.com/ArweaveTeam/arweave/pull/498/commits/9818fdfdf7e4a1bb2c87b83cd1f86d82bb1e58d5

apps/arweave/src/ar_nonce_limiter_server_worker.erl
JamesPiechota: Fixed: https://github.com/ArweaveTeam/arweave/pull/498/commits/9818fdfdf7e4a1bb2c87b83cd1f86d82bb1e58d5

ldmberman (approved):

PR #502, opened 2023-11-22
ci: add elvis linter

hlolli:


apps/arweave/src/ar.erl
```diff
@@ -328,16 +328,16 @@ parse_cli_args(["local_peer", Peer | Rest], C = #config{ local_peers = Peers })
...
-		C = #config{ transaction_blacklist_urls = URLs} ) ->
+		C = #config{transaction_blacklist_urls = URLs}) ->
 	parse_cli_args(Rest, C#config{ transaction_blacklist_urls = [URL | URLs] });
 parse_cli_args(["transaction_whitelist", File | Rest],
-		C = #config{ transaction_whitelist_files = Files } ) ->
+		C = #config{transaction_whitelist_files = Files}) ->
 	parse_cli_args(Rest, C#config{ transaction_whitelist_files = [File | Files] });
 parse_cli_args(["transaction_whitelist_url", URL | Rest],
-		C = #config{ transaction_whitelist_urls = URLs} ) ->
+		C = #config{transaction_whitelist_urls = URLs}) ->
```
JamesPiechota, line 340: Why do we remove spaces here, but not [here](https://github.com/ArweaveTeam/arweave/pull/502/files#diff-f148c1ce17ebe7e20d6eed136cc4f0c0e517b978a199351bb7e50aca826a8de4R452)

hlolli: that is a good question, I believe the complaint was just about the parentheses and I did more than was needed to satisfy this lint rule

apps/arweave/src/ar_arql_db.erl
```diff
@@ -247,7 +247,9 @@ handle_call({select_txs_by, Opts}, _, #{ conn := Conn } = State) ->
 	{Time, Reply} = timer:tc(fun() ->
```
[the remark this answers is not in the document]
hlolli: I have this PR still open that removes this file entierly
https://github.com/ArweaveTeam/arweave/pull/478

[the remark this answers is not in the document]
hlolli: but you're right about that this isn't as it was, I'll adjust

apps/arweave/src/ar_block.erl
```diff
@@ -484,8 +482,7 @@ generate_block_data_segment_base(B) ->
 							integer_to_binary(ScheduledRateDividend),
 							integer_to_binary(ScheduledRateDivisor),
 							integer_to_binary(B#block.packing_2_5_threshold),
-							integer_to_binary(B#block.strict_data_split_threshold)
-							| Props
+							integer_to_binary(B#block.strict_data_split_threshold) | Props
```
JamesPiechota, line 485: Is this an elvis change? Personally I prefer to keep newlines consistent within an operator. For this operator we've made the choice to put each list element on its own line, so I would prefer to also put the `| Props` on its own line to be consistent. I've been bitten before by inconsistent new lines where only one element is put on the same line as the others and it gets missed when ssomeone is scannign the code because they expect all elements to be on new lines.

apps/arweave/src/ar_block_cache.erl
```diff
@@ -690,13 +688,13 @@ test_block_cache() ->
 	add(bcache_test, B2),
 	mark_tip(bcache_test, block_id(B2)),
 	add(bcache_test, B2_2 = on_top(random_block(1), B2)),
-	?assertMatch({B1_2, [B1], {{not_validated, ExpectedStatus}, _Timestamp}},
+	?assertMatch({B1_2, [B1], {{not_validated, ExpectedStatus}, _}},
```
JamesPiechota, line 691: ditto above. I don't think we should be removing context information from Erlang code and should leave it up to the engineer(s) maintaining that section of code what is most intelligible.

apps/arweave/src/ar_block.erl
hlolli: the default rule says that there can't be space on the left side of an operator, this has the effect that we have

```erlang
ValA +
ValB +
ValC.
```

and not

```erlang
ValA
+ ValB
+ ValC
```

personally I'm not trained to read code like thie

```js
{
  a: 1
  ,b: 2
  ,c: 3
 }
 ```

File: apps/arweave/src/ar_data_sync_worker.erl [hunk 5]
```diff
@@ -219,7 +221,7 @@ sync_range({Start, End, Peer, TargetStoreID, RetryCount} = Args) ->
 					ok;
 				false ->
 					case ar_http_iface_client:get_chunk_binary(Peer, Start2, any) of
-						{ok, #{ chunk := Chunk } = Proof, Time, TransferSize} ->
+						{ok, #{ chunk := Chunk } = Proof, _, _} ->
```
JamesPiechota, line 224:
```suggestion
						{ok, #{ chunk := Chunk } = Proof, _Time, _TransferSize} ->
```

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -1013,22 +1013,21 @@ get_info(Peer) ->
...
+				headers => p2p_headers(),
+				connect_timeout => 500,
+				timeout => 2 * 1000
+			}),
+		PeerArray = ar_serialize:dejsonify(Body),
+		lists:map(fun ar_util:parse_peer/1, PeerArray)
 	catch _:_ -> unavailable
 	end.
 
+
```
JamesPiechota, line 1030:
```suggestion
```

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -60,18 +60,20 @@ loop(TimeoutRef) ->
...
+					Term -> From ! {read_complete_body, Term}
+				catch
+					Exception:Reason -> From ! {read_complete_body, {Exception, Reason}}
+				end,
+				loop(TimeoutRef);
+			{read_body_chunk, From, Req, Size, Timeout} ->
+				try ar_http_req:read_body_chunk(Req, Size, Timeout) of
+					Term -> From ! {read_body_chunk, Term}
+				catch
+					Exception:Reason -> From ! {read_body_chunk, {Exception, Reason}}
```
JamesPiechota, line 74: Ditto above

apps/arweave/src/ar_block_cache.erl
hlolli: I don't fully understand why I removed this there, maybe I was on auto-pilot, but elvis has no problem with unused underscore variables in most places

example on line 124

apps/arweave/src/ar_inflation.erl
```diff
@@ -50,40 +50,30 @@ calculate_post_15_y1_extra() ->
 pre_15_calculate(Height) when Height =< ?REWARD_DELAY ->
 	1;
 pre_15_calculate(Height) ->
-	?WINSTON_PER_AR
-		* 0.2
-		* ?GENESIS_TOKENS
-		* math:pow(2, -(Height - ?REWARD_DELAY) / ?PRE_15_BLOCK_PER_YEAR)
-		* math:log(2)
-        / ?PRE_15_BLOCK_PER_YEAR.
+	?WINSTON_PER_AR * 0.2 * ?GENESIS_TOKENS *
```
JamesPiechota, line 53: Ditto earlier comment. I'd remove elvis constraints on this sort of thing - long equations are hard to parse and genearlly the engineer writing/maintaining the code will know what formatting is the most intelligible . Something that will be hard for a pattern matching engine to get right

apps/arweave/src/ar_inflation.erl
```diff
@@ -50,40 +50,30 @@ calculate_post_15_y1_extra() ->
...
 calculate_base_pre_fork_2_5(Height) ->
-	?WINSTON_PER_AR
-		* (
-			0.2
-			* ?GENESIS_TOKENS
-			* math:pow(2, -(Height) / ?BLOCK_PER_YEAR)
-			* math:log(2)
-		)
-		/ ?BLOCK_PER_YEAR.
+	?WINSTON_PER_AR * (
```
JamesPiechota, line 71: I'm fine if we have a convention of `*` at the end or the beginning of a line, but I think we should leave line breaks in equations and long lists/tuples upt to the engineer

apps/arweave/src/ar_mempool.erl
```diff
@@ -390,7 +390,7 @@ find_low_priority_txs(Iterator, {MempoolHeaderSize, MempoolDataSize})
 		when
 			MempoolHeaderSize > ?MEMPOOL_HEADER_SIZE_LIMIT;
 			MempoolDataSize > ?MEMPOOL_DATA_SIZE_LIMIT ->
-	{{_Utility, TXID, _Status} = _Element, Iterator2} = gb_sets:next(Iterator),
+				{{_Utility_, TXID, _Status} = _, Iterator2} = gb_sets:next(Iterator),
```
JamesPiechota, line 393: Why does `_Utility_` have a double `_`?

JamesPiechota, line 393:
```suggestion
				{{_Utility_, TXID, _Status} = _Element, Iterator2} = gb_sets:next(Iterator),
```

apps/arweave/src/ar_inflation.erl
hlolli: this rule for spaces around operators can be solved in other ways than I did it, I mode more use of the line here which I can adjust and make these operations be performed on individual lines

apps/arweave/src/ar_merkle.erl
```diff
@@ -15,12 +15,18 @@
 %%% as verification of those proofs.
 -record(node, {
 	id,
-	type = branch,	% root | branch | leaf
-	data,			% The value (for leaves).
-	note,			% The offset, a number less than 2^256.
-	left,			% The (optional) ID of a node to the left.
-	right,			% The (optional) ID of a node to the right.
-	max,			% The maximum observed note at this point.
+	% root | branch | leaf
```
JamesPiechota, line 18: I'd leave this the old way for the same reason as the newlines in long operators. To me the old version is easier to read and understand. It won't always be that way, but I think we should leave it up to the engineers maintaining

hlolli: Elvis is fine with this:

apps/arweave/src/ar_storage.erl
```diff
@@ -299,28 +299,30 @@ put_tx_confirmation_data(B) ->
...
-			end
-	end.
+    case ar_kv:get(tx_confirmation_db, TXID) of
+        {ok, Binary} ->
+            {ok, binary_to_term(Binary)};
+        not_found ->
+            {ok, Config} = application:get_env(arweave, config),
+            case lists:member(arql, Config#config.disable) of
+                true ->
+                    {error, not_found};
```
JamesPiechota, line 309: I think this will break stuff. If there's a rule forcing that all function return values match `{error, XXX}` or `{ok, XXX}` we should disable it.

If we really want to we can enable it later, but it will have wide ranging impact on logic so it should be its own PR

apps/arweave/src/ar_storage.erl
```diff
@@ -299,28 +299,30 @@ put_tx_confirmation_data(B) ->
...
+                    try ar_arql_db:select_block_by_tx_id(ar_util:encode(TXID)) of
+                        {ok, #{
+                            height := Height,
+                            indep_hash := EncodedIndepHash
+                        }} ->
+                            {ok, {Height, ar_util:decode(EncodedIndepHash)}};
+                        _ ->
+                            {error, not_found}
+                    catch
+                        _Class:_Exception ->
```
JamesPiechota, line 320: looks like we lost the `{error, timeout}` return

apps/arweave/src/ar_storage.erl
hlolli: this isn't Elvis, it was my hands getting dirty on cleaning the code where it was not single case statement, I'll revert

apps/arweave/src/ar_block.erl
JamesPiechota: I'm fine with 
```
ValA +
ValB +
ValC.
```

or 

```
ValA
+ ValB
+ ValC
```

and think in general commas should be treated differently from math operators for the reason you mention (commas generally bind to the left value, but it's less clear with binary math operators which value they more logically bind to - think of grammar school long addition, in the US we're taught to put the `+` at the beginning of every line)

But that said, I retract my original point on the `| Elements` and defer to the elvis rule

apps/arweave/src/ar_block_cache.erl
```diff
@@ -71,7 +71,7 @@ add(Tab,
 			SolutionSet3 = sets:from_list([H | Remaining]),
 			[{_, Tip}] = ets:lookup(Tab, tip),
 			[{_, Set}] = ets:lookup(Tab, links),
-			[{_, C = {MaxCDiff, _H}}] = ets:lookup(Tab, max_cdiff),
+			[{_, C = {MaxCDiff, _}}] = ets:lookup(Tab, max_cdiff),
```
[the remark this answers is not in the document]
hlolli: reverted

apps/arweave/src/ar_block_cache.erl
hlolli: reverted

apps/arweave/src/ar_data_sync_worker.erl
hlolli: applied

apps/arweave/src/ar_inflation.erl
hlolli: it's now newlined for each math operation

apps/arweave/src/ar_inflation.erl
hlolli: it's now newlined for each math operation

apps/arweave/src/ar_merkle.erl
hlolli: it's like the screenshot now

apps/arweave/src/ar_storage.erl
hlolli: it's reverted, now it's just not_found and

apps/arweave/src/ar_storage.erl
hlolli: maybe take a look at the file now if it makes sense, I return error, timeout from the catch clause

apps/arweave/src/ar_mempool.erl
hlolli: it's fixed

apps/arweave/src/ar_inflation.erl
```diff
@@ -50,40 +50,30 @@ calculate_post_15_y1_extra() ->
...
-		* EXDivisor
-		* 2
-		* Ln2Dividend
-		div (
-			10
-			* ?BLOCKS_PER_YEAR
-			* Ln2Divisor
-			* EXDividend
-		).
+	?GENESIS_TOKENS * ?WINSTON_PER_AR * EXDivisor * 2 * Ln2Dividend div (
```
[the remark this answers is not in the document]
hlolli: fixed

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -877,17 +878,17 @@ handle(<<"GET">>, [<<"block_index2">>], Req, _Pid) ->
...
+	handle(<<"GET">>, [<<"block_index2">>, From, To], Req, Pid);
 
-handle(<<"GET">>, [<<"block_index2">>, From, To], Req, _Pid) ->
+handle(<<"GET">>, [<<"block_index2">>, From, To], Req, Pid) ->
 	erlang:put(encoding, binary),
-	handle(<<"GET">>, [<<"block_index">>, From, To], Req, _Pid);
+	handle(<<"GET">>, [<<"block_index">>, From, To], Req, Pid);
 
-handle(<<"GET">>, [<<"block_index">>, From, To], Req, _Pid) ->
+handle(<<"GET">>, [<<"block_index">>, From, To], Req, _) ->
```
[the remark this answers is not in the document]
hlolli: fixed

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -60,18 +60,20 @@ loop(TimeoutRef) ->
...
-			case catch ar_http_req:read_body_chunk(Req, Size, Timeout) of
-				Term ->
-					From ! {read_body_chunk, Term}
-			end,
-			loop(TimeoutRef);
+			{read_complete_body, From, Req, SizeLimit} ->
+				try ar_http_req:body(Req, SizeLimit) of
+					Term -> From ! {read_complete_body, Term}
+				catch
+					Exception:Reason -> From ! {read_complete_body, {Exception, Reason}}
```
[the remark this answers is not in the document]
hlolli: maybe @vird or @ldmberman could chip in here, what is returned in this loop code when an error is caught? Are we safe to just return {error, Reason} or should we return nothing?

apps/arweave/src/ar_storage.erl
JamesPiechota: I think the issue is the formatting of returns. Previously only when we caught `{'EXIT', {timeout, {gen_server, call, [ar_arql_db, _]}}}` did we return `{error, timeout}` - now any error will trigger an `{error, timeout}`. I'm not sure if it will make a difference, but I'm a little wary of functionality changes in a large formatting PR.

If there aren't any callers of `get_tx_confirmation_data` that handle `{error, timeout}`, then it's fine and we can go with your earlier iteration of `{error, Exception}` (or whatever)

but if the callers of `get_tx_confirmation_data` handle `{error, timeout}` or handle `{error, SOMETHING}`, then this diff might break those callers handling logic.

hlolli: I don't want any functional change, I'm aware that enforcing single case statement may be mutually exclusive with that goal. But let's look at this:

```sh
> git grep get_tx_confirmation_data                                           
apps/arweave/src/ar_header_sync.erl:250:        case ar_storage:get_tx_confirmation_data(TXID) of
apps/arweave/src/ar_http_iface_middleware.erl:1543:                                     case ar_storage:get_tx_confirmation_data(TXID) of
apps/arweave/src/ar_storage.erl:10:             read_tx/1, read_tx_data/1, update_confirmation_index/1, get_tx_confirmation_data/1,
apps/arweave/src/ar_storage.erl:301:get_tx_confirmation_data(TXID) ->
apps/arweave/test/ar_tx_blacklist_tests.erl:324:                        ?assertMatch({ok, {_, _}}, ar_storage:get_tx_confirmation_data(TXID))
apps/arweave/test/ar_tx_blacklist_tests.erl:351:                        ?assertMatch({ok, {_, _}}, ar_storage:get_tx_confirmation_data(TXID)
```

Excluding tests, we have callers in ar_http_iface_middleware.erl and ar_header_sync.erl

Looking at ar_http_iface_middleware.erl
it pattern matches on

```erl
{ok, {Height, BH}} ->
.... elided ...
						not_found ->
							{404, #{}, <<"Not Found.">>, Req};
						{error, not_found} ->
							{404, #{}, <<"Not Found.">>, Req};
						{error, timeout} ->
							{503, #{}, <<"ArQL unavailable.">>, Req}
```

and ar_header_sync.erl

which pattern matches on

```erl
		{ok, {Height, _BH}} ->
.....
		not_found ->
			ar_events:send(tx, {ready_for_unblacklisting, TXID}),
			{noreply, State};
		{error, Reason} ->
			?LOG_WARNING([{event, failed_to_read_tx_confirmation_index},
					{error, io_lib:format("~p", [Reason])}]),
			{noreply, State}
```

the former one only matches timeout errors, while the latter one matches any error. So any other error, Reason that isn't timeout would not fly with ar_http_iface_middleware but would be ok in ar_header_sync.

I can write up a ticket to make it consistent and allow any error reason to be sent and received, because surely timeout is only one of meany reasons.

JamesPiechota: Given what the callers are doing, and what the behavior was before, it looks like what you have now is good. The only change in functionality is now it will return {error, timeout} for some scenarios where it crashed outright before. I think that change in functionality is okay. (assuming I understand the whole `{'EXIT', {...` match which I'm not sure i do)

PR #503, opened 2023-11-27
Address 3 mining performance bottlenecks

JamesPiechota:
This PR addresses 3 mining performance bottlenecks

### 1. Overly aggressive mining session invalidation
Previously the mining server maintained a single active mining session tied to the VDF session of its most recently received VDF step. When that mining session changed, any mining tasks opened against a prior session were invalidated and discarded. This led to scenarios where the mining server would read an entire 400 chunk recall range from disk, only to discard the chunks before hashing them. Under normal situations this was a rare edge case that only occurred at the VDF session boundaries (e.g. once or twice ever 20 minutes). However if the mining server ever fell behind, or perhaps its VDF server jumped ahead, the server would receive VDF steps out of order. Occasionally this caused a "flutter" of session invalidations as steps from two different sessions were received alternatively - each time causing some number of mining tasks to be invalided and discarded.

This PR addresses this by having the mining server maintaining 2 concurrent valid sessions and keep them synced with the the current and previous VDF sesssions. This eliminates the flutter (a very old step from 3 sessions ago, will simply be discarded without wasting any CPU or read bandwidth on it), and also keeps a longer window open so that steps at session boundaries aren't lost.

### 2. Single-threaded mining server task queue doesn't scale with number of partitions
Previously the mining server maintained a single-threaded task queue. In general processing a task from the queue is very fast - most of the work is offloaded to separate io and hashing threads. However even if the slowest operation takes a fraction of a millisecond - it can still become a bottleneck as a miner has more partitions. For example at a full weave of 40 partitions the task queue needs to process the following operations every VDF step:

- 1 `computed_output` task
- 16,000 `io_thread_recall_range_chunk` tasks
- 16,000 `io_thread_recall_range2_chunk` tasks
- 40 `mining_thread_computed_h0` tasks
- 16,000 `mining_thread_computed_h1`
- 16,000 `mining_thread_computed_h2`

Even assuming a 1 second VDF step and a fraction of a millisecond for each task - that volume of tasks can overwhelm a single thread. On a Ryzen 9 7950x (3Ghz), it takes 1.5 seconds to process all the tasks for a 12 partition setup (separate from the actual hashing and disk reads - this is 1.5 seconds of pure task management overhead). 

This PR addresses this by sharding the task queue across multiple mining workers (1 per core). This works nicely since a full mining cycle (h0 -> read chunk 1 -> h1 -> read chunk 2 -> h2) can be managed by a single worker pretty cleanly. The end result is a mining task queue that can scale with the number of cores on a machine.

### 3. Stale cache items
Previously there were some edge cases in the chunk cache management that caused some chunks to remain past their usefulness. This limited the cache space available for new chunks and, depending on the miner configuration, could have a measurable impact on performance.

### 4. Memory spike on launch
When booting a node a ~300MB initialization message was sent to more processes than was needed. This caused a spike in memory usage that could take some time to be garbage collected and, occasionally, could cause an out of memory exception if the spike coincided with other high-memory use operations (e.g. syncing and mining).

This PR addresses the spike by limiting the number of processes that receive the initialization message.

### 5. Some refactoring
- Broke the hashing threads out to their own `ar_mining_hash` gen_server
- Refactored some of the process supervisor config into a `?CHILD_WITH_ARGS` macro to complement the `?CHILD` macro

### 6. Introduce data-doctor utility
Added a new utility to handle data-related maintenance. Right now it only includes 2 commands:
- `data-doctor merge`: merge multiple storage modules into 1
- `data-doctor bench`: runs a disk read benchmark that simulates the performance you can expect when reading chunks while mining. This differs from `fio` or `dd` benchmarks as it uses the same code paths and access patterns used during mining.


apps/arweave/src/ar_events_sup.erl
```diff
@@ -42,7 +42,7 @@ init([]) ->
 		?CHILD(ar_events, chunk, worker),
 		%% Events: removed
 		?CHILD(ar_events, peer, worker),
-		%% Events: initializing, account_tree_initialized, initialized,
```
JamesPiechota, line 45: The initializing event was only handled by ar_nonce_limiter - but since so many processes listen to `node_state` it was causing a spike in memory usage during launch. On launch the block list shared via initializing can be ~300MB, so each process that listens to `node_state` (even if it doesn't handling`initializing`) receives the block list and temporarily stores it locally within the process. The GC can then take some time to clean it up - which can then cause trouble when other memory-intensive operations start up (e.g. syncing and mining)

JamesPiechota: The fix is to remove the `initializing` event and just have `ar_node_worker` notify `ar_nonce_limiter` directly

apps/arweave/src/ar_mining_hash.erl
```diff
@@ -0,0 +1,128 @@
...
+-export([init/1, handle_cast/2, handle_call/3, handle_info/2, terminate/2]).
+
+-include_lib("arweave/include/ar.hrl").
+-include_lib("arweave/include/ar_config.hrl").
+-include_lib("arweave/include/ar_consensus.hrl").
+-include_lib("arweave/include/ar_mining.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+-record(state, {
+	hashing_threads				= queue:new(),
```
hlolli, line 16:
```suggestion
	hashing_threads = queue:new(),
```

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -0,0 +1,535 @@
...
+				false ->
+					?LOG_DEBUG([{event, mining_debug_no_io_thread_found_for_range},
+						{worker, State#state.name},
+						{range_start, RecallRange1Start},
+						{range_end, RecallRange1Start + ?RECALL_RANGE_SIZE}]),
+					ok
+			end;
+		false ->
+			%% Re-add the task so that it can bd executed later once some cache space frees up.
+			add_task(self(), computed_h0, Candidate)
```
vird: Maybe we should fix it in future.
1. Limit pending h0 tasks
Because we need to check how many h0 tasks are already in queue. So if this task is oldest and we hit some limit on h0 tasks, so we don't add it because we are overloaded
2. Decrease priority to prevent CPU waste
If queue is already filled with lots of tasks dry-run pull/push is not kinda good thing. So if this task was pulled with some priority and we can't handle it right now, we should push it with reduced priority so ensure we will not pull this same task again and waste CPU one more time

JamesPiechota: the tasks are currently sorted based on task type and step number. `compute_h0` is the lowest priority, `computed_h0` is the next lowest. Within a task type the older the step number the lower the priority. So this task is already likely to be the highest priority pending task (typically the h1/h2 and chunk1/chunk2 tasks get handled immediately, and it's only the compute_h0 and computed_h0 that can get queued)

Also every time there's a new VDF session, any pending tasks for the retired VDF session are purged from the queue.

I agree the pop/push behavior isn't ideal, but I think it's also negligible in terms of performance. In my investigation it wasn/t the gb_sets operations that were slow, but the work that happens when a task is actually executed (and even that was only a problem at high partition counts for a single thread - so the new parallelized task queue should address all the task management bottlenecks that I've seen)

apps/arweave/src/ar_mining_hash.erl
JamesPiechota: this formatting is consistent with the ar_mining_server formatting (which is where this file was split out from). I don't feel strongly about whether it's better to line up the assignments or not (or whether it's even important enough to standardize), but I'm reluctant to change it in this one place since it is consistent with how this group of files was formatted in the past.

hlolli: they didn't line up the equal sign on line 16 and 17, could be github's graphical representation, feel free to ignore

JamesPiechota: yeah, probably the tabs vs. spaces issue.

JamesPiechota: If we want ot move the needle on that, my suggestion is we break it into two steps:
1. enable the linter and have it enforce the current repo standards (e.g. tabs)
2. once that's enabled and running smoothly (and if we decide that spaces really is better than tabs), we find a time to convert the whole repo to spaces and update the linter rules

Step 1 I think could be very useful. I don't feel strongly about step 2 so long as we can find a time to do the conversion.

apps/arweave/src/ar_mining_server.erl
```diff
@@ -228,89 +174,157 @@ handle_cast({post_solution, Solution}, State) ->
...
-terminate(_Reason, #state{ hashing_threads = HashingThreads }) ->
-	[Thread ! stop || Thread <- queue:to_list(HashingThreads)],
+terminate(_Reason, _State) ->
 	ok.
 
 %%%===================================================================
 %%% Private functions.
 %%%===================================================================
+update_upper_bound(PartitionUpperBound, State) ->
+	#state{ partition_upper_bound = CurrentUpperBound } = State,
```
ldmberman, line 241: Do we still need partition_upper_bound in State?

Also, what do you think about naming the function something like `update_largest_seen_upper_bound` to make it clear it's not the bound used for picking recall ranges?

apps/arweave/src/ar_mining_sup.erl
```diff
@@ -0,0 +1,42 @@
...
+
+start_link() ->
+	supervisor:start_link({local, ?MODULE}, ?MODULE, []).
+
+%% ===================================================================
+%% Supervisor callbacks.
+%% ===================================================================
+
+init([]) ->
+	%% Largest multiple of 2 less than the number of schedulers (but no less than 2).
```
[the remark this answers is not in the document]
JamesPiechota: Yeah... I don't love it, but it does allow the ar_mining_worker to be much simpler as it only has to keep track of a single session. And given the severity of the task queue bottleneck I think 4 workers is enough for a decent core to handle a full weave.... So having half the workers idle didn't seem like a huge problem (although I agree it's icky)

apps/arweave/src/ar_mining_server.erl
JamesPiechota: Good point - we don't! removed. https://github.com/ArweaveTeam/arweave/pull/503/commits/e34a6961ddc613b2febc2eaaebba3e88c568ac41

apps/arweave/test/ar_mining_server_tests.erl
```diff
@@ -178,7 +178,8 @@ do_test_chunk_cache_size_with_mocks(H1s, H2s, RecallRange2s, FirstChunks) ->
 	try
 		ar_test_node:mine(),
 		ar_test_node:wait_until_height(Height),
-		ar_test_node:wait_until_mining_paused(),
+		%% wait until the mining has stopped
+		timer:sleep(10000),
```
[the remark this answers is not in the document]
JamesPiechota: Done! https://github.com/ArweaveTeam/arweave/pull/503/commits/e34a6961ddc613b2febc2eaaebba3e88c568ac41

apps/arweave/src/ar_coordination.erl
```diff
@@ -83,7 +74,8 @@ init([]) ->
 	
 	%% using timer:apply_after so we can cancel pending timers. This allows us to send the
 	%% h1 batch as soon as it's full instead of waiting for the timeout to expire.
-	{ok, H1BatchTimerRef} = timer:apply_after(?BATCH_TIMEOUT_MS, ?MODULE, send_h1_batch_to_peer, []),
+	{ok, H1BatchTimerRef} = timer:apply_after(
```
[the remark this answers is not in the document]
JamesPiechota: Ah, thanks. Done! https://github.com/ArweaveTeam/arweave/pull/503/commits/e34a6961ddc613b2febc2eaaebba3e88c568ac41

PR #505, opened 2023-12-01
feat: improved https tls protocol support

hlolli:


apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -64,128 +76,95 @@ start() ->
...
-	Domain = Config#config.gateway_domain,
-	CustomDomains = Config#config.gateway_custom_domains,
-	Dispatch = cowboy_router:compile([{'_', ?HTTP_IFACE_ROUTES}]),
-	HttpIfaceEnv = #{ dispatch => Dispatch },
-	SniHosts = derive_sni_hosts(CustomDomains),
+start_https_iface_listener(Config) ->
+	DataDir = Config#config.data_dir,
+	AlternativeHostnames = Config#config.alternative_hostnames,
+	CertfilePath = filename:join([DataDir, "tls/cert.pem"]),
+	KeyfilePath = filename:join([DataDir, "tls/key.pem"]),
```
vird, line 117: +TODO add watcher which will monitor changes of certificate files, because let's encrypt certbot can silently modify those files so your HTTPS will be outdated in several months because renewed and stored cert is not used by your app until restart

vird:
LGTM

apps/arweave/include/ar_config.hrl
```diff
@@ -186,6 +184,8 @@
 	defragmentation_modules = [],
 	block_throttle_by_ip_interval = ?DEFAULT_BLOCK_THROTTLE_BY_IP_INTERVAL_MS,
 	block_throttle_by_solution_interval = ?DEFAULT_BLOCK_THROTTLE_BY_SOLUTION_INTERVAL_MS,
+	tls_cert_file = not_set, %% required to enable TLS
+  tls_key_file = not_set,  %% required to enable TLS
```
vird, line 188: Even spacing suggested

apps/arweave/src/ar.erl
```diff
@@ -498,6 +502,14 @@ parse_cli_args(["defragment_module", DefragModuleString | Rest], C) ->
 		io:format("~ndefragment_module value must be in the [number],[address] format.~n~n"),
 		erlang:halt()
 	end;
+parse_cli_args(["tls_cert_file", CertFilePath | Rest], C) ->
+    AbsCertFilePath = filename:absname(CertFilePath),
+    ar_util:assert_file_exists_and_readable(AbsCertFilePath),
```
JamesPiechota, line 507: Rather than `throw` what about doing an `io:format` + `erlang:halt`? I think that's what we do elsewhere in the cli parsing (e.g. [here](https://github.com/ArweaveTeam/arweave/pull/505/files#diff-f148c1ce17ebe7e20d6eed136cc4f0c0e517b978a199351bb7e50aca826a8de4R420))

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -65,157 +64,34 @@ start() ->
...
-	cacertfile_when_present(apex_domain).
-
-cacertfile_when_present(Domain) ->
-	PrivDir = code:priv_dir(arweave),
-	Filename = case Domain of
-		apex_domain ->
-			filename:join([PrivDir, "tls/cacert.pem"]);
-		CustomDomain ->
-			filename:join([PrivDir, "tls", CustomDomain, "cacert.pem"])
+			{ok, _} = cowboy:start_clear(ar_http_iface_listener, TransportOpts, ProtocolOpts);
```
JamesPiechota, line 87: This is branch with TLS, right? It looks like the TransportOpts and ProtocolOpts have changed vs. what they were before. Just calling that out - I don't know enough to comment onw hether the changes could cause a problem with the non-TLS flow

JamesPiechota: ok it looks like it's only `max_connections` vs. `max_concurrent_streams` that's changed. Same comment though: I have no idea if that change matters - just flagging it

apps/arweave/include/ar_config.hrl
hlolli: not practicing what I'm preaching, fixed

apps/arweave/src/ar.erl
```diff
@@ -498,6 +502,14 @@ parse_cli_args(["defragment_module", DefragModuleString | Rest], C) ->
 		io:format("~ndefragment_module value must be in the [number],[address] format.~n~n"),
 		erlang:halt()
 	end;
+parse_cli_args(["tls_cert_file", CertFilePath | Rest], C) ->
```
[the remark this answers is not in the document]
hlolli: I added it to the help message, yeh no harm in having it there

apps/arweave/src/ar_http_iface_server.erl
hlolli: this branch on line 87 is normal http, and that doesn't need any adjustments

apps/arweave/src/ar_config.erl
```diff
@@ -392,20 +392,18 @@ parse_options([{<<"disable">>, Features} | _], _) ->
 	{error, {bad_type, disable, array}, Features};
 
 parse_options([{<<"gateway">>, Domain} | Rest], Config) when is_binary(Domain) ->
```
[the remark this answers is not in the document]
hlolli: good catch, it's removed

apps/arweave/src/ar_config.erl
```diff
@@ -392,20 +392,18 @@ parse_options([{<<"disable">>, Features} | _], _) ->
...
-		true ->
-			parse_options(Rest, Config#config{ gateway_custom_domains = CustomDomains });
-		false ->
-			{error, bad_custom_domains}
-	end;
+parse_options([{<<"custom_domains">>, _} | Rest], Config) ->
+	?LOG_WARNING("Deprecated option found 'custom_domains': "
+			" this option has been removed and is a no-op.", []),
+	parse_options(Rest, Config);
 parse_options([{<<"custom_domains">>, CustomDomains} | _], _) ->
```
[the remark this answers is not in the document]
hlolli: removed the guard

apps/arweave/src/ar_http_iface_server.erl
hlolli: the latter comment I didn't see, now I see where you're coming from, yes it seems the api changed, or rather, changed even in earlier version, maybe there's backwards compatibility built in.

ProtocolOpts: https://ninenines.eu/docs/en/ranch/2.1/manual/ranch_tcp/

I wonder if it's even removed, checking...

hlolli: ok I'm looking at http2 manual, but we are using http1.1, let me fix that

hlolli: https://ninenines.eu/docs/en/cowboy/2.10/manual/cowboy_http2/

hlolli: reverted, max_connections is a part of the TransportOption as it was. We can research further ranch upgrades and the options needed for seamless migration, for another day.

apps/arweave/src/ar_config.erl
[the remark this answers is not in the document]
JamesPiechota: the `when is_binary` is still there in my diff. Do you need to push? Or maybe I misunderstood?

JamesPiechota (approved):

PR #515, opened 2024-01-04
Feature/pool

ldmberman:


apps/arweave/include/ar_config.hrl
```diff
@@ -191,7 +191,11 @@
 	cm_api_secret = not_set,
 	cm_exit_peer = not_set,
 	cm_peers = [],
-	cm_poll_interval = ?DEFAULT_CM_POLL_INTERVAL
+	cm_poll_interval = ?DEFAULT_CM_POLL_INTERVAL,
+	pool_server = false,
+	pool_client = false,
+	pool_host = not_set,
+	pool_api_key = not_set
```
JamesPiechota, line 198: The idea is to use a shared secret here? like with the `cm_api_secret`

apps/arweave/include/ar_pool.hrl
```diff
@@ -0,0 +1,35 @@
...
+
+%% The frequency in milliseconds of asking the pool proxy or CM exit node about new jobs.
+-define(FETCH_JOBS_FREQUENCY_MS, 5000).
+
+%% The time in milliseconds we wait before retrying a failed fetch jobs request.
+-define(FETCH_JOBS_RETRY_MS, 2000).
+
+%% @doc A collection of mining jobs.
+-record(jobs, {
+	vdf = [], % The information about a single VDF output (a "job").
```
JamesPiechota, line 16: Can we name this `job` instead of `vdf`?

apps/arweave/include/ar_pool.hrl
```diff
@@ -0,0 +1,35 @@
...
+	diff = 0, % Partial difficulty.
```
JamesPiechota, line 17: What about `partial_diff` instead of `diff`?

apps/arweave/src/ar.erl
```diff
@@ -269,7 +269,13 @@ show_help() ->
...
 			{"cm_exit_peer (IP:port)", "The peer to send mining solutions to in the "
 					"coordinated mining mode. You need to also set coordinated_mining, "
-					"cm_api_secret, and cm_peer."}
+					"cm_api_secret, and cm_peer."},
+			{"pool_server", "Configure the node as a pool server. The pool node may not "
+					"participate in the coordinated mining."},
+			{"pool_client", "Configure the node as a pool client. The node may be an "
+					"exit peer in the coordinated mining setup or a standalone node."},
+			{"pool_api_key", "API key for the requests to pool proxy."},
+			{"pool_host", "The pool proxy address"}
```
JamesPiechota, line 278: What do you think about `pool_proxy` instead of `pool_host`?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -504,6 +507,66 @@ handle(<<"POST">>, [<<"block2">>], Req, Pid) ->
 	erlang:put(post_block2, true),
 	post_block(request, {Req, Pid, binary}, erlang:timestamp());
 
+%% Accept a (partial) solution from a pool proxy or a CM node and validate it.
```
JamesPiechota, line 510: I think I'm getting a little lost about all the different entities in the system. I see:
- Pool Server
- Pool Proxy
- Standalone Pool Client
- CM Exit Node Pool Client

Is that all of them? Any chance you can put an ascii diagram in somewhere (maybe `ar_pool.erl`) to clarify the communication? Is it something like

```

                                +---> Standalone Pool Client
                                |
Pool Server <--> Pool Proxy <---+
                                |
                                +---> CM Exit Node Pool Client <--> CM Miner
```

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -236,16 +224,6 @@ pre_validate_proof_sizes(B, PrevB, Peer) ->
 			invalid
 	end.
 
-validate_proof_size(PoA, Height) ->
```
JamesPiechota, line 239: Not directly relevant, but a question I've had: is there a rule of thumb for knowing which code doesn't need to be aware of old fork heights?

Like any code that processes *new* blocks or transactions can remove the old fork checking code. But any code that needs to process *old* blocks or transactions still need to be aware of historical forks, right? (e.g. validating the full chain from genesis)

Are there certain modules (like ar_block_pre_vaidator?) that are only invoked for new blocks/transaction and other modules that might be invoked on really old blocks?

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -709,24 +687,11 @@ pre_validate_poa(B, PrevB, PartitionUpperBound, H0, H1, Peer) ->
...
-				case B#block.height == ar_fork:height_2_7() of
-					true ->
-						PrevB#block.weave_size;
-					false ->
-						infinity
-				end
-		end,
-	ArgCache = {BlockStart1, RecallByte1, TXRoot1, BlockSize1, ?STRICT_DATA_SPLIT_THRESHOLD,
-			{spora_2_6, B#block.reward_addr}, RebaseThreshold},
+	ArgCache = {BlockStart1, RecallByte1, TXRoot1, BlockSize1,
```
JamesPiechota, line 690: Do we also have to update these lines to account for the ArgCache structure?

https://github.com/ArweaveTeam/arweave/pull/515/files#diff-c71b6858e1c10e9f544c7eb46e9f333f4d95f2afb7f2ee2e7eff5dbe1ba23871R409

https://github.com/ArweaveTeam/arweave/pull/515/files#diff-c71b6858e1c10e9f544c7eb46e9f333f4d95f2afb7f2ee2e7eff5dbe1ba23871L441

apps/arweave/include/ar_pool.hrl
JamesPiechota: Hm, actually it looks like we refer to this as 3 diferent things: `vdf`, `job`, and then later `triplet`. Can we unify the 3 different names?

What about moving the `get_step_triplets` and related functions out of `ar_nonce_limiter` and putting them in `ar_pool` and then using the term `job` everywhere (i.e. replace `vdf` here, and `triplet` there with `job`)

apps/arweave/src/ar.erl
```diff
@@ -586,6 +601,26 @@ start(Config) ->
 	end,
 	start_dependencies().
 
+validate_cm_pool_config(Config) ->
+	case {Config#config.coordinated_mining, Config#config.pool_server} of
+		{true, true} ->
+			io:format("~nThe pool server node cannot participate "
```
JamesPiechota, line 607: Possible to add a comment explaining why the server can't participate in coordinated mining?

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: And actually it might also help to diagram how the partial solution is shared. It sounds like there are 2 different flows:

1. No coordinated mining:
```
Standalone Pool Client -- partial solution -> Pool Proxy -- partial solution -> Pool Sever
```

2. Coordinated mining:
```
CM Miner -- partial solution -> CM Exit Node Pool Client -- partial solution -> Pool Proxy -- partial solution -> Pool Server
```

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -630,6 +630,74 @@ cm_publish_send(Peer, Solution) ->
 		body => ar_serialize:jsonify({Json})
 	})).
 
+%% @doc Fetch the jobs from the pool proxy or coordinated mining exit peer.
+get_jobs(PeerOrURL, PrevOutput, Pool) ->
```
JamesPiechota, line 634: What does `Pool` refer to here?

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -630,6 +630,74 @@ cm_publish_send(Peer, Solution) ->
...
+	{Peer, Headers, BasePath} =
+		case Pool of
```
JamesPiechota, line 636: Can this block and [this one](https://github.com/ArweaveTeam/arweave/pull/515/files#diff-e6210d30a0b0d54dd30e8bfa45bba38a93369c1b0b19419a8f5891e163e45c40R656-R662) be deduplicated?

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -1184,6 +1252,10 @@ cm_p2p_headers() ->
 	{ok, Config} = application:get_env(arweave, config),
 	add_header(<<"x-cm-api-secret">>, Config#config.cm_api_secret, p2p_headers()).
 
+pool_client_headers() ->
+	{ok, Config} = application:get_env(arweave, config),
+	add_header(<<"x-pool-api-key">>, Config#config.pool_api_key, p2p_headers()).
```
JamesPiechota, line 1257: I forget where we landed on TLS/SSL - right now by default all peers communicate unencrypted via HTTP. If that's the case I think using a shared secret will be a problem.

apps/arweave/src/ar_mining_server.erl
```diff
@@ -594,7 +592,17 @@ post_solution(not_set, Solution, State) ->
...
+		{ok, _} ->
+			ok;
+		{error, Reason} ->
+			?LOG_WARNING([{event, found_partial_solution_but_failed_to_reach_exit_node},
+					{reason, io_lib:format("~p", [Reason])}]),
+			ar:console("We found a partial solution but failed to reach the exit node, "
+					"error: ~p.", [io_lib:format("~p", [Reason])])
 	end;
 post_solution(ExitPeer, Solution, _State) ->
 	case ar_http_iface_client:cm_publish_send(ExitPeer, Solution) of
```
JamesPiechota, line 608: Do we still need `cm_publish_send`? It'd be great to remove it and just using `post_partial_solution`

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: Also (and this may just be my ignorance) it would be great to define "partial solution" somewhere. As I understand it, it's a valid solution that satisfies a lower difficulty threshold than is required to publish a block. *But* it can also satisfy the higher difficulty threshold. So the set of all partial solutions includes the set of all full solutions - right?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2509,6 +2572,118 @@ post_block(enqueue_block, {B, Peer}, Req, ReceiveTimestamp) ->
...
+			{200, #{}, jiffy:encode(#{}), Req2};
+		{error, body_size_too_large} ->
+			{413, #{}, <<"Payload too large">>, Req};
+		{error, timeout} ->
+			{500, #{}, <<"Handler timeout">>, Req}
+	end.
+
+handle_get_jobs(PrevOutput, Req) ->
+	{ok, Config} = application:get_env(arweave, config),
+	CMExitNode = Config#config.coordinated_mining == true andalso
```
JamesPiechota, line 2629: Can this check be refactored somewhere and deuplicated? Maybe in `ar_config.erl` or `ar_coordination.erl`?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -504,6 +507,66 @@ handle(<<"POST">>, [<<"block2">>], Req, Pid) ->
...
+%% If the given VDF output is present in the latest 10 VDF steps, return only the steps
+%% strictly above the given output. If the given output is our latest output,
+%% wait for up to ?GET_JOBS_TIMEOUT_S and return an empty list if no new steps are
+%% computed by the time. Also, only return the steps strictly above the latest block.
+%%
+%% If we are a pool server, return the current network difficulty along with the VDF
+%% information.
+%%
+%% If we are a CM exit node and a pool client, return the partial difficulty provided
+%% by the pool.
```
JamesPiechota, line 543: For my own understanding  - why do we have CM miners communicate with the pool proxy via the CM exit node? e.g. why don't they query jobs directly from the pool proxy?

apps/arweave/src/ar_serialize.erl
```diff
@@ -1820,3 +1825,214 @@ decode_if_set(JSON, JSONProperty, Decoder, Default) ->
...
+	{Keys} = Struct,
+	Output = ar_util:decode(proplists:get_value(<<"output">>, Keys, <<>>)),
+	StepNumber = binary_to_integer(proplists:get_value(<<"global_step_number">>, Keys,
+			<<"0">>)),
+	PartitionUpperBound = binary_to_integer(proplists:get_value(<<"partition_upper_bound">>,
+			Keys, <<"0">>)),
+	#job{ output = Output, global_step_number = StepNumber,
+			partition_upper_bound = PartitionUpperBound }.
+
+partial_solution_to_json_struct(
```
JamesPiechota, line 1871: Since a "solution" is also a "partial solution" (I think?) can we remove this distinction an only have a single solution serializer/deserializer? (e.g. remove `solution_to_json_struct` or remove `partial_solution_to_json_struct`)

apps/arweave/src/ar_mining_server.erl
```diff
@@ -293,6 +264,16 @@ refresh_workers(State) ->
 	WorkersBySession2 = refresh_workers(SessionsToAdd, SessionsToRemove2, WorkersBySession),
 	State#state{ workers = WorkersBySession2 }.
 
+get_current_session_keys() ->
+	case ar_pool:is_client() of
+		false ->
+			{CurrentSessionKey, CurrentSession} = ar_nonce_limiter:get_current_session(),
```
JamesPiechota, line 270: What do you think about moving the session logic into `ar_nonce_limiter`? 

Like instead of doing the `case ar_pool:is_client()` when calling out to `ar_nonce_limiter`, have ar_nonce_limiter do the `case ar_pool:is_clien()` itself and return the correct sessions. This could localize the complexity of determining the correct sessions to the `ar_nonce_limiter` module and allow callers of the `ar_nonce_limiter` to be agnostic to how the sessions are determined

apps/arweave/src/ar.erl
```diff
@@ -586,6 +601,26 @@ start(Config) ->
 	end,
 	start_dependencies().
 
+validate_cm_pool_config(Config) ->
```
JamesPiechota, line 604: What happens if a node is a VDF Client *and* a pool client? Will the VDF updates from the VDF server be ignored?

JamesPiechota: Hm, or I guess the VDF Client uses the updates from teh VDF server to validate blocks, but uses the Jobs from the Pool server for mining? What happens if they don't agree? Like if the node receives a Job from the server which it can't validate based on the VDF chain it has received from the VDF Server?


What do you think about only allowing a node to be *either* a Pool Client or a VDF client? And if it's a Pool Client it gets all VDF updates from the Pool Server?

apps/arweave/src/ar_node_worker.erl
```diff
@@ -493,59 +493,110 @@ handle_info({event, miner, {found_solution, Solution, PoACache, PoA2Cache}}, Sta
...
+	%% Check solution difficulty
+	Diff = get_current_diff(Timestamp),
+	PassesDiffCheck =
+		case PassesSeedCheck of
+			{false, Reason2} ->
+				{false, Reason2};
+			true ->
+				case binary:decode_unsigned(SolutionH, big) > Diff of
+					false ->
+						ar_events:send(solution, {processed, #{ source => Source }}),
```
JamesPiechota, line 550: what does `processed` refer to here?

apps/arweave/src/ar_node_worker.erl
```diff
@@ -493,59 +493,110 @@ handle_info({event, miner, {found_solution, Solution, PoACache, PoA2Cache}}, Sta
...
+	end,
+	PassesKeyCheck =
+		case PassesDiffCheck of
+			{false, Reason3} ->
+				{false, Reason3};
+			true ->
+				case RewardKey of
+					not_found ->
+						ar_events:send(solution,
+							{rejected, #{ reason => missing_key_file, source => Source }}),
```
JamesPiechota, line 574: Why do some of the `solution` events include `reason` value and others don't?

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -449,6 +483,29 @@ handle_call({get_step_checkpoints, StepNumber, SessionKey}, _From, State) ->
...
+		#vdf_session{ step_checkpoints_map = Map, seed = Seed,
+				upper_bound = UpperBound, next_upper_bound = NextUpperBound } ->
+			case maps:get(StepNumber, Map, not_found) of
+				not_found ->
+					{reply, not_found, State};
+				L ->
+					{_NextSeed, IntervalNumber, _NextVDFDifficulty} = SessionKey,
+					IntervalStart = IntervalNumber * ?NONCE_LIMITER_RESET_FREQUENCY,
+					UpperBound2 =
+						case get_entropy_reset_point(IntervalStart, StepNumber) of
```
JamesPiechota, line 499: Do we also need to choose Seed vs. NextSeed based on the entropy reset point here?

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -1232,6 +1289,34 @@ debug_double_check(Label, Result, Func, Args) ->
...
+		false ->
+			[Triplet | filter_step_triplets(Triplets, LowerBounds)]
+	end.
+
+get_triplets(_StepNumber, _Steps, _ResetPoint, _UpperBound, _NextUpperBound, 0) ->
+	[];
+get_triplets(_StepNumber, [], _ResetPoint, _UpperBound, _NextUpperBound, _N) ->
+	[];
+
+get_triplets(StepNumber, [Step | Steps], ResetPoint, UpperBound, NextUpperBound, N) ->
```
JamesPiechota, line 1307: Is there any chance, if N is big enough, that we might through 2 reset points?

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -118,6 +127,17 @@ get_step_checkpoints(StepNumber, NextSeed, StartIntervalNumber, NextVDFDifficult
 get_step_checkpoints(StepNumber, SessionKey) ->
 	gen_server:call(?MODULE, {get_step_checkpoints, StepNumber, SessionKey}, infinity).
 
+%% @doc Return the cached last step checkpoints, seed, and partition upper bound
+%% for the given step.
+%% Return not_found if the VDF session or cached last step checkpoints are not found.
+get_step_checkpoints_seed_upper_bound(StepNumber, NextSeed, StartIntervalNumber,
```
JamesPiechota, line 133: As a way to better understand how all this works together I'm trying to wrap my head around the need to have both `get_step_checkpoints_seed_upper_bound` and `get_triplets`... is there any way to merge the two? Or name them something that clarifies how they're used?

`get_step_checkpoints_seed_upper_bound` in particular is a mouthful and I find myself getting lost as I try to keep everything in my head as I trace through the code. Could we split it and use `get_step_checkpoints` as is? And add an additional call or calls to query the seed and upper bound?

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+-include_lib("arweave/include/ar_config.hrl").
+-include_lib("arweave/include/ar_consensus.hrl").
+-include_lib("arweave/include/ar_mining.hrl").
+-include_lib("arweave/include/ar_pool.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+-record(state, {
+	%% The most recent keys come first.
+	session_keys = [],
+	%% Key => [{Output, StepNumber, PartitionUpperBound, Seed, Diff}, ...]
```
JamesPiechota, line 20: Instead of a 5-tuple, what do you think about defining a record to store these values? It actually feels like `#job` works here, right? So maybe we need to change the `#job` that is defined in `ar_pool.hrl`? I'm getting a little twisted about what a job actually is... but It seemsliek the 5 values here are the input requirements for a pool client to do work, so feels like it might be a good definition of `job`

apps/arweave/include/ar_pool.hrl
JamesPiechota: Okay now having read though `ar_pool.erl` I'm not sure of naming. But I do think we can clarify things... triplet vs. vdf vs. job vs. the 5-tuple in ar_pool.erl (https://github.com/ArweaveTeam/arweave/pull/515/files#r1444973868) all feel pretty similar to me.


What about using the `job` record here to store the 5 values from the 5-tuple in `ar_pool.erl`... and then maybe coming up with a better name for `triplet`? I kinda feel like the full definition of a "step" actually includes `output`, `step_number`, `partition_upper_bound` and `seed`... 

Not sure... but I am having trouble keeping all the different terms straight and when they overlap vs. differ... so I"m expecting some pain down the road once we go live and have to start fixing bugs

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+	SessionKeys = State#state.session_keys,
+	JobCache = State#state.jobs_by_session_key,
+	{reply, get_jobs(PrevOutput, SessionKeys, JobCache), State};
+
+handle_call(get_latest_output, _From, State) ->
+	case State#state.session_keys of
+		[] ->
+			{reply, <<>>, State};
+		[Key | _] ->
+			{O, _SN, _U, _S, _Diff} = hd(maps:get(Key, State#state.jobs_by_session_key)),
```
JamesPiechota, line 86: Rather than unpack and return the latest output. Could we remove this functino an djust have a `get_latest_job` that returns the latest `#job` record, and then the caller can access whatever field it needs?

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+			{reply, O, State}
+	end;
+
+handle_call({process_partial_solution, Solution}, From, State) ->
+	#state{ request_pid_by_ref = Map } = State,
+	Ref = make_ref(),
+	case process_partial_solution(Solution, Ref) of
```
JamesPiechota, line 93: The `process_partial_solution` series of checks mirrors `prepare_solution` and `post_solution` checks in `ar_mining_server` in that it performs a series of validations and then ends by sending a `found_solution` event. Any way we can align the meaning to make that similarity/symmetry more explicit? E.g. rename `process_partial_solutino` to `prepare` or `post` - or rename the functions in `ar_mining_server` to `process`

Also `ar_mining_server` uses `atom` matching to differentiate different steps in the process (e.g. `prepare_solution(poa2, ....)` vs. here adds onto the name of the fucntion (e.g. `process_partial_solution_poa2(...)`

Can we align those two conventions? either both use `atom` or both use function name? I think it will also make reviewing/updating/understanding the code easier in the future.

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+		{32, 32, 32, 32, 32, 32, L1, L2} when L1 =< ?DATA_CHUNK_SIZE, L2 =< ?DATA_CHUNK_SIZE ->
+			%% The second chunk may be either empty or 256 KiB. ar_poa:validate/1 does
+			%% the proper verification - here we simply protect against payload size abuse.
+			%% We are not strict about the first chunk here to simplify tests.
+			process_partial_solution_poa2(Solution, Ref);
+		_ ->
+			#partial_solution_response{ status = <<"rejected_bad_poa">> }
+	end.
+
+process_partial_solution_poa2(Solution, Ref) ->
```
JamesPiechota, line 277: This is just part of the poa2 validation, right? More happens down under `process_partial_solution_poa`?

What is this actually checking? that if recall_byte2 is not set poa2 is also not set? Can we get rid o this and do that check in`process_partial_solution_poa2`? Or maybe name it something different so it's clear that it's not actually validating poa2?

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+					#partial_solution_response{ status = <<"rejected_bad_poa">> }
+			end;
+		_ ->
+			process_partial_solution_partition_number(Solution, Ref)
+	end.
+
+process_partial_solution_partition_number(Solution, Ref) ->
+	PartitionNumber = Solution#mining_solution.partition_number,
+	PartitionUpperBound = Solution#mining_solution.partition_upper_bound,
+	Max = max(0, PartitionUpperBound div ?PARTITION_SIZE - 1),
```
JamesPiechota, line 297:
```suggestion
	Max = ?MAX_PARTITION_NUMBER(PartitionUpperBound),
```

apps/arweave/src/ar_pool.erl
JamesPiechota: Hm, I see it also mirrors and duplicates a lot of checks in `ar_block_pre_validator`... any chance we can deduplicate those two code paths? If you want to punt all these refactoring changes to me, I can take a look

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+		mining_address = MiningAddress,
+		preimage = Preimage,
+		solution_hash = SolutionH
+	} = Solution,
+	H0 = ar_block:compute_h0(NonceLimiterOutput, PartitionNumber, Seed, MiningAddress),
+	case ar_block:compute_solution_h(H0, Preimage) of
+		SolutionH ->
+			process_partial_solution_pow(Solution, Ref, H0);
+		_ ->
+			#partial_solution_response{ status = <<"rejected_wrong_hash">> }
```
JamesPiechota, line 328: Will this branch ever be triggered? It looks like `compute_solution_h` just does:
```
crypto:hash(sha256, << H0:32/binary, Preimage/binary >>)
```

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+process_partial_solution_pow(Solution, Ref, H0) ->
+	#mining_solution{
+		nonce = Nonce,
+		poa1 = #poa{ chunk = Chunk1 },
+		solution_hash = SolutionH,
+		preimage = Preimage,
+		recall_byte2 = RecallByte2,
+		poa2 = #poa{ chunk = Chunk2 }
+	} = Solution,
+	{H1, Preimage1} = ar_block:compute_h1(H0, Nonce, Chunk1),
```
JamesPiechota, line 340: We're not fully checking the recall bytes here, right? I guess that happens earlier? I see we do assert that RecallByte2 is undefined if this is an H1 solution, but we don't do any RecallByte2 or RecallByte checking when validating the H2 solution.

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+					PoACache = {{BlockStart1, RecallByte1, TXRoot1, BlockSize1,
+							{spora_2_6, MiningAddress}}, ChunkID},
+					process_partial_solution_difficulty(Solution, Ref, PoACache, PoA2Cache)
+			end
+	end.
+
+process_partial_solution_difficulty(Solution, Ref, PoACache, PoA2Cache) ->
+	#mining_solution{ solution_hash = SolutionH } = Solution,
+	Diff = ar_node:get_current_diff(),
+	case binary:decode_unsigned(SolutionH, big) > Diff of
```
JamesPiechota, line 422: This checks the network difficulty, right? Where do we check the partial difficulty? (did I miss the check above somewhere?)

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+		start_interval_number = StartIntervalNumber,
+		next_vdf_difficulty = NextVDFDifficulty,
+		nonce_limiter_output = Output,
+		seed = Seed,
+		partition_upper_bound = PartitionUpperBound
+	} = Solution,
+	case ar_nonce_limiter:get_step_checkpoints_seed_upper_bound(StepNumber, NextSeed,
+			StartIntervalNumber, NextVDFDifficulty) of
+		not_found ->
+			#partial_solution_response{ status = <<"rejected_vdf_not_found">> };
```
JamesPiechota, line 442: In this case could we respond `accepted`? Like is this a valid partial solution?

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+				Solution#mining_solution{
+					last_step_checkpoints = LastStepCheckpoints,
+					%% ar_node_worker will fetch the required steps based on the prev block.
+					steps = not_found
+				},
+			ar_events:send(miner, {found_solution, {pool, Ref},
+					Solution2, PoACache, PoA2Cache}),
+			noreply;
+		_ ->
+			#partial_solution_response{ status = <<"rejected_bad_vdf">> }
```
JamesPiechota, line 454: Ditto above: isn't this a valid partial solution?

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,764 @@
...
+	SessionKey = {NextSeed, IntervalNumber, NextVDFDifficulty},
+	SessionKeys = State#state.session_keys,
+	SessionKeys2 = [SessionKey | SessionKeys],
+	JobList = [{Job#job.output, Job#job.global_step_number,
+			Job#job.partition_upper_bound, Seed, Diff} || Job <- VDF],
+	PrevJobList = maps:get(SessionKey, State#state.jobs_by_session_key, []),
+	JobList2 = JobList ++ PrevJobList,
+	JobsBySessionKey = maps:put(SessionKey, JobList2, State#state.jobs_by_session_key),
+	{SessionKeys3, JobsBySessionKey2} =
+		case length(SessionKeys2) == 3 of
```
JamesPiechota, line 119: I haven't traced through all the flows, is there any chance that SessionKeys jumps to length of 4 between calls to cache_jobs? (thereby skipping the remove step forever)

apps/arweave/src/ar_pool_worker.erl
```diff
@@ -0,0 +1,90 @@
+-module(ar_pool_worker).
```
JamesPiechota, line 1: What about renaming this `ar_pool_client`? And maybe `ar_pool` -> `ar_pool_server`?

elsewhere we use the `x` and `x_worker` to describe a worker that handles tasks for `x`

apps/arweave/src/ar_pool_worker.erl
```diff
@@ -0,0 +1,90 @@
...
+	?LOG_WARNING([{event, unhandled_call}, {module, ?MODULE}, {request, Request}]),
+	{reply, ok, State}.
+
+handle_cast(fetch_jobs, State) ->
+	PrevOutput = ar_pool:get_latest_output(),
+	{ok, Config} = application:get_env(arweave, config),
+	{Peer, Pool} =
+		case {Config#config.coordinated_mining, Config#config.cm_exit_peer} of
+			{true, not_set} ->
+				{Config#config.pool_host, true};
```
JamesPiechota, line 47:
```suggestion
				%% We are a CM Exit Node
				{Config#config.pool_host, true};
```

apps/arweave/src/ar_pool_worker.erl
```diff
@@ -0,0 +1,90 @@
...
+			{true, ExitPeer} ->
+				{ExitPeer, false};
```
JamesPiechota, line 49:
```suggestion
				%% We are a CM Miner
				{ExitPeer, false};
```

apps/arweave/src/ar_pool_worker.erl
```diff
@@ -0,0 +1,90 @@
...
+			_ ->
+				{Config#config.pool_host, true}
```
JamesPiechota, line 51:
```suggestion
				%% We are a non-CM Miner
				{Config#config.pool_host, true}
```

apps/arweave/src/ar_pool_worker.erl
```diff
@@ -0,0 +1,90 @@
...
+	end,
+	{ok, #state{}}.
+
+handle_call(Request, _From, State) ->
+	?LOG_WARNING([{event, unhandled_call}, {module, ?MODULE}, {request, Request}]),
+	{reply, ok, State}.
+
+handle_cast(fetch_jobs, State) ->
+	PrevOutput = ar_pool:get_latest_output(),
+	{ok, Config} = application:get_env(arweave, config),
```
JamesPiechota, line 43: Should we also check that mining is enabled?

apps/arweave/src/ar_pool_worker.erl
```diff
@@ -0,0 +1,90 @@
...
+emit_pool_jobs(Jobs) ->
+	SessionKey = {Jobs#jobs.next_seed, Jobs#jobs.interval_number,
+			Jobs#jobs.next_vdf_difficulty},
+	emit_pool_jobs(Jobs#jobs.vdf, SessionKey, Jobs#jobs.diff).
+
+emit_pool_jobs([], _SessionKey, _PartialDiff) ->
+	ok;
+emit_pool_jobs([{O, SN, U} | Steps], SessionKey, PartialDiff) ->
+	Args = {SessionKey, SN, O, U, PartialDiff},
+	ar_events:send(pool, {job, Args}),
```
JamesPiechota, line 89: Right now it looks like only ar_mining_server subscribes to the `pool` event - if that's the case I'd recommend doing a direct 1:1 cast. There've been at least 2-3 OOM and performance issues directly related to inefficiencies in the `ar_events` broadcast code.

apps/arweave/include/ar_config.hrl
ldmberman: The difference is nodes will connect with the pool over https.

apps/arweave/src/ar_http_iface_middleware.erl
ldmberman: The motivation is to save traffic.

apps/arweave/src/ar_nonce_limiter.erl
ldmberman: No, we always use "Seed" to determine recall ranges.

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: My soft opinion is that this *may* fall under premature optimization in that we're adding complexity to the code to optimize for a scenario that may not be an issue. 

i.e. We're only optimizing the case where many small miners have coordinated into a CM cluster, but we aren't optimizing the case where many independent small miners join the pool directly. It may turn out that neither scenario causes enough traffic to be an issue, or it may turn out that the CM case rarely happens - in both those scenarios the optimization won't help us at all, but we're still left with a lot of code that needs to duplicately handle communicating with an Exit Node as well as communicating directly with the Pool. The cost there is both in terms of increased code complexity, but also the increased surface area to be tested and retested.

I personally would prefer to start simple and have all nodes communicate directly with the pool and only add in "Exit Node as Proxy" if we find a material bottleneck in practice.

apps/arweave/include/ar_config.hrl
JamesPiechota: Oh awesome.

apps/arweave/src/ar_mining_server.erl
```diff
@@ -564,11 +557,15 @@ post_solution(error, _State) ->
 post_solution(Solution, State) ->
 	{ok, Config} = application:get_env(arweave, config),
 	post_solution(Config#config.cm_exit_peer, Solution, State).
+
+post_solution(not_set, Solution, #state{ is_pool_client = true }) ->
+	ar_pool:post_partial_solution(Solution);
```
JamesPiechota, line 562:
```suggestion
	%% When posting a partial solution the pool client will skip many of the validation steps
	%% that are normally performed before sharing a solution
	ar_pool:post_partial_solution(Solution);
```

apps/arweave/src/ar_mining_server.erl
```diff
@@ -442,8 +423,13 @@ prepare_solution(Candidate, State) ->
 		start_interval_number = StartIntervalNumber,
 		step_number = StepNumber
 	},
-	prepare_solution(last_step_checkpoints, Candidate, Solution).
-	
+	case IsPoolClient of
```
JamesPiechota, line 426:
```suggestion
	%% A pool client does not validate VDF before sharing a solutino
	case IsPoolClient of
```

apps/arweave/src/ar.erl
JamesPiechota: Got it. And I see now that you've updated the ar_mining_server to skip all the "validation" steps prior to posting a partial solution. I've added a couple of suggested comments to call this out in the code.

apps/arweave/src/ar_node_worker.erl
JamesPiechota: yikes of course. Sorry I missed that!

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,766 @@
...
+	SessionKeys2 = [SessionKey | SessionKeys],
+	JobList = [{Job#job.output, Job#job.global_step_number,
+			Job#job.partition_upper_bound, Seed, Diff} || Job <- JobList],
+	PrevJobList = maps:get(SessionKey, State#state.jobs_by_session_key, []),
+	JobList2 = JobList ++ PrevJobList,
+	JobsBySessionKey = maps:put(SessionKey, JobList2, State#state.jobs_by_session_key),
+	{SessionKeys3, JobsBySessionKey2} =
+		case length(SessionKeys2) == 3 of
+			true ->
+				{RemoveKey, SessionKeys4} = lists:last(SessionKeys2),
```
vird, line 122: `lists:last(SessionKeys2)`
returns
```
{<<36,43,54,48,135,161,52,89,2,146,163,253,208,50,149,139,196,13,  
             112,248,56,43,31,27,139,165,1,223,112,221,211,39,107,168,244,74,  
             72,151,91,107,160,237,78,204,32,214,223,153>>,  
           24923,646784}
```
So
```
{RemoveKey, SessionKeys4} =
```
causes badmatch

```
RemoveKey = lists:last(SessionKeys2),
SessionKeys4 = lists:sublist(SessionKeys2, length(SessionKeys2) - 1),
```
fixed version, but ugly

apps/arweave/src/ar_pool.erl
```diff
@@ -0,0 +1,766 @@
...
+			interval_number = IntervalNumber,
+			next_vdf_difficulty = NextVDFDifficulty } = Jobs,
+	SessionKey = {NextSeed, IntervalNumber, NextVDFDifficulty},
+	SessionKeys = State#state.session_keys,
+	SessionKeys2 = [SessionKey | SessionKeys],
+	JobList = [{Job#job.output, Job#job.global_step_number,
+			Job#job.partition_upper_bound, Seed, Diff} || Job <- JobList],
+	PrevJobList = maps:get(SessionKey, State#state.jobs_by_session_key, []),
+	JobList2 = JobList ++ PrevJobList,
+	JobsBySessionKey = maps:put(SessionKey, JobList2, State#state.jobs_by_session_key),
```
vird, line 118: Accidental `JobList` reuse (left and right side) causes badmatch. JobList on left side should be renamed
```
JobList2 = [{Job#job.output, Job#job.global_step_number,
			Job#job.partition_upper_bound, Seed, Diff} || Job <- JobList],

PrevJobList = maps:get(SessionKey, State#state.jobs_by_session_key, []),
JobList3 = JobList2 ++ PrevJobList,
JobsBySessionKey = maps:put(SessionKey, JobList3, State#state.jobs_by_session_key),
```

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -504,6 +507,66 @@ handle(<<"POST">>, [<<"block2">>], Req, Pid) ->
...
+%%     [
+%%       {"output": "...", "global_step_number": "...", "partition_upper_bound": "..."},
+%%       ...
+%%     ],
+%%   "partial_diff": "...",
+%%   "next_seed": "...",
+%%   "interval_number": "...",
+%%   "next_vdf_difficulty": "..."
+%% }
+handle(<<"GET">>, [<<"jobs">>, EncodedPrevOutput], Req, _Pid) ->
```
vird, line 557: There is no /jobs endpoint (with no output)
This causes error on client

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -630,6 +630,74 @@ cm_publish_send(Peer, Solution) ->
...
+			false ->
+				{PeerOrURL, cm_p2p_headers(), ""};
+			true ->
+				#{ host := Host, port := Port, path := P } = uri_string:parse(PeerOrURL),
+				{{binary_to_list(Host), Port}, pool_client_headers(), binary_to_list(P)}
+		end,
+	handle_get_jobs_response(ar_http:req(#{
+		peer => Peer,
+		method => get,
+		path => BasePath ++ "/jobs/" ++ binary_to_list(ar_util:encode(PrevOutput)),
```
vird, line 646: PrevOutput "" effectively makes call to /jobs, and that endpoint is missing

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2509,6 +2572,118 @@ post_block(enqueue_block, {B, Peer}, Req, ReceiveTimestamp) ->
...
+					false;
+				_ ->
+					{true, S}
+			end
+		end,
+		200,
+		(?GET_JOBS_TIMEOUT_S) * 1000
+	),
+	Steps = case Result of false -> []; {true, S} -> S end,
+	{NextSeed, IntervalNumber, NextVDFDiff} = ar_nonce_limiter:session_key(Info),
```
vird, line 2675: Missing export in ar_nonce_limiter

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2509,6 +2572,118 @@ post_block(enqueue_block, {B, Peer}, Req, ReceiveTimestamp) ->
...
+				[] ->
+					false;
+				_ ->
+					{true, S}
+			end
+		end,
+		200,
+		(?GET_JOBS_TIMEOUT_S) * 1000
+	),
+	Steps = case Result of false -> []; {true, S} -> S end,
```
vird, line 2674: Does not handle error case (in case of timeout)

apps/arweave/src/ar_pool_worker.erl
```diff
@@ -0,0 +1,90 @@
...
+%%%===================================================================
+
+emit_pool_jobs(Jobs) ->
+	SessionKey = {Jobs#jobs.next_seed, Jobs#jobs.interval_number,
+			Jobs#jobs.next_vdf_difficulty},
+	emit_pool_jobs(Jobs#jobs.jobs, SessionKey, Jobs#jobs.partial_diff).
+
+emit_pool_jobs([], _SessionKey, _PartialDiff) ->
+	ok;
+emit_pool_jobs([{O, SN, U} | Steps], SessionKey, PartialDiff) ->
```
vird, line 87:
```
emit_pool_jobs([Job | Steps], SessionKey, PartialDiff) ->
  #job{ output = O, global_step_number = SN,
    partition_upper_bound = U } = Job,
```

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -630,6 +630,74 @@ cm_publish_send(Peer, Solution) ->
...
+
+%% @doc Post the partial solution to the pool.
+post_partial_solution(PeerOrURL, Solution, IsPool) ->
+	{Peer, Headers, BasePath} =
+		case IsPool of
+			false ->
+				{PeerOrURL, cm_p2p_headers(), ""};
+			true ->
+				#{ host := Host, port := Port, path := P } = uri_string:parse(PeerOrURL),
+				{{binary_to_list(Host), Port}, pool_client_headers(), P}
```
vird, line 661: Missing binary_to_list(P) as in get_jobs

apps/arweave/src/ar_http_iface_client.erl
vird: But actually binary_to_list is not correct
There should be pattern matching for path
Because if http://host:123, then path is <<>>
But if http://host:123/ then path is "/"
But binary_to_list("/") causes bad argument

apps/arweave/src/ar_http_iface_client.erl
vird: But actually binary_to_list is not correct
There should be pattern matching for path
Because if http://host:123/, then path is <<>>
But if http://host:123/ then path is "/"
But binary_to_list("/") causes bad argument

apps/arweave/src/ar_http_iface_middleware.erl
vird: There could be a case when workers are behind firewall, and cm host:port is actually port-forwarded from other network. In this case worker will have job, but can't post solution

This is kinda edge case, but for China it's kinda possible. Otherwise you need setup VPN for whole cluster

apps/arweave/src/ar_http_iface_client.erl
vird: Also i've added content-type, because proxy refused to parse POST json without this header
```
Headers2 = pool_client_headers(),
Headers3 = add_header(<<"content-type">>, "application/json", Headers2),
{{binary_to_list(Host), Port}, Headers3, binary_to_list(P)}
```

vird: upd
```
Headers3 = add_header(<<"content-type">>, <<"application/json">>, Headers2),
```

apps/arweave/src/ar_nonce_limiter.erl
ldmberman: `get_step_checkpoints_seed_upper_bound` fetches the information about a particular step while `get_triplets` returns the latest range.

What do you think about `get_step_info(StepN, SessionKey, [last_step_checkpoints, seed, partition_upper_bound])` where the last argument is a list of quantities that we'd like to receive?

apps/arweave/src/ar_http_iface_middleware.erl
ldmberman: A "normal" solution may not satisfy the difficulty threshold either (e.g., because the difficulty has just jumped up) so the main difference is in how we use them - partial solutions serve to prove the ongoing mining activity. Yes, the set of all solutions of the blocks produced by the pool come from "partial" solutions technically (they go through the same pipeline).

apps/arweave/src/ar_block_pre_validator.erl
ldmberman: The first 4 elements did not change so `insert_element(5, ...` results in the same order of the elements.

apps/arweave/src/ar.erl
JamesPiechota: Works for me! Mostly I was just looking to use one name - and I agree, `pool_host` sounds like the better of the two!

apps/arweave/src/ar_block_pre_validator.erl
JamesPiechota: got it. I guess this is why I'm not a fan of erlang tuple manipulation it always feels incredibly brittle and obfuscated to me. But it is what it is. Comment resolved!

apps/arweave/src/ar_node_worker.erl
JamesPiechota: Could we use `partial` instead? So the list of different solution states would be:
`rejected`: invalid solution for some reason
`stale`: valid solution that is now too old
`partial`: valid solution that does not pass the network difficulty check
`accepted`: valid solution that passes the network difficulty check and is not too old

apps/arweave/src/ar.erl
ldmberman: I was defensive here blocking the use-case which I find unlikely to be useful. However, I cannot see why it won't work at the moment so we can enable it if you prefer.

JamesPiechota: Got it. Nope I like your approach of preventing unlikely/unuseful use cases to limit complexity! My comment was more because I didn't understand why it cannot participate technically - but since that's not the case, no need to elaborate further. Your code as is works for me!

apps/arweave/src/ar_block_pre_validator.erl
ldmberman: I do not think we can draw a separation boundary on the module level. Yes, the code processing new blocks does not need to be aware of the old forks once the latest fork height drops below the checkpoint block (in practice, we usually keep it for some time before removing). `ar_block_pre_validator` is only concerned with the recent blocks, yes. The code that computes block hashes (`ar_block:indep_hash`) must be fork-aware as this is how the node verifies the integrity of the old blocks it downloads and asserts they belong to the current chain.

apps/arweave/src/ar_http_iface_client.erl
ldmberman: It is a boolean that is true when the request is to be made to the pool. The name is confusing - should I rename it to `IsPoolRequest` or do you prefer to approach it differently?

apps/arweave/src/ar_http_iface_client.erl
ldmberman: We plan to set up https on the proxy. The support for https requests on the client side is in place.

apps/arweave/src/ar_http_iface_client.erl
JamesPiechota: Hm... I guess this is similar to the other discussion... my preference would be to always query jobs from the Pool and not go through the CM Exit Node. The idea would be that it will simplify the code by eliminating all these dual codepaths - if after launch we find that the network traffic of the pool is a problem we can tackle that as a specific and holistic optimization (and the solution we find at that time could be something completely different and might not involve using the Exit node a proxy for the pool)

apps/arweave/src/ar_block_pre_validator.erl
JamesPiechota: Got it. Okay, so no rule of thumb I can follow. I'll just be careful with removing fork height code!

apps/arweave/src/ar_http_iface_client.erl
ldmberman: Yep, I see your point. However, I was wrong - it was not the traffic that was our main concern; see @vird's comment [here](https://github.com/ArweaveTeam/arweave/pull/515/files#r1448529287).

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: Okay I think I understand. Is this accurate:
1. When a CM cluster joins a pool, the Pool only interacts with the Exit Node. So from the perspective of the pool, only a single miner has joined the pool. The pool doesn't know (and doesn't need to know) how many actual miners are running behind the exit node.

2. By the same token the CM miners themselves never interact with the pool host, they only interact with the Exit node so from their perspective, the Exit node is the pool host (they get jobs from the exit node and share partial solutions with the exit node)

apps/arweave/src/ar_pool_worker.erl
ldmberman: Hm, good point. I would go further and require the flag to be set at the configuration stage since it does not make sense to be a pool client and not mine and since we have a mine flag we can as well force the user to set them both to be very explicit about what's going on, what do you think?

apps/arweave/src/ar_http_iface_client.erl
JamesPiechota: Got it. I've added a comment to clarify my understanding over there. It does sound like the dual path system is required to support CM and Pools. With that in mind I do think we'll want to give some thought to how to remove duplication and complexity... but I think that can wait until this first pass is done.

I've added an issue to flag this for later: https://github.com/ArweaveTeam/arweave-dev/issues/354

I realize adding an issue is the first step towards never actually cleaning the code up, but I also agree that it's no longer a priority (I had originally hoped we could remove the entire codepath, but I now see why it's necessary). I'll do my best to take a look sometime after you merge but before we release. 🤞

JamesPiechota: As for the variable name here - how about reversing it and calling it `GetJobsFromExitNode` or something like that?

apps/arweave/src/ar_http_iface_client.erl
JamesPiechota: Got it. 👍

apps/arweave/src/ar_pool_worker.erl
JamesPiechota: sounds good!

apps/arweave/src/ar_pool_worker.erl
JamesPiechota: I like it! Prevent all non-useful configurations 💯

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: Logged a backlog issue to see if there's anyway to simplify the duplicate codepaths: https://github.com/ArweaveTeam/arweave-dev/issues/354

but not a priority for this PR.

apps/arweave/src/ar_nonce_limiter.erl
JamesPiechota: I like it. It unifies the naming and makes clear what the function is returning.

ldmberman: On second thought, get_step_info introduces a bit of unpleasant complexity, I broke down the function into three instead - https://github.com/ArweaveTeam/arweave/pull/515/commits/6dffaa2ca2204bcc0392b42b5b39959c162ccc98

apps/arweave/src/ar_pool.erl
ldmberman: I renamed it to process_partial_solution_poa2_size

apps/arweave/src/ar_pool.erl
ldmberman: I'd say they have structural similarities but are semantically different so different names and some code duplication are justified - `prepare_solution` builds one whereas `process_partial_solution` processes an external solution.

I prefer different functions to atom matching because the linter can catch mistakes (by highlighting unused or undefined functions). I had a temptation to refactor prepare_solution but was discouraged by the way `prepare_solution(poa1` and `prepare_solution(poa2` are intertwined [here](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_mining_server.erl#L567).

JamesPiechota: Got it. Is `process_partial_solution` basically validating the partial solution? And in that way perhaps is more semantically similar to the code in `ar_block_pre_validator` and `ar_node_utils`? If so I can add a note to to https://github.com/ArweaveTeam/arweave-dev/issues/354 to look into deduplicating common code there.

Ultimately I'm selfishly trying to reduce work for myself in the future - I've spent a lot of time tracing through the different validation paths, sometimes picking the wrong set of validation code to look at, etc... If I can take the 4 different validation paths we have now (`ar_pool`, `ar_block_pre_validator`, `ar_node_utils`, `ar_mining_server`) and unify at least parts of them, it'll reduce the time I spend in the future debugging and reading code.

Also: I added a note to the refactoring issue [here](https://github.com/ArweaveTeam/arweave-dev/issues/354) to also remove the atom matching from `prepare_solution`

apps/arweave/src/ar_pool.erl
ldmberman: Added a clarifying comment

apps/arweave/src/ar_pool.erl
ldmberman: We validate RecallByte2 further in process_partial_solution_poa

apps/arweave/src/ar_pool.erl
ldmberman: If it's not due to some server bug, it should be an outdated job the proxy let through - I think, it's helpful to let the proxy know what happens (it may use it to clean up the job cache) and decide whether they want to reward the miner for it.

apps/arweave/src/ar_pool.erl
ldmberman: Added a clarifying comment

apps/arweave/src/ar_pool.erl
ldmberman: The cache is maintained to have at most two sessions at all times.

apps/arweave/src/ar_http_iface_middleware.erl
ldmberman: Added ar_coordination:is_exit_peer

apps/arweave/src/ar_mining_server.erl
```diff
@@ -594,7 +592,17 @@ post_solution(not_set, Solution, State) ->
 			ar:console("WARNING: the solution we found is invalid. Check logs for more "
 					"details~n");
 		{true, PoACache, PoA2Cache} ->
-			ar_events:send(miner, {found_solution, Solution, PoACache, PoA2Cache})
+			ar_events:send(miner, {found_solution, miner, Solution, PoACache, PoA2Cache})
+	end;
+post_solution(ExitPeer, Solution, #state{ is_pool_client = true }) ->
+	case ar_http_iface_client:post_partial_solution(ExitPeer, Solution, false) of
```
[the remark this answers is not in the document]
ldmberman: Posting to pool requires an API key. The current implementation assumes only the exit node is configured with one.

apps/arweave/src/ar_nonce_limiter.erl
ldmberman: We only consider one VDF session here and every session has one reset point (the "first" one).

apps/arweave/src/ar_mining_server.erl
ldmberman: `cm_publish_send` posts to `/coordinated_mining/publish`. I would not mix coordinated mining interactions with the pool interactions cause they are different use-cases and have slightly different trust requirements. An alternative might be to always send partial solutions directly to the pool from every node, but it complicates configuration a bit.

apps/arweave/src/ar_serialize.erl
ldmberman: Done

apps/arweave/src/ar.erl
```diff
@@ -541,6 +547,14 @@ parse_cli_args(["cm_exit_peer", Peer | Rest], C) ->
 			io:format("Peer ~p is invalid.~n", [Peer]),
 			parse_cli_args(Rest, C)
 	end;
+parse_cli_args(["pool_server" | Rest], C) ->
+	parse_cli_args(Rest, C#config{ pool_server = true });
+parse_cli_args(["pool_client" | Rest], C) ->
```
JamesPiechota, line 552: Is pool_client only set when the current node is part of a CM set and is uing the CM exit node to communicate with the pool?

apps/arweave/src/ar.erl
```diff
@@ -541,6 +547,14 @@ parse_cli_args(["cm_exit_peer", Peer | Rest], C) ->
...
+	parse_cli_args(Rest, C#config{ pool_client = true });
+parse_cli_args(["pool_api_key", Key | Rest], C) ->
+	parse_cli_args(Rest, C#config{ pool_api_key = list_to_binary(Key) });
+parse_cli_args(["pool_host", Host | Rest], C) ->
```
JamesPiechota, line 556: Earlier I don't think we had both pool_server and pool_host, so renaming pool_proxy to pool_host IMO made sense. Now, though, since server and host are often used interchangeably in the industry the two variables are confusing. Is pool_server the node arweave node backing the pool_host? If so what about "pool_arweave_node" or something? Or rename pool_host to pool_gateway?

apps/arweave/src/ar.erl
JamesPiechota: ah wait, I see these are true/false values and not IP addresses. Got it.

apps/arweave/src/ar.erl
JamesPiechota: Oh wait reading the ar.erl description what about:
is_pool_client (true/false)
is_pool_server (true/false)
pool_server_address (IP address)?

apps/arweave/src/ar.erl
JamesPiechota, line 552: In a CM cluster would both the CM Exit Node and CM Miner have pool_client = true?

ldmberman: Yep, is_pool_client implies we need to fetch jobs from the pool

apps/arweave/src/ar.erl
ldmberman: Renamed

JamesPiechota (approved):

PR #518, opened 2024-01-06
Update README.md

joaomoa:
Erlang is also needed as a dependency

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ For more information, refer to the [mining guide](https://docs.arweave.org/info/
 To install the dependencies on Ubuntu 22 (recommended), run:
 
 ```sh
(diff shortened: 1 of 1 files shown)


vird:
Recipe in MR is not cross-platform
For ubuntu 22 it seems should work. But I recommend more robust https://github.com/virdpool/miner/blob/master/install_ubuntu_22.04.sh#L23-L29

For ubuntu 20 provided recipe will be non functional (not proper version will be installed). But this ugly recipe works
https://github.com/virdpool/miner/blob/master/install_ubuntu_20.04.sh#L22-L75

hlolli:
a stable version of erlang isn't included by default in aptitute mirrors of Debian/Ubuntu distros. If this PR is to be merged, a better README needs to be added in how to add some of the 3rd party mirrors that do provide a decent distribution of erlang+otp 24/25 or later.

PR #519, opened 2024-01-08
Fix case statement

joaomoa:
The true case is ok. False is when we should have a warning

### DIFF
--- a/apps/arweave/src/ar_header_sync.erl
+++ b/apps/arweave/src/ar_header_sync.erl
@@ -280,6 +280,8 @@ handle_info({event, disksup, {remaining_disk_space, "default", true, _Percentage
 		true ->
 			case State#state.is_disk_space_sufficient of
 				true ->
+					ok
+				false ->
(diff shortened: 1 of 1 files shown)


apps/arweave/src/ar_header_sync.erl
```diff
@@ -280,6 +280,8 @@ handle_info({event, disksup, {remaining_disk_space, "default", true, _Percentage
 		true ->
 			case State#state.is_disk_space_sufficient of
 				true ->
+					ok
+				false ->
```
JamesPiechota, line 284: I believe in this case `is_disk_space_sufficient` is being used as a toggle to limit the number of time that the warning is printed. It should only be printed once, the first time that `Bytes < DiskPoolSize + DiskCacheSize + BufferSize div 2`. Once it prints it once it sets `is_disk_space_sufficient` false and won't print again until `is_disk_space_sufficient` is set to `true` and then `false`

Current logic is correct I believe.

PR #522, opened 2024-01-31
Mining Performance Changes

JamesPiechota:
This PR makes several changes to improve mining performance - particularly at higher partition counts (10+).

The 3 biggest changes:

1. 
Finish work done earlier to distribute the mining workload across worker processes that scale with the number of partitions mined. Previously the task management was centralized even if the hashing and io was not. This task management became a bottleneck starting at around 8 partitions. With the new distributed workload, the task management work is distributed across cores allowing it to scale so long as the CPU has sufficient cores or clockspeed to handle the number of partitions being mined.

2.
Manually trigger erlang's garbage collector periodically for each of the mining processes. Several of the processes involved in mining "touch" chunks. The way Erlang's memory management works each time a process touches a chunk it gets a reference to the full 100MiB recall range.

Also the default garbage collection frequency is based on the amount of activity a given process does. So for some processes the default GC may happen infrequently. This can lead to the 100MiB recall ranges staying in memory long enough to trigger memory issues.

By manually GC'ing periodically we ensure memory is released regularly.

3.
Tune erlang's memory management parameters to better deal with Arweave's unique memory access pattern (namely: Arweave allocates a lot of 100MiB memory blocks)


apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -1090,17 +1093,30 @@ apply_external_update2(Update, State) ->
...
 		{step_number, Session#vdf_session.step_number},
 		{length, length(Steps)}]),
-	State2 = cache_session(State, SessionKey, CurrentSessionKey, Session),
+
+	{_, SessionInterval, _} = SessionKey,
+	CurrentSessionInterval = case CurrentSessionKey of
+		{_, Interval, _} -> Interval;
+		_ -> -1
+	end,
+	State2 = case SessionInterval > CurrentSessionInterval of
```
ldmberman, line 1110: The current design is such that assigning the current session has to be consensus-based only (i.e. happen only in apply_tip). What is the motivation for this change?

apps/arweave/src/ar_packing_server.erl
```diff
@@ -528,18 +528,16 @@ calling_function([_, {_, _, _, _}|[{Module, Function, Arity, _}|_]]) ->
 calling_function(_) ->
     "unknown".
 
-record_packing_request(Type, Packing, From) when
-	is_atom(Type), is_atom(Packing), is_list(From) ->
+record_packing_request(Type, RequestedPacking, StoredPacking, From) ->
+	Type2 = case RequestedPacking == StoredPacking of
```
ldmberman, line 532: Sometimes StoredPacking passed is `unpacked`, sometimes - `unpack`, is it intentional? What do you think about only tracking the actual packings and unpackings (where randomx_encrypt/decrypt_chunk's are called)?

apps/arweave/src/ar_nonce_limiter.erl
JamesPiechota: Ah, got it. The issue this was try to solve was the following:
1. Sometimes the VDF server would update its current session before the VDF client
2. The client would receive those updates, but wouldn't mine against them since the session was neither the current, nor previous session (i.e. it treated the updates as stale updates)
3. This caused the miner to have periods of 0 hashrate

I can find another way to address the issue without breaking consensus!

apps/arweave/src/ar_packing_server.erl
JamesPiechota: Hm, probably not intentional. I'll review! And I agree tracking actual packing/unpacking is best - the old logic tracked everything without distinguishing whether an encrypt/decrypt actually happened, this "fix" intended to update the logic to flag when an encrypt/decrypt actually happened. I can review again and maybe simply remove the tracking for no-ops rather than tag them as no-ops.

JamesPiechota: Fixed!

apps/arweave/src/ar_nonce_limiter.erl
JamesPiechota: Fixed!

PR #528, opened 2024-03-07
Add support for in-place storage module repacking

ldmberman:


apps/arweave/src/ar.erl
```diff
@@ -84,12 +84,39 @@ show_help() ->
...
 					"syncing the weave data between the offsets 22 TB and 23 TB. Make sure "
 					"the corresponding disk contains some extra space for the proofs and "
-					"other metadata, about 10% of the configured size."},
+					"other metadata, about 10% of the configured size."
+
+					"You may repack a storage module in-place. To do that, specify "
+					"storage_module "
+					"{partition_number},{packing},repack_in_place,{target_packing}. "
+					"For example, if you want to repack a storage module "
+					"22,En2eqsVJARnTVOSh723PBXAKGmKgrGSjQ2YIGwE_ZRI to the new address "
```
JamesPiechota, line 98: Is this correct?

`storage_module 22,En2eqsVJARnTVOSh723PBXAKGmKgrGSjQ2YIGwE_ZRI,repack_in_place,Q5EfKawrRazp11HEDf_NJpxjYMV385j21nlQNjR8_pY`

will load data from 

`data_dir/storage_modules/storage_module_22_En2eqsVJARnTVOSh723PBXAKGmKgrGSjQ2YIGwE_ZRI`

repack it to `Q5EfKawrRazp11HEDf_NJpxjYMV385j21nlQNjR8_pY` and then write it back to `data_dir/storage_modules/storage_module_22_En2eqsVJARnTVOSh723PBXAKGmKgrGSjQ2YIGwE_ZRI`

Once once the repacking is complete the miner will need to rename `data_dir/storage_modules/storage_module_22_En2eqsVJARnTVOSh723PBXAKGmKgrGSjQ2YIGwE_ZRI` to `data_dir/storage_modules/storage_module_22_Q5EfKawrRazp11HEDf_NJpxjYMV385j21nlQNjR8_pY`

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -382,12 +413,12 @@ get_chunk_group_size() ->
 	{ok, Config} = application:get_env(arweave, config),
 	Config#config.chunk_storage_file_size.
 
-read_repack_cursor(StoreID) ->
-	Filepath = get_filepath("repack_cursor", StoreID),
+read_repack_cursor(StoreID, TargetPacking) ->
+	Filepath = get_filepath("repack_in_place_cursor", StoreID),
```
JamesPiechota, line 417: the repack cursor is still used when repacking across modules as well as repacking in place right? (i.e. I shouldn't read to much into the `"repack_in_place_cursor"` name)

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -236,9 +262,7 @@ init(StoreID) ->
 			?LOG_WARNING(WarningMessage);
 		false ->
 			ok
-	end,
-	{ok, #state{ file_index = FileIndex2, store_id = StoreID,
-			repack_cursor = read_repack_cursor(StoreID) }}.
```
JamesPiechota, line 241: Oh do we only load the cursor from disk now if we're repacking in place? Not also if we're repacking across storage modules?

apps/arweave/src/ar_chunk_storage.erl
JamesPiechota: Oh actually I should read into it. Looks like we're no longer loading the cursor from disk if we're repacking across storage_modules?

https://github.com/ArweaveTeam/arweave/pull/528/files#diff-70860f94e3504d0cdd89e58a65a090fadb6a75eabc58f0a89aeae829109e8dbbL241

ldmberman: The cross-module repacking does not use a cursor. This here is a part of the spora_2_5 repacking process that was turned off at some point.

apps/arweave/src/ar.erl
ldmberman: Yes

apps/arweave/src/ar_chunk_storage.erl
JamesPiechota: ahh, got it. I was wondering why I didn't notice it when I had started looking into the repack resume bug. Makes sense.

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -756,6 +801,8 @@ repack(Cursor, RightBound, Packing, StoreID) ->
 repack(Start, End, NextCursor, RightBound, Packing, StoreID) when Start >= End ->
 	repack(NextCursor, RightBound, Packing, StoreID);
 repack(Start, End, NextCursor, RightBound, RequiredPacking, StoreID) ->
+	{ok, Config} = application:get_env(arweave, config),
+	RepackIntervalSize = 262144 * Config#config.repack_batch_size,
```
JamesPiechota, line 805:
```suggestion
	RepackIntervalSize = ?DATA_CHUNK_SIZE * Config#config.repack_batch_size,
```

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -824,29 +873,42 @@ repack(Start, End, NextCursor, RightBound, RequiredPacking, StoreID) ->
 						ok;
 					(AbsoluteOffset, {_, TXRoot, _, _, _, ChunkSize}, ok) ->
 						PaddedOffset = ar_data_sync:get_chunk_padded_offset(AbsoluteOffset),
-						case ar_sync_record:is_recorded(PaddedOffset, ar_data_sync) of
-							{{true, RequiredPacking}, StoreID} ->
+						case ar_sync_record:is_recorded(PaddedOffset, ar_data_sync, StoreID) of
+							{true, RequiredPacking} ->
 								?LOG_WARNING([{event,
```
JamesPiechota, line 878: This shouldn't happen right? Since the cursor and get_next_synced_interval should prevent us from trying to repack a chunk that has already been repacked

apps/arweave/src/ar_config.erl
```diff
@@ -34,6 +34,16 @@ parse(Config) when is_binary(Config) ->
 parse_storage_module(IOList) ->
 	Bin = iolist_to_binary(IOList),
 	case binary:split(Bin, <<",">>, [global]) of
+		[PartitionNumberBin, PackingBin, <<"repack_in_place">>, ToPackingBin] ->
+			PartitionNumber = binary_to_integer(PartitionNumberBin),
+			true = PartitionNumber >= 0,
+			parse_storage_module(PartitionNumber, ?PARTITION_SIZE, PackingBin, ToPackingBin);
```
JamesPiechota, line 40: This means we only support repacking full 3.6TB storage modules in place, right?

JamesPiechota: oops, never mind. saw the bit below

apps/arweave/src/ar.erl
```diff
@@ -84,12 +84,39 @@ show_help() ->
...
 					"storage_module "
 					"22,1000000000000,En2eqsVJARnTVOSh723PBXAKGmKgrGSjQ2YIGwE_ZRI will be "
 					"syncing the weave data between the offsets 22 TB and 23 TB. Make sure "
 					"the corresponding disk contains some extra space for the proofs and "
-					"other metadata, about 10% of the configured size."},
+					"other metadata, about 10% of the configured size."
+
+					"You may repack a storage module in-place. To do that, specify "
+					"storage_module "
+					"{partition_number},{packing},repack_in_place,{target_packing}. "
```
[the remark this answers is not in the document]
JamesPiechota: Never mind! I saw the handling case below. It does work.

apps/arweave/src/ar_data_sync_sup.erl
```diff
@@ -43,12 +43,20 @@ init([]) ->
...
+			?CHILD_WITH_ARGS(ar_data_sync, worker, Name, [Name, {StoreID, none}])
 		end,
 		Config#config.storage_modules
 	),
-
 	DefaultStorageModuleWorker = ?CHILD_WITH_ARGS(ar_data_sync, worker,
-		ar_data_sync_default, [ar_data_sync_default, "default"]),
-	Children = SyncWorkers ++ StorageModuleWorkers ++ [DefaultStorageModuleWorker],
+		ar_data_sync_default, [ar_data_sync_default, {"default", none}]),
+	RepackInPlaceWorkers = lists:map(
```
JamesPiechota, line 52: Do we need to spawn ar_data_sync workers for repack_in_place storage modules? I suspect we don't want to sync new data for a storage module that's been repacked in place (right?) - but maybe ar_data_sync gets used for something else that is required while repacking in place?

JamesPiechota (approved):
LGTM!

apps/arweave/src/ar_chunk_storage.erl
ldmberman: The cursor is not strictly in sync with the process and serves mostly to save time after a restart. However, since we are only looking at the intervals packed with a different packing this here should not happen, yes.

apps/arweave/src/ar_data_sync_sup.erl
ldmberman: We use them to fetch chunk metadata to filter out small chunks [here](https://github.com/ArweaveTeam/arweave/blob/feature/in-place-sm-repacking/apps/arweave/src/ar_chunk_storage.erl#L847).

JamesPiechota: Got it. Thanks!

apps/arweave/src/ar_chunk_storage.erl
JamesPiechota: :+1: thanks.

PR #537, opened 2024-03-21
Attempt to recognize "seeding-only" chunks early

ldmberman:
Send a distinct reply to POST /chunk.

### DIFF
--- a/apps/arweave/src/ar_data_sync.erl
+++ b/apps/arweave/src/ar_data_sync.erl
@@ -135,7 +135,7 @@ add_chunk(DataRoot, DataPath, Chunk, Offset, TXSize) ->
 				case ar_kv:get(DiskPoolChunksIndex, DiskPoolChunkKey) of
 					{ok, _DiskPoolChunk} ->
 						%% The chunk is already in disk pool.
-						synced;
+						{synced_disk_pool, EndOffset2};
 					not_found ->
 						case DataRootOffsetReply of
 							not_found ->
@@ -165,6 +165,13 @@ add_chunk(DataRoot, DataPath, Chunk, Offset, TXSize) ->
 	case CheckSynced of
 		synced ->
 			ok;
+		{synced_disk_pool, EndOffset4} ->
+			case is_estimated_long_term_chunk(DataRootOffsetReply, EndOffset4) of
+				false ->
+					temporary;
+				true ->
+					ok
+			end;
 		{error, _} = Error4 ->
 			Error4;
 		{ok, {DataPathHash2, DiskPoolChunkKey2, {EndOffset3, PassesBase3, PassesStrict3,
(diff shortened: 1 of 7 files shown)


apps/arweave/src/ar_data_sync.erl
```diff
@@ -199,11 +206,44 @@ add_chunk(DataRoot, DataPath, Chunk, Offset, TXSize) ->
...
+	IsTipCovered = is_tip_covered(),
+	case DataRootOffsetReply of
+		not_found ->
+			%% A chunk from a pending transaction.
+			IsTipCovered;
+		{ok, {TXStartOffset, _}} ->
+			WeaveSize = ar_node:get_current_weave_size(),
+			Size = ar_node:get_recent_max_block_size(),
+			AbsoluteEndOffset = TXStartOffset + EndOffset,
+			case AbsoluteEndOffset > WeaveSize - Size * 4 of
```
JamesPiechota, line 231: Why do we use `Size * 4` here and `Size * 2` in `is_tip_covered`? Like if AbsoluteEndOffset is at `WeaveSize - Size * 3` and `is_tip_covered` is `true` - but we only cover `Weave - Size * 2`... would that cause a problem?

apps/arweave/src/ar_storage_module.erl
```diff
@@ -172,6 +175,17 @@ get_all(Start, End) ->
...
 
+%% @doc Return true if the given Offset belongs to at least one storage module.
+has_any(Offset) ->
+	{ok, Config} = application:get_env(arweave, config),
+	has_any(Offset, Config#config.storage_modules).
+
+%% @doc Return true if the given range is covered by the configured storage modules.
+has_range(Start, End) ->
+	{ok, Config} = application:get_env(arweave, config),
+	Intervals = get_unique_sorted_intervals(Config#config.storage_modules),
```
JamesPiechota, line 186: I'm definitely not suggesting we worry about optimizing anything without testing - but subjectively/qualitatively: why do we think that sorting all the intervals on each POST /chunk request won't materially impact performance?

Is it because we believe that almost all the time the number of intervals to be sorted, even for a large-ish miner, isn't that many (i.e. order of thousands rather than millions)?

apps/arweave/src/ar_data_sync.erl
ldmberman: Hm, good point, the offsets between `-Size*4` and `-Size*2` would be excluded for no good reason. [This](https://github.com/ArweaveTeam/arweave/pull/537/commits/deb70d9e6e446b97817f32b9483853e6487e0c4c) makes more sense. `4` is the constant that expresses the confidence a reorg won't change the offset whereas `2` expresses the expected conservative offset "drift" in the mempool.

apps/arweave/src/ar_storage_module.erl
ldmberman: Hm, that's about 10 extra ms, do you prefer to cache it?

apps/arweave/src/ar_data_sync.erl
JamesPiechota: looks good!

apps/arweave/src/ar_storage_module.erl
JamesPiechota: It looks like `has_range` is called from `is_offset_vicinity_covered` which is in turn called once or twice for every call to `add_chunk`. So we're estimating that this would add 10-20ms every time we write a chunk? Is that correct? Or is only called under certain scenarios when writing a chunk (I couldn't quite trace through all the branches)

If it's every time we write a chunk, it sounds like it could impact the syncing phase runtime. Maybe?

e.g. if we assume 14,400,000 chunks per partition, and 200 concurrent sync jobs. Then we're looking at an increase in time of (14,400,000 / 200) * 10-20ms = 720-1440seconds or 12-24 minutes.

Okay even in that case it actually doesn't sound super problematic. Syncing and packing a full partition on a Ryzen 9 will take 11 hours, so we're adding maybe 2-3%. It's not nothing, but not sure anyone will notice.

Assuming all the math above is correct - what do you think? How hard would it be to cache and is it worth the engineering time and code complexity to save 2-3% of the sync/pack time per partition?

ldmberman: There was an extra call (fixed) - so we invoke `is_offset_vicinity_covered` once per add_chunk. Also, forgot to mention - I assumed ~10k storage modules.

I've made an easy patch which [caches the list](https://github.com/ArweaveTeam/arweave/pull/537/commits/8ef211e2ba4058a9b2a2a996ddedabbde14f3d88). It's a bit of a compromise due to some memory-movement overhead - ideally, we create an ordered ets set and use it inside has_range.

JamesPiechota: oh wait it only adds 10ms for 10k storage modules? In that case no need to cache I don't think!

JamesPiechota: So at current size it's a fraction of an millisecond, right? I don't think that's something we need worry about then! By the time it takes a long time to search, I'm sure we'll have other, bigger bottlenecks :)

JamesPiechota: Does the cache ever need to be invalidated? Like if the weave grows?

ldmberman: The configuration (including storage modules) is static at the moment.

JamesPiechota (approved):
LGTM!

PR #538, opened 2024-03-25
Don't log a warning if we fail to push a VDF update

JamesPiechota:
unless it's an unexpected error


vird:
Aha, so we still log it but in debug logs. So title a bit misleading. LGTM

PR #542, opened 2024-03-28
Add GET /chunk_proof(2) endpoints for serving merkle proofs without t…

ldmberman:
…he chunks

### DIFF
--- a/apps/arweave/src/ar_data_sync.erl
+++ b/apps/arweave/src/ar_data_sync.erl
@@ -5,7 +5,8 @@
 -export([start_link/2, join/1, add_tip_block/2, add_block/2,
 		is_chunk_proof_ratio_attractive/3,
 		add_chunk/5, add_data_root_to_disk_pool/3, maybe_drop_data_root_from_disk_pool/3,
-		get_chunk/2, get_tx_data/1, get_tx_data/2, get_tx_offset/1, has_data_root/2,
+		get_chunk/2, get_chunk_proof/2, get_tx_data/1, get_tx_data/2,
+		get_tx_offset/1, has_data_root/2,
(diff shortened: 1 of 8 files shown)


apps/arweave/src/ar_data_sync.erl
```diff
@@ -1451,11 +1485,38 @@ get_chunk(Offset, SeekOffset, Pack, Packing, StoredPacking, StoreID) ->
...
 
-%% @doc Read the chunk as well as its metadata.
-%% Response is of the format:
+get_chunk_proof(Offset, SeekOffset, StoredPacking, StoreID) ->
+	case read_chunk_with_metadata(Offset, SeekOffset, StoredPacking, StoreID, false) of
+		{error, Reason} ->
+			{error, Reason};
+		{ok, DataPath, AbsoluteOffset, TXRoot, ChunkSize, TXPath} ->
+			CheckProof =
+				case validate_fetched_chunk({AbsoluteOffset, DataPath, TXPath, TXRoot,
```
JamesPiechota, line 1494: Why do we need to validate the chunk here? 

We're reading the chunk and proof from local storage, right? So are we just double checking that the proof is correct and hasn't been corrupted since we last wrote it to disk?

JamesPiechota (approved):

apps/arweave/src/ar_data_sync.erl
ldmberman: We do not strictly have to; yes, we check against corruption or validate the proofs that were copied over manually from somewhere.

JamesPiechota: Got it. Makes sense.

PR #543, opened 2024-03-29
Search peers for missing PoAs for mined blocks

ldmberman:


apps/arweave/src/ar_mining_server.erl
```diff
@@ -667,6 +688,40 @@ may_be_empty_poa(not_set) ->
...
+fetch_poa_from_peers(RecallByte) ->
+	Peers = ar_data_discovery:get_bucket_peers(RecallByte div ?NETWORK_DATA_BUCKET_SIZE),
+	From = self(),
+	lists:foreach(
+		fun(Peer) ->
+			spawn(
+				fun() ->
+					?LOG_INFO([{event, last_moment_proof_search},
+							{peer, ar_util:format_peer(Peer)}]),
+					case fetch_poa_from_peers(Peer, RecallByte) of
```
JamesPiechota, line 700:
```suggestion
					case fetch_poa_from_peer(Peer, RecallByte) of
```

apps/arweave/src/ar_mining_server.erl
```diff
@@ -667,6 +688,40 @@ may_be_empty_poa(not_set) ->
...
+		Peers
+	),
+	receive
+		{fetched_last_moment_proof, PoA} ->
+			PoA
+		after ?FETCH_POA_FROM_PEERS_TIMEOUT_MS ->
+			not_found
+	end.
+
+fetch_poa_from_peers(Peer, RecallByte) ->
```
JamesPiechota, line 717:
```suggestion
fetch_poa_from_peer(Peer, RecallByte) ->
```

apps/arweave/src/ar_mining_server.erl
```diff
@@ -576,18 +590,25 @@ prepare_solution(poa2, Candidate,
...
-					{nonce, Nonce},
-					{partition, PartitionNumber},
-					{mining_address, ar_util:safe_encode(MiningAddress)}]),
-			ar:console("WARNING: we have mined a block but failed to fetch "
-					"the chunk proofs required for publishing it. "
-					"Check logs for more details~n"),
-			error;
+			ar:console("WARNING: we have mined a block but did not find the PoA2 proofs "
+					"locally - searching the peers...~n"),
+			case fetch_poa_from_peers(RecallByte2) of
```
JamesPiechota, line 595: This part of ar_mining_worker would also benefit from this change, right? https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_mining_worker.erl#L437

I think that bit is invoked in coordinated mining

apps/arweave/src/ar_mining_server.erl
```diff
@@ -667,6 +688,40 @@ may_be_empty_poa(not_set) ->
...
+							ok;
+						PoA ->
+							From ! {fetched_last_moment_proof, PoA}
+					end
+				end)
+		end,
+		Peers
+	),
+	receive
+		{fetched_last_moment_proof, PoA} ->
```
JamesPiechota, line 711:
```suggestion
         %% The first spawned process to fetch a PoA from a peer will trigger this `receive` and allow
         %% `fetch_poa_from_peers` to exit. All other processes that complete later will trigger the
         %% `handle_info({fetched_last_moment_proof, _}, State) ->` above (which is a no-op)
		{fetched_last_moment_proof, PoA} ->
```

JamesPiechota: Is that comment a correct understanding of what's happening here? Are we sure the `receive` will be processed before `handle_info`? I haven't dug this deep into erlang's mailbox handling logic before

apps/arweave/src/ar_mining_server.erl
```diff
@@ -230,6 +233,9 @@ handle_info({garbage_collect, StartTime, GCResult}, State) ->
 	end,
 	{noreply, State};
 
+handle_info({fetched_last_moment_proof, _}, State) ->
```
JamesPiechota, line 236:
```suggestion
handle_info({fetched_last_moment_proof, _}, State) ->
    %% This is a no-op to handle "slow" response from peers that were queried by `fetch_poa_from_peers`
    %% Only the first peer to respond with a PoA will be handled, all other responses will fall through to here
    %% an be ignored.
```

apps/arweave/src/ar_mining_server.erl
ldmberman: Yep, receive will dequeue one message, the rest goes to handle_info

JamesPiechota:
LGTM!

PR #545, opened 2024-04-04
Improvement/block rebase

ldmberman:


apps/arweave/src/ar_block_cache.erl
```diff
@@ -21,10 +22,10 @@
...
 %% not_validated: block is not validated yet
 %% none: null status
--type validation_status() :: on_chain | validated | not_validated | none.
--type nonce_limiter_validation_status() :: nonce_limiter_validated | 
-	nonce_limiter_validation_scheduled | awaiting_nonce_limiter_validation.
--type block_status() :: validation_status() | {not_validated, nonce_limiter_validation_status()}.
+%-type validation_status() :: on_chain | validated | not_validated | none.
+%-type nonce_limiter_validation_status() :: nonce_limiter_validated |
+%	nonce_limiter_validation_scheduled | awaiting_nonce_limiter_validation.
+%-type block_status() :: validation_status() | {not_validated, nonce_limiter_validation_status()}.
```
JamesPiechota, line 28: you can just delete those if you want. I'd added them when I was trying to undesr the ar_block_cache for the checkpoint work. Probably not useful as a one-off

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1026,10 +740,10 @@ get_block_anchors_and_recent_txs_map(BlockTXPairs) ->
 apply_block(State) ->
 	case ar_block_cache:get_earliest_not_validated_from_longest_chain(block_cache) of
 		not_found ->
-			%% Nothing to do - we are at the longest known chain already.
-			{noreply, State};
+			may_be_rebase(State);
```
JamesPiechota, line 743:
```suggestion
			maybe_rebase(State);
```

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1046,6 +760,76 @@ apply_block(State) ->
...
+	{noreply, State};
+may_be_rebase(_B, [], State) ->
+	{noreply, State};
+may_be_rebase(#block{ cumulative_diff = CDiff } = B,
+		[#block{ cumulative_diff = CDiff } = Sib | Siblings], State) ->
+	#block{ nonce_limiter_info = Info } = B,
+	#block{ nonce_limiter_info = SibInfo } = Sib,
+	StepNumber = Info#nonce_limiter_info.global_step_number,
+	SibStepNumber = SibInfo#nonce_limiter_info.global_step_number,
+	case {StepNumber > SibStepNumber, Sib#block.reward_addr == B#block.reward_addr} of
```
JamesPiechota, line 790: The idea here is if the sibling block has a *lower* `global_step_number` we want to switch to the sibling branch and then rebuild a new block using teh current solution and apply that block onto the sibling?

How come we don't have to check cumulative diff? Are we guaranteed that 2 blocks at the same height have the same cumulative difficulty?

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1926,3 +1710,317 @@ dump_mempool(TXs, MempoolSize) ->
...
+					true ->
+						true
+				end
+		end,
+
+	%% Check solution difficulty
+	PrevDiffPair = ar_difficulty:diff_pair(PrevB),
+	LastRetarget = PrevB#block.last_retarget,
+	PrevTS = PrevB#block.timestamp,
+	DiffPair = {_PoA1Diff, Diff} = ar_retarget:maybe_retarget(PrevB#block.height + 1,
```
JamesPiechota, line 1801: we can't use `get_current_diff` any more because if we're switching to the sibling tip the difficulty cached in `node_state` may not be correct?

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1926,3 +1710,317 @@ dump_mempool(TXs, MempoolSize) ->
...
+		solution_hash = SolutionH,
+		start_interval_number = IntervalNumber,
+		step_number = StepNumber,
+		steps = SuppliedSteps
+	} = Solution,
+	MerkleRebaseThreshold = ?MERKLE_REBASE_SUPPORT_THRESHOLD,
+
+	#block{ indep_hash = PrevH, timestamp = PrevTimestamp,
+			wallet_list = WalletList,
+			nonce_limiter_info = PrevNonceLimiterInfo } = PrevB,
```
JamesPiechota, line 1737: One change here is that we use `PrevB` instead of the block values cached in `node_state`, right? This is because teh values cached in `node_state` are based on the current Tip block - but if we're in the middle of rebasing onto a sibling, the new 'tip' may not be the cached tip, right?

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1926,3 +1710,317 @@ dump_mempool(TXs, MempoolSize) ->
...
+					{step_number, StepNumber},
+					{prev_step_number, PrevStepNumber},
+					{prev_next_seed, ar_util:encode(PrevNextSeed)},
+					{output, ar_util:encode(NonceLimiterOutput)}]),
+			{noreply, State}
+	end.
+
+update_solution_cache(H, Args, State) ->
+	#{ solution_cache := Map, solution_cache_records := Q } = State,
+	Q2 = queue:in(H, Q),
```
JamesPiechota, line 2011: I think this line can move into the `false` branch below since it's not used in the `true` branch

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1926,3 +1710,317 @@ dump_mempool(TXs, MempoolSize) ->
...
+	case maps:is_key(H, Map) of
+		true ->
+			State;
+		false ->
+			Map2 = maps:put(H, Args, Map),
+			{Map3, Q3} =
+				case queue:len(Q2) > 5 of
```
JamesPiechota, line 2018: This logic is basically saying:
1. we only want to cache 5 solutions at max
2. if we exceed 5, we remove the oldest one from the `solution_cache`
3. `solution_cache_records` is only used to track which solution is oldest

right?

apps/arweave/src/ar_node_worker.erl
ldmberman: No two blocks at the same time may have different cdiff e.g. during retargets.

There is a pattern matching in the argument deconstruction; perhaps, I should make it more explicit.

apps/arweave/src/ar_node_worker.erl
ldmberman: The previous block is always cached but this time it is not necessarily the tip. I keep the approach that the node never "shifts to the side" - every state change is always a cumulative difficulty increase; a rebase may be looked at as a regular reorg except for a mining solution is reused.

apps/arweave/src/ar_node_worker.erl
ldmberman: yep

apps/arweave/src/ar_node_worker.erl
ldmberman: Yes, added some docs

apps/arweave/src/ar_node_worker.erl
JamesPiechota: Ah I see it now. Thanks!

JamesPiechota (approved):

ldmberman:
@JamesPiechota FYI I pushed https://github.com/ArweaveTeam/arweave/pull/545/commits/c6ca466d146f47c0cabfc4d973430b62a5661033 on top of previously reviewed.

JamesPiechota:
> @JamesPiechota FYI I pushed [c6ca466](https://github.com/ArweaveTeam/arweave/commit/c6ca466d146f47c0cabfc4d973430b62a5661033) on top of previously reviewed.
> 
> UPD nevermind, a similar change was in master already.

@ldmberman got it. let me know if you need me to re-review

PR #547, opened 2024-04-11
pool worker name

vird:


apps/arweave/src/ar.erl
```diff
@@ -311,7 +311,8 @@ show_help() ->
 			{"is_pool_client", "Configure the node as a pool client. The node may be an "
 					"exit peer in the coordinated mining setup or a standalone node."},
 			{"pool_api_key", "API key for the requests to the pool."},
-			{"pool_server_address", "The pool address"}
+			{"pool_server_address", "The pool address"},
+			{"pool_worker_name", "The pool worker name"}
```
JamesPiechota, line 315: is `pool_worker` the same as `pool_client`? i.e. Above we have a `is_pool_client` config. if `pool_worker` and `pool_client` are the same, can we call this `pool_client_name`?

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -1356,7 +1356,13 @@ cm_p2p_headers() ->
 
 pool_client_headers() ->
 	{ok, Config} = application:get_env(arweave, config),
-	add_header(<<"x-pool-api-key">>, Config#config.pool_api_key, p2p_headers()).
+	Headers1 = add_header(<<"x-pool-api-key">>, Config#config.pool_api_key, p2p_headers()),
+	case Config#config.pool_worker_name of
+		not_set ->
+			Headers1;
+		Worker_name ->
```
JamesPiechota, line 1363:
```suggestion
		WorkerName ->
```

apps/arweave/src/ar.erl
vird: We can call it pool_client_name, but it will be more confusing to miners from other cryptocurrencies because anybody else it's calling either `worker` or `worker name`

JamesPiechota: Got it. I do think we should be consistent within our own code first and then consistent across blockchains second. 

Do you want to update all 'pool client" instances to be "pool worker"? Sounds like that might be the best approach - although it'll be a pain to handle compatibility. 

Otherwise as a quick fix is suggest we continue using pool client until we can make a consistent update across the codebase. Having two names will just make things harder to maintain (and will be confusing to miners as well when they are configuring their miner)

vird: we have already intersection in names
worker is kinda reserved for worker inside mining process
client is for connecting pool
inside client there is just label what "pool label" called "worker" should be displayed on pool

JamesPiechota: All I'm saying is that right her in this section of code we hvae 3 different names:
client
server
worker

a user seeing that will assume they refer to 3 different roles. If some of them refer to the same role we should name them the same thing. Otherwise we'll get questions about whether or not they have to set `is_pool_client` if they want to set their `pool_worker_name`

I don't have a strong opinion about client vs. worker. but I do have a strong opinion that we should be consistent - particularly within a given functinal role.

I don't get what the problem is here? we literallly call it "pool_client" 2 lines earlier, doesn't it makes sense that it will be confusing if we cal it "pool_worker" here? 

Again I"m fine if we want to refer to this entity as a "pool_worker" everywhere, but I *really* don't want to introduce inconsistency that we don't have to. It just makes the code harder to understand, use, and maintain for no good reason.

vird: `pool_client` is mode of our arweave node
`pool_worker_name` is optional label on pool, that's why I added name
We can explicitly say in description that it's optional

JamesPiechota: Put another way. 

We have is_pool_client and is_pool_server implying those are the two roles you can configure your miner as.

if we introduce `pool_worker_name` the implication is that that is a 3rd role and we would need `is_pool_worker` - or the implication is that it independent of the two roles and so you could set `pool_worker_name` even if you set `is_pool_server`

it just gets needlessly confusing since the only time you set the name is if you ar a `pool_client` so it makes sense that it woudl be `pool_client_name`

apps/arweave/src/ar.erl
```diff
@@ -311,7 +311,8 @@ show_help() ->
 			{"is_pool_client", "Configure the node as a pool client. The node may be an "
 					"exit peer in the coordinated mining setup or a standalone node."},
 			{"pool_api_key", "API key for the requests to the pool."},
-			{"pool_server_address", "The pool address"}
+			{"pool_server_address", "The pool address"},
+			{"pool_worker_name", "(optional) The pool worker name"}
```
JamesPiechota, line 315: @vird fine if you insist on keeping the naming inconsistent, then I think you need to explain the flag better. Right now it is confusing to me, likely for users, and for new developers to the codebase. Nowhere else in our documentation do we refer to a pool worker. We refer to pool clients and pool servers - so your documentation should clarify what a pool worker name is and how it different from the pool client and pool server

apps/arweave/src/ar.erl
vird: pool_client_worker_name?

JamesPiechota: haha, I honestly do not understand what the difference is between a worker and a client. To me they are synonyms. If you think they are different thing that need to be referred differentl: okay, I can go along with that. But I personally would like to see it in the config flag help  - some explanation for what a "worker" is and why it is distinct from a "client"

vird: Added more verbose description

apps/arweave/src/ar.erl
```diff
@@ -311,7 +311,10 @@ show_help() ->
 			{"is_pool_client", "Configure the node as a pool client. The node may be an "
 					"exit peer in the coordinated mining setup or a standalone node."},
 			{"pool_api_key", "API key for the requests to the pool."},
-			{"pool_server_address", "The pool address"}
+			{"pool_server_address", "The pool address"},
+			{"pool_worker_name", "(optional) The pool worker name"
```
JamesPiechota, line 315:
```suggestion
			{"pool_worker_name", "(optional) The pool worker name. "
```

apps/arweave/src/ar.erl
JamesPiechota: Deal 🤝 

Looks good. I added a suggested space so the words don't run together.

JamesPiechota (approved):
LGTM. I added some suggested whitespace in the documentation, otherwise 🤝

apps/arweave/src/ar.erl
vird: tl;dr word client has no sence in terms of pool. So it was accidentally introduced in our code and now we are relying on it as standard

I propose to remove word client at all (in future, not now)
word client is only used in is_pool_client and in cm_peer description
Also in vdf_client_peer, but that's different

is_pool_client was introduced same time as is_pool_server to distinguish those 2 modes
So currently client word is totally redundant

How it should work. No matter CM + pool or CM + solomine. CM peers should be configured same way (currently it's maybe not)
So the only place when is_pool_client can be present it's exit peer, so in this case we already have pool_server_address

is_pool_server maybe keep as is

JamesPiechota: 👍 

logged: https://github.com/ArweaveTeam/arweave-dev/issues/464

PR #551, opened 2024-04-17
Fix/cm pool job routing

ldmberman:


apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -3198,27 +3203,32 @@ handle_mining_h2(Req, Pid) ->
...
+					case catch ar_serialize:json_map_to_candidate(JSON) of
+						{'EXIT', _} ->
+							{400, #{}, jiffy:encode(#{ error => invalid_json }), Req2};
+						Candidate ->
+							?LOG_INFO([{event, h2_received},
+									{peer, ar_util:format_peer(Peer)}]),
+							case {ar_pool:is_client(), ar_coordination:is_exit_peer()} of
+								{true, true} ->
+									PoolPeer = ar_pool:pool_peer(),
+									Jobs = #pool_cm_jobs{ h1_read_jobs = [Candidate] },
```
JamesPiechota, line 3217: `h1_read_jobs` means we've calculated an H2 solution be we don't have the h1 chunk locally so we're submitting a job for a different peer to "read the h1 chunk", right?

I'm trying to sort out why we submit an `h1` related job after mining an `h2` solution.

JamesPiechota (approved):
One question

apps/arweave/src/ar_http_iface_middleware.erl
ldmberman: Yes, except for we do not look for the first chunk at the second peer.

JamesPiechota: got it... I think. I can sort it out later. Resolving!

PR #555, opened 2024-04-20
fix: mining stats incorrect after repack in place

JamesPiechota:
Mining stats now allow for two storage modules to have the same store id but different packing

### DIFF
--- a/apps/arweave/src/ar_mining_stats.erl
+++ b/apps/arweave/src/ar_mining_stats.erl
@@ -770,24 +770,60 @@ format_vdf_speed(VDFSpeed) ->
 %%%===================================================================
 
 mining_stats_test_() ->
-	[
-		{timeout, 30, fun test_read_stats/0},
-		{timeout, 30, fun test_h1_stats/0},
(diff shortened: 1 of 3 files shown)


apps/arweave/src/ar_mining_stats.erl
```diff
@@ -334,7 +335,7 @@ get_overall_average(PartitionPeer, Stat, TotalCurrent) ->
 	end.
 
 get_partition_data_size(PartitionNumber) ->
-    Pattern = {{partition, PartitionNumber, storage_module, '_'}, '$1'},
+    Pattern = {{partition, PartitionNumber, storage_module, '_', packing, '_'}, '$1'},
```
ldmberman, line 338: We sum all present packings but mine with only one key so the reported metrics are skewed. Should not we look for the mining address-related packing here?

JamesPiechota: ah yeah, you're right.

JamesPiechota: Updated in latest commits

ldmberman (approved):

PR #556, opened 2024-04-24
Implement webhook for tracking tx synchronization

ldmberman:


apps/arweave/src/ar_data_sync.erl
```diff
@@ -1660,6 +1665,38 @@ get_tx_offset(TXIndex, TXID) ->
 			{error, failed_to_read_offset}
 	end.
 
+get_tx_offset_data_in_range(TXOffsetIndex, TXIndex, Start, End) ->
```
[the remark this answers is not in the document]
JamesPiechota: looking at `ar_kv:get_range3` it looks like it will only include transactions whose data is fully contained within Start and End. Which sounds like the right approach to me!

apps/arweave/src/ar_storage_module.erl
```diff
@@ -204,3 +219,41 @@ get_all(Start, End, [StorageModule | StorageModules], FoundModules) ->
...
+			end
+		end,
+		StorageModules
+	).
+
+get_cover2(Start, End, _StorageModules)
+		when Start >= End ->
+	[];
+get_cover2(Start, _End, [{BucketSize, Bucket, _Packing} | _StorageModules])
+		when BucketSize * Bucket > Start ->
```
JamesPiechota, line 249: By my read this should be `BucketSize * Bucket > End` 

What am I missing?

apps/arweave/src/ar_sync_record.erl
```diff
@@ -201,6 +202,17 @@ get_next_synced_interval(Offset, RightBound, ID, StoreID) ->
 			ar_ets_intervals:get_next_interval(TID, Offset, RightBound)
 	end.
 
+%% @doc Return the lowest unsynced interval with the end offset strictly above the given Offset
+%% and with the right bound at most RightBound.
```
JamesPiechota, line 206: What's the difference between "end offset" and "right bound"? I may have the layout flipped in my head, but I would have assumed they are the same?

apps/arweave/src/ar_webhook.erl
```diff
@@ -29,10 +20,27 @@
 	{<<"content-type">>, <<"application/json">>}
 ]).
 
+-define(MAX_TX_OFFSET_CACHE_SIZE, 100000).
+
+-record(tx_offset_cache, {
+	txid_timestamp = #{},
```
JamesPiechota, line 26:
```suggestion
	txid_timestamp = #{}, %% map TXID to timestamp
```

JamesPiechota: what do you think about adding `to` in the map names to clarify which element is the key vs. value? More important for `txid_start_offset_end_offset` than `txid_timestamp`

apps/arweave/src/ar_webhook.erl
```diff
@@ -29,10 +20,27 @@
...
+	timestamp_txid = gb_sets:new(),
+	txid_start_offset_end_offset = #{},
+	start_offset_txid_end_offset = gb_sets:new(),
+	size = 0,
+	txid_status = #{}
```
JamesPiechota, line 31:
```suggestion
	txid_to_status = #{} %% map TXID to Status
```

apps/arweave/src/ar_webhook.erl
```diff
@@ -29,10 +20,27 @@
 	{<<"content-type">>, <<"application/json">>}
 ]).
 
+-define(MAX_TX_OFFSET_CACHE_SIZE, 100000).
+
+-record(tx_offset_cache, {
+	txid_timestamp = #{},
+	timestamp_txid = gb_sets:new(),
+	txid_start_offset_end_offset = #{},
+	start_offset_txid_end_offset = gb_sets:new(),
```
JamesPiechota, line 29:
```suggestion
	start_offset_txid_end_offset = gb_sets:new(), %% set of {StartOffset, TXID, EndOffset} tuples
```

apps/arweave/src/ar_webhook.erl
```diff
@@ -29,10 +20,27 @@
 	{<<"content-type">>, <<"application/json">>}
 ]).
 
+-define(MAX_TX_OFFSET_CACHE_SIZE, 100000).
```
JamesPiechota, line 23:
```suggestion
-define(MAX_TX_OFFSET_CACHE_SIZE, 100_000). %% Max number of transactions cached
```

apps/arweave/src/ar_config.erl
```diff
@@ -693,6 +693,7 @@ parse_webhook([], Webhook) ->
 parse_webhook_events([Event | Rest], Events) ->
```
JamesPiechota, line 693: Do we have any documention on webhooks and how to specify them?

If not maybe worth adding to ar.erl? Perhaps as a section at the end with an explanation for config-only options

it looks like we have *some* config-only documentation under the ar.erl `config_file` option (it describes the `semaphores` option)

apps/arweave/src/ar_webhook.erl
```diff
@@ -134,16 +151,60 @@ handle_info({event, block, {new, Block, _Source}}, State) ->
 handle_info({event, block, _}, State) ->
 	{noreply, State};
 
-handle_info({event, tx, {new, TX, _Source}}, State) ->
+handle_info({event, tx, {new, TX, _Source}},
+		#state{ listen_to_transaction_stream = true } = State) ->
 	?LOG_DEBUG("Web hook triggered on: tx ~s", [ar_util:encode(TX#tx.id)]),
```
JamesPiechota, line 156:
```suggestion
```

apps/arweave/src/ar_webhook.erl
```diff
@@ -124,7 +140,8 @@ handle_cast(Msg, State) ->
 %%									 {stop, Reason, State}
 %% @end
 %%--------------------------------------------------------------------
-handle_info({event, block, {new, Block, _Source}}, State) ->
+handle_info({event, block, {new, Block, _Source}},
+		#state{ listen_to_block_stream = true } = State) ->
 	?LOG_DEBUG("Web hook triggered on: block ~s", [ar_util:encode(Block#block.indep_hash)]),
```
JamesPiechota, line 145:
```suggestion
```

apps/arweave/src/ar_webhook.erl
```diff
@@ -134,16 +151,60 @@ handle_info({event, block, {new, Block, _Source}}, State) ->
...
 	call_webhook(URL, Headers, TX, transaction),
 	{noreply, State};
 
+handle_info({event, tx, {registered_offset, TXID, EndOffset, Size}},
+		#state{ listen_to_transaction_data_stream = true } = State) ->
+	Start = EndOffset - Size,
+	Cache = State#state.tx_offset_cache,
+	case cache_is_tx_data_marked_synced(TXID, Cache) of
+		true ->
+			{noreply, State};
```
JamesPiechota, line 168:
```suggestion
			%% This event has already been processed for TXID, no need to process again
			{noreply, State};
```

apps/arweave/src/ar_webhook.erl
```diff
@@ -29,10 +20,27 @@
 	{<<"content-type">>, <<"application/json">>}
 ]).
 
+-define(MAX_TX_OFFSET_CACHE_SIZE, 100000).
+
+-record(tx_offset_cache, {
```
JamesPiechota, line 25:
```suggestion
%% Functions which update or query the #tx_offset_cache are named with the `cache_` prefix e.g. `cache_is_tx_data_marked_synced/2`
-record(tx_offset_cache, {
```

apps/arweave/src/ar_webhook.erl
```diff
@@ -239,16 +300,306 @@ do_call_webhook(URL, Headers, Entity, Event, _N) ->
...
+process_updated_tx_data([], _StoreID, State) ->
+	State;
+process_updated_tx_data([{TXID, Start, End} | Data], StoreID, State) ->
+	Cache = State#state.tx_offset_cache,
+	case cache_is_tx_data_marked_synced(TXID, Cache) of
+		true ->
+			process_updated_tx_data(Data, StoreID, State);
+		false ->
+			process_updated_tx_data(Data, StoreID,
+					maybe_call_transaction_data_synced(Start, End, TXID, StoreID, State))
```
JamesPiechota, line 355:
```suggestion
					maybe_call_transaction_data_synced_webhook(Start, End, TXID, StoreID, State))
```

(I don't feel strongly about this rename - I think it's clearer, but also makes a long function longer....)

apps/arweave/src/ar_webhook.erl
```diff
@@ -239,16 +300,306 @@ do_call_webhook(URL, Headers, Entity, Event, _N) ->
...
+	]),
+	State.
+
+process_updated_tx_data([], _StoreID, State) ->
+	State;
+process_updated_tx_data([{TXID, Start, End} | Data], StoreID, State) ->
+	Cache = State#state.tx_offset_cache,
+	case cache_is_tx_data_marked_synced(TXID, Cache) of
+		true ->
+			process_updated_tx_data(Data, StoreID, State);
```
JamesPiechota, line 352:
```suggestion
			%% transaction_data_synced webhook has already been sent, advance to next range
			process_updated_tx_data(Data, StoreID, State);
```

apps/arweave/src/ar_webhook.erl
```diff
@@ -239,16 +300,306 @@ do_call_webhook(URL, Headers, Entity, Event, _N) ->
...
+		false ->
+			process_updated_tx_data(Data, StoreID,
```
JamesPiechota, line 354:
```suggestion
			%% Maybe send the transaction_data_synced webhook before advancing to the next range
			process_updated_tx_data(Data, StoreID,
```

apps/arweave/src/ar_webhook.erl
```diff
@@ -239,16 +300,306 @@ do_call_webhook(URL, Headers, Entity, Event, _N) ->
...
+		start_offset_txid_end_offset = Set
+	} = Cache,
+	Iterator = gb_sets:iterator_from({Start - 1, infinity, infinity}, Set),
+	cache_get_tx_offset_data_from_iterator(End, Iterator).
+
+cache_get_tx_offset_data_from_iterator(End, Iterator) ->
+	case gb_sets:next(Iterator) of
+		none ->
+			[];
+		{{Start2, TXID, End2}, Iterator2} when Start2 < End ->
```
JamesPiechota, line 386: This function includes all ranges for which Start is less than End (i.e. it may include ranges for which End2 is greater than End)

By my read the [get_tx_offset_data_in_range](https://github.com/ArweaveTeam/arweave/pull/556/files#diff-2934ddad2ed46ac194a6be7abde4be6d440bc7565e1128bba234fad9fd2bd5d2R1668) function only includes ranges where both Start and End are within the queried Range.

Just confirming that I've understood the differing logic correctly and that this difference is expected/fine.

apps/arweave/src/ar_data_sync.erl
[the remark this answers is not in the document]
ldmberman: Fixed a bug [here](https://github.com/ArweaveTeam/arweave/pull/556/commits/aaa996b757a01dff6205c38c3c7d323a02386144) where a TXID won't be discovered if the given range is somewhere in the middle.

The function returns a list of {TXID, StartOffset, EndOffset} where StartOffset, EndOffset are transaction borders (not clipped by the given range) for all TXIDs intersecting the given range.

apps/arweave/src/ar_webhook.erl
ldmberman: Updated it [here](https://github.com/ArweaveTeam/arweave/pull/556/commits/eafad4b57c915808f2a7948d54d943e5315971ee) - changed from {Start, TXID, End} to {End, TXID, Start} because gb_sets does not have an ar_kv:get_prev analogue; the logic should be the same across two places now.

apps/arweave/src/ar_storage_module.erl
ldmberman: The modules are sorted so if the module's start is larget than Start there is no cover and we need to return not_found. There was another bug here though, [the fix](https://github.com/ArweaveTeam/arweave/pull/556/commits/277d9fd6289c0d541a561e75c29ec53cca8695e5).

apps/arweave/src/ar_sync_record.erl
ldmberman: Renamed it to [end offset upper bound](https://github.com/ArweaveTeam/arweave/pull/556/commits/9bff8968d2fb4c66faf098743cfc04a04b36cc1a).

apps/arweave/src/ar_storage_module.erl
JamesPiechota: Ah right, I was flipping it in my head. `[Start...End]` has to be *fully* contained within the module (not the other way around: module fully contained within range), so if the module's range starts after `Start` then at least part of the search range is outside the module. Got it!

apps/arweave/src/ar_data_sync.erl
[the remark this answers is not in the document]
JamesPiechota: Thanks! That explanation helps! I've suggested updating the comment with your wording.

apps/arweave/src/ar_webhook.erl
```diff
@@ -29,10 +20,29 @@
 	{<<"content-type">>, <<"application/json">>}
 ]).
 
+-define(MAX_TX_OFFSET_CACHE_SIZE, 100_000). %% Max number of transactions cached
+
+%% Functions which update or query the #tx_offset_cache are named with the `cache_` prefix
+%% e.g. `cache_is_tx_data_marked_synced/2`.
+-record(tx_offset_cache, {
+	txid_to_timestamp = #{}, % Map TXID => Timestamp
+	timestamp_to_txid = gb_sets:new(), % Ordered set of {Timestatmp, TXID}
```
JamesPiechota, line 29: Sorry :( I didn't mean to add `_to_` to the set names, just the map names (to differentiate key from value). I'm not sure how to improve the set names over what you had before (e.g. `timestamp_txid`) and I'm not sure adding `_to_` to the set names helps with readability.

Commenting here just as FYI. I"m also fine if you leave the `_to_` in there on the set names - we can update later if needed.

JamesPiechota (approved):

apps/arweave/src/ar_storage_module.erl
```diff
@@ -159,6 +162,22 @@ get_all(Start, End) ->
 	{ok, Config} = application:get_env(arweave, config),
 	get_all(Start, End, Config#config.storage_modules, []).
 
+%% @doc Return the list of at least one {Start, End, StoreID} covering the given range
+%% or not_found. The given StoreID (may be none) has a higher chance to be picked in case
+%% there are several storage modules covering the same range.
```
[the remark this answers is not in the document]
ldmberman: It would be

RangeA: [{2, 8, sm_1}]
RangeB: [{7, 10, sm_1}, {10, 13, sm_2}]
RangeC: [{7, 10, sm_1}, {10, 20, sm_2}, {20, 25, sm_3}]

apps/arweave/src/ar_webhook.erl
ldmberman: No worries! I do not have a strong preference either, reverted

apps/arweave/src/ar_storage_module.erl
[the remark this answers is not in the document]
ldmberman: Added some more [docs and tests](https://github.com/ArweaveTeam/arweave/pull/556/commits/7cf7f5b9ccf5e2cbfbac7ca7327c0b3608bb5d15)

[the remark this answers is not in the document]
JamesPiechota: Awesome! That helps a lot!

JamesPiechota:
LGTM!

JamesPiechota (approved):

apps/arweave/src/ar_webhook.erl
JamesPiechota: Great!

PR #557, opened 2024-04-24
pool share label

vird:


apps/arweave/include/ar_mining.hrl
```diff
@@ -28,7 +28,8 @@
 	seed = not_set, %% serialized
 	session_key = not_set, %% serialized
 	start_interval_number = not_set, %% serialized
-	step_number = not_set %% serialized
+	step_number = not_set, %% serialized
+	label = <<"not_set">> %% not term, for prevent term table pollution DoS
```
JamesPiechota, line 32: "term" here mean `atom`? Like you want to prevent a malicious actor from submitting a lot of different shares with different labels and overflowing the erlang 1,048,576 atom limit?

JamesPiechota (approved):
LGTM! 

I had a question about one comment. If I understand correctly, might be worth changing "term" to "atom"

PR #558, opened 2024-04-29
Refactor block time  history handling

JamesPiechota:
Refactor block time  history handling
1. extract common functionality into ar_block_time_history
2. simply a lot of function calls by passing a block down the call chain and then only extracting the relevant information at the leaf of the call stack

### DIFF
--- a/apps/arweave/src/ar_block_time_history.erl
+++ b/apps/arweave/src/ar_block_time_history.erl
@@ -1,0 +1,130 @@
+-module(ar_block_time_history).
+
+-export([history_length/0, has_history/1, get_history/1, get_history_from_blocks/2, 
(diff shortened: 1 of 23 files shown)


apps/arweave/src/ar_block_time_history.erl
```diff
@@ -0,0 +1,130 @@
...
+validate_hashes(History, [H | Hashes]) ->
+	case validate_hash(H, History) of
+		true ->
+			validate_hashes(tl(History), Hashes);
+		false ->
+			false
+	end.
+
+validate_hash(H, History) ->
+	H == hash(lists:sublist(History, history_length())).
```
ldmberman, line 98: An extra sublist here

ldmberman (approved):

PR #563, opened 2024-05-01
Add detailed fetch proofs for solutions logging

ldmberman:


apps/arweave/src/ar_data_sync.erl
```diff
@@ -329,10 +329,38 @@ get_chunk(Offset, #{ packing := Packing } = Options) ->
...
+				true ->
+					UnpackedReply = ar_sync_record:is_recorded(Offset, {?MODULE, unpacked}),
+					?LOG_ERROR([{event, chunk_record_not_associated_with_packing},
+							{tags, [solution_proofs]},
+							{store_id, StoreID},
+							{seek_offset, SeekOffset},
+							{is_recorded_unpacked, io_lib:format("~p", [UnpackedReply])}])
+			end,
+			{error, chunk_not_found};
 		_ ->
```
JamesPiechota, line 350: Would it be worth also logging the value of the reply here? In case it's something unexpected?

JamesPiechota: also maybe the result of `ets:lookup(sync_records, {ID, StoreID})`?

I'm wondering if the `sync_records` table got corrupt, maybe the `{ID, StoreID}` is not responding correctly

apps/arweave/src/ar_mining_server.erl
```diff
@@ -598,13 +604,20 @@ prepare_solution(poa2, Candidate,
 		partition_upper_bound = PartitionUpperBound } = Candidate,
 	case read_poa(RecallByte2, Chunk2, MiningAddress) of
 		error ->
+			Modules = ar_storage_module:get_all(RecallByte2 + 1),
```
JamesPiechota, line 607: right now the `read_poa` wrapper throws out some error information [here](https://github.com/ArweaveTeam/arweave/pull/563/files#diff-ea8e957c63acb7497b0ed4f9201b40f18e5d5b4fa371e67b4cff05850e4df854R796-R810)

What do you think about either passing that error information up to here so we can log it, or printing it out in the wrapper directly?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -1600,6 +1685,13 @@ read_chunk_with_metadata(
 					case ar_sync_record:is_recorded(Offset, StoredPacking, ?MODULE,
 							StoreID) of
 						false ->
+							?LOG_ERROR([{event,
```
JamesPiechota, line 1688: Similar comment as above, what do you tink about logging `ets:lookup(sync_records, {ID, Type, StoreID})` to see if maybe the `sync_records` db has gotten corrupt somehow so even the top level key is not found.

PR #564, opened 2024-05-02
Improvement/public vdf server

ldmberman:


JamesPiechota:
Looks good!

This makes 3 changes right?

1. All VDF clients will pull by default (unless `disable vdf_server_pull`)
2. A VDF server can opt to be "public" by setting `enable public_vdf_server`
3. A VDF server will cache its latest VDF data in ETS so that it can better handle a ton of GET requests if needed

JamesPiechota (approved):

PR #566, opened 2024-05-08
Fix bug where VDF client would get stuck validating orphaned VDF

ldmberman:


apps/arweave/src/ar_node_worker.erl
```diff
@@ -1041,18 +1045,17 @@ apply_block(State) ->
 			{noreply, State};
 		{B, [PrevB | _PrevBlocks], {{not_validated, awaiting_nonce_limiter_validation},
 				_Timestamp}} ->
-			case maps:get(nonce_limiter_validation_scheduled, State, false) of
+			H = B#block.indep_hash,
+			case maps:get({nonce_limiter_validation_scheduled, H}, State, false) of
```
JamesPiechota, line 1049: Is this correct?

Initial chain:
```
A -> B
```
We enter this code path and request validation for B. Let's say another block comes in:
```
A -> B -> C
```
B is still the earliest so we will continue to wait on B's validation to come in (and won't start working on C in the meantime).

Then a bunch more blocks come in:
```
A -> B -> C
    \> D -> E -> F
```
Under the old code even though D was now identified as the earliest not validated in longest chain, we'd continue to wait on B's validation since the `nonce_limiter_validation_scheduled` flag was set. This caused a problem if the VDF server was on a different fork as B would never validate.

With your fix since the `nonce_limit_validation_scheduled` flag is now based on the hash, we enter this code path and see that `{nonce_limiter_validation_scheduled, D}` is not set and so we start validating that. This unblocks the node and it can continue.

B will never validate (and will eventually trigger a `refuse_validation` event which will clear the flag so we stop waiting)

Is that correct?

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -306,9 +306,8 @@ request_validation(H, #nonce_limiter_info{ output = Output,
 			case ar_config:use_remote_vdf_server() of
 				true ->
 					%% Wait for our VDF server(s) to validate the remaining steps.
-					spawn(fun() ->
-						timer:sleep(1000),
-						request_validation(H, Info, PrevInfo) end);
+					%% Alternatively, the network may abandon this block.
```
JamesPiechota, line 309: We send `refuse_validation` to clear the `scheduled` flag, but if H is still the earliest unvalidated in longest chain, we would expect it to be re-validated the next time we go to apply a block, right?

JamesPiechota (approved):
LGTM! I had 2 questions to make sure I understand the code - but assuming I have, then it looks good!

apps/arweave/src/ar_node_worker.erl
ldmberman: Yes, precisely!

apps/arweave/src/ar_nonce_limiter.erl
ldmberman: Yep

PR #570, opened 2024-05-15
Fix/vdf client incomplete prev session

ldmberman:


apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -92,6 +97,23 @@ handle_cast(pull, State) ->
...
+	{{value, RawPeer}, Q2} = queue:out(Q),
+	State2 = State#state{ remote_servers = queue:in(RawPeer, Q2) },
+	case resolve_peer(RawPeer) of
+		error ->
+			%% Push the peer to the back of the queue.
+			{noreply, State2};
+		{ok, Peer} ->
+			case fetch_and_apply_session_and_previous_session(Peer) of
+				ok ->
+					{noreply, State};
```
JamesPiechota, line 111: Is there any risk that the VDF server will respond with a stale session because it itself has gotten stuck? Or perhaps it's a slow VDF server? i.e. is there any concern that we might get stuck always asking the same slow VDF server for the session and never rotate the queue?

apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -21,6 +21,11 @@
 start_link() ->
 	gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
 
+%% @doc Pick a server from the list of configured VDF servers and request complete
+%% VDF sessions from them. Failures to resolve a domain of make a request are not retried.
```
JamesPiechota, line 25: Not retrying and only asking one VDF server is okay because we'll end up retrying with a new VDF server the next time we try to apply the block and can't validate it, right?

apps/arweave/src/ar_nonce_limiter_client.erl
ldmberman: If the server is stuck, the function should return a non-ok response, e.g., a #nonce_limiter_update_response with a step number indicating we are ahead. Good catch, I've found and patched two places where [it was not the case](https://github.com/ArweaveTeam/arweave/pull/570/commits/538b5978787f0acebe1fd4ce6a1f3a90dc85d16f).

JamesPiechota: makes sense!

JamesPiechota (approved):
LGTM!

Unfortunately our auto tests aren't running... I'll try to get them running again so we can confirm no test failures.

PR #577, opened 2024-05-23
Fix/vdf step checkpoints

ldmberman:


apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -121,6 +121,9 @@ name_route([<<"vdf">>, <<"session">>]) ->
 	"/vdf/session";
 name_route([<<"vdf2">>, <<"session">>]) ->
 	"/vdf2/session";
+name_route([<<"vdf3">>, <<"session">>]) ->
```
JamesPiechota, line 124: Eventually we can get rid of the old formats, right? We need to introduce them mostly to handle the upgrade from one release to the next, but once everyoneis on the same release we can retire the old endpoints, right?

E..g. we can retire format 1, /vdf, and /vdf/session whenever we're ready, right? (since it's been a long time and at least 1 hard fork since we introduce them)

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -1341,6 +1352,18 @@ get_triplets(StepNumber, [Step | Steps], ResetPoint, UpperBound, NextUpperBound,
...
+	{[], 0};
+filter_step_triplets_with_checkpoints([{_, StepNumber, _} = Triplet | Triplets], Map) ->
+	case maps:is_key(StepNumber, Map) of
+		true ->
+			{List, NSkipped} = filter_step_triplets_with_checkpoints(Triplets, Map),
+			{[Triplet | List], NSkipped};
+		false ->
+			{List, NSkipped} = filter_step_triplets_with_checkpoints(Triplets, Map),
+			{List, NSkipped + 1}
+	end.
```
JamesPiechota, line 1365:
```suggestion
	{List, NSkipped} = filter_step_triplets_with_checkpoints(Triplets, Map),
	case maps:is_key(StepNumber, Map) of
		true ->
			{[Triplet | List], NSkipped};
		false ->
			{List, NSkipped + 1}
	end.
```

apps/arweave/src/ar_http_iface_server.erl
ldmberman: Makes sense

apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -39,68 +39,92 @@ init([]) ->
...
+								#nonce_limiter_update_response{ session_found = false } ->
+									case fetch_and_apply_session_and_previous_session(Peer) of
+										ok ->
+											ar_util:cast_after(800, ?MODULE, pull),
+											{noreply, State};
+										_ ->
+											gen_server:cast(?MODULE, pull),
+											{noreply, State2}
+									end;
+								#nonce_limiter_update_response{ step_number = StepNumber }
```
JamesPiechota, line 85: Interesting. This logic didn't exist before? So a client could get stuck on a slow VDF server until taht VDF server timed out or threw some other error?

oh, actually, every we call `pull` we rotate server anyways, right? so no stickiness.

apps/arweave/src/ar_nonce_limiter_server.erl
```diff
@@ -12,6 +12,10 @@
 
 -record(state, {}).
 
+%% @doc The number of steps for which we include step checkpoints in the full session update.
+%% Does not apply to previous session updates.
+-define(SESSION_UPDATE_INCLUDE_STEP_CHECKPOINTS_COUNT, 10).
```
JamesPiechota, line 17: What do you thknk about increasing this?

Worst case sizing, assuming N steps of checkpoints returned every second (N * 50 * 32 byets):

10:
Request: 16KB
Per Day: 1.4GB

25:
Request: 40KB
Per Day: 3.5GB

100:
Request: 160KB
Per Day: 14GB

For my miner ~96% of requests yielded 1 step. So these "worse case" numbers are likely pretty inflated. Even a 100 step cap might only add ~600MB of traffic a day. which is a relatively small delta over the current web traffic.

apps/arweave/src/ar_http_iface_server.erl
JamesPiechota: Logged for later: https://github.com/ArweaveTeam/arweave-dev/issues/513

apps/arweave/src/ar_serialize.erl
```diff
@@ -375,7 +375,30 @@ nonce_limiter_update_to_binary(2 = _Format, #nonce_limiter_update{
 	CheckpointLen = length(Checkpoints),
```
JamesPiechota, line 368: Do you think we can remove the `checkpoints` field from `#nonce_limiter_update`? Now that we have `ar_serialize` query the step_checkpoints_map, I think the `checkpoints` field is redundant, right? (and may int eh future be a source of discrepancy/error)

Not necessary to remove it in this PR, but if that sounds like a good move I can add an issue for later. (whenever "later" actually comes)

apps/arweave/src/ar_nonce_limiter_server.erl
ldmberman: Yes, good point, I do not mind trying 100

apps/arweave/src/ar_nonce_limiter_server.erl
```diff
@@ -109,12 +113,27 @@ terminate(_Reason, _State) ->
...
 	StepNumber = Session#vdf_session.step_number,
 	Checkpoints = maps:get(StepNumber, Session#vdf_session.step_checkpoints_map, []),
+	StepCheckpointsMap = Session#vdf_session.step_checkpoints_map,
 	%% Clear the step_checkpoints_map to cut down on the amount of data pushed to each client.
+	RecentStepNumbers = get_recent_step_numbers(StepNumber),
+	StepCheckpointsMap2 = maps:with(RecentStepNumbers, StepCheckpointsMap),
 	#nonce_limiter_update{ session_key = SessionKey,
 			is_partial = IsPartial, checkpoints = Checkpoints,
-			session = Session#vdf_session{ step_checkpoints_map = #{} } }.
+			session = Session#vdf_session{ step_checkpoints_map = StepCheckpointsMap2 } }.
```
JamesPiechota, line 128: See [comment below](https://github.com/ArweaveTeam/arweave/pull/577/files#r1612193904)

We can (eventually) remove the `checkpoints` field, right? Since the data it sores is redundant to `session#vdf_session.step_checkpoints_map`

apps/arweave/src/ar_serialize.erl
```diff
@@ -443,8 +446,51 @@ binary_to_nonce_limiter_update(<< NextSeed:48/binary,
...
+			session = Session = #vdf_session{ step_number = StepNumber, seed = Seed,
+					upper_bound = UpperBound, next_upper_bound = NextUpperBound2,
+					steps = parse_32b_list(Steps),
+					step_checkpoints_map = StepCheckpointsMap } },
+	case decode_session_key(PrevSessionKeyBin) of
+		undefined ->
+			{ok, Update};
+		error ->
+			{error, invalid1};
+		SessionKey ->
```
JamesPiechota, line 476:
```suggestion
		PrevSessionKey ->
```

Right? or is this actually the current session key?

JamesPiechota (approved):
LGTM!

PR #578, opened 2024-05-27
Fix/vdf server rotation

JamesPiechota:


apps/arweave/include/ar_blacklist_middleware.hrl
```diff
@@ -74,9 +80,15 @@ end).
 		[<<"vdf">>] ->
 			{get_vdf, maps:get(get_vdf, LimitByIP, 180)};
 		[<<"vdf">>, <<"session">>] ->
-			{get_vdf, maps:get(get_vdf, LimitByIP, 180)};
+			{get_vdf_session, maps:get(get_vdf_session, LimitByIP, 30)};
+		[<<"vdf2">>, <<"session">>] ->
+			{get_vdf_session, maps:get(get_vdf_session, LimitByIP, 30)};
+		[<<"vdf3">>, <<"session">>] ->
+			{get_vdf_session, maps:get(get_vdf_session, LimitByIP, 30)};
```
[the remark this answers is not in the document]
ldmberman: Are you referring to numerous "vdf_validation_start .." messages? Those should be serial in fact, at most about two messages per second. I added a [patch](https://github.com/ArweaveTeam/arweave/pull/578/commits/82b3d9ed27e41e625ce4c777f73c34c2acbaad17) to avoid making extra session requests when they are not needed.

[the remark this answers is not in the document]
JamesPiechota: Ah no, something different. Let me find the logs. Sounds like a separate issue.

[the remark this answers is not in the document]
JamesPiechota: Hm, yeah we can ignore this. I remember seeing some logs yesterday where it looks like my miner was validating the same block multiple times - like it received the same block several times from the same IP (an ario node I think), and kicked off multiple concurrent validations processes for it. But I can't find the logs anymore. So let's ignore.

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -1074,117 +1074,127 @@ apply_external_update2(Update, State) ->
...
+					%% CurrentStepNumber (our most recently processed step number)
+					%% and ignores it and any lower steps found in Session.
+					{_, Steps} = get_step_range(Session, CurrentStepNumber + 1, StepNumber),
+					State2 = apply_external_update4(State, SessionKey, Session, Steps),
 					{reply, ok, State2}
-			end;
-		CurrentSession ->
-			#vdf_session{ step_number = CurrentStepNumber } = CurrentSession,
-			case CurrentStepNumber + 1 == StepNumber of
-				true ->
```
JamesPiechota, line 1116: It looks like we lost this clause in the refactoring. But I don't know if it's needed any more - I admit I can't full trace all the nested case statements to confirm the flow.

JamesPiechota: Ah maybe this clause is just a special case of https://github.com/ArweaveTeam/arweave/pull/578/files#diff-f702581474afe210531b226e33de974a1134010509c618ed920cd12da463f82eR1151-R1156

apps/arweave/src/ar_nonce_limiter_server.erl
```diff
@@ -12,9 +12,13 @@
 
 -record(state, {}).
 
+%% @doc The number of steps and the corresponding step checkpoints to include in every
+%% regular update.
+-define(REGULAR_UPDATE_INCLUDE_STEPS_COUNT, 2).
```
JamesPiechota, line 17: The reason for this is to limit the number of `/vdf3/session` requests that the client makes? i.e. if 1 step gaps are common, then by always sending 2 steps we can elimininate all those /vd3/session queries that we used to do for 1-step gaps

Is that about right? or is there some other reason?

apps/arweave/src/ar_nonce_limiter_server.erl
```diff
@@ -79,25 +98,11 @@ handle_cast(Cast, State) ->
 	{noreply, State}.
 
 handle_info({event, nonce_limiter, {computed_output, Args}}, State) ->
-	{SessionKey, StepNumber, Output, _PartitionUpperBound} = Args,
-	case ar_nonce_limiter:get_session(SessionKey) of
-		not_found ->
-			?LOG_WARNING([{event, computed_output_session_not_found},
-					{session_key, ar_nonce_limiter:encode_session_key(SessionKey)},
-					{step_number, StepNumber}]),
+	case is_vdf_server() of
```
JamesPiechota, line 101: Should we just not run the `ar_nonce_limiter_server_sup` if it's not a vdf server? That would trim out the server and all server_workers as well

JamesPiechota: It looks like we only start the workers if we have `nonce_limiter_client_peers` defined, so it's only the `_server` that's running anyways

apps/arweave/src/ar_nonce_limiter_server.erl
ldmberman: Yes, that's the reason

apps/arweave/src/ar_nonce_limiter.erl
ldmberman: Yes, exactly

apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -67,43 +75,47 @@ handle_cast(pull, State) ->
...
-										_ ->
-											gen_server:cast(?MODULE, pull),
 											{noreply, State2}
 									end;
 								#nonce_limiter_update_response{ step_number = StepNumber }
 										when StepNumber >= SessionStepNumber ->
-									%% We are ahead of this server; try another one,
-									%% if there are any.
-									gen_server:cast(?MODULE, pull),
+									%% We are ahead of this server. Re-try soon.
```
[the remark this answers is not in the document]
ldmberman: Yes, good point, I will split StepNumber >= SessionStepNumber into > (rotate) and == (do not rotate, re-try soon).

JamesPiechota (commented):
LGTM!

PR #579, opened 2024-05-29
Feature/hybrid vdf mode

ldmberman:


apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -303,7 +303,7 @@ request_validation(H, #nonce_limiter_info{ output = Output,
 			spawn(fun() -> ar_events:send(nonce_limiter, {invalid, H, 2}) end);
 		{RemainingStepsToValidate, NumAlreadyComputed}
 		  		when StartStepNumber + NumAlreadyComputed < StepNumber ->
-			case ar_config:use_remote_vdf_server() of
+			case ar_config:use_remote_vdf_server() and not ar_config:run_hybrid_vdf_mode() of
```
JamesPiechota, line 306: Our intent is to basically have 3 different modes that can be mixed and matched, right?
1. Compute VDF locally (e.g. `compute_vdf`)
2. Share VDF with clients (e.g. `vdf_server`)
3. Get VDF from servers (e.g. `vdf_client`)

I'm having a little trouble projecting those 3 modes onto our legacy flags/config. E.g. where does `run_hybrid_vdf_mode()` fit in? Does it imply `vdf_server andalso vdf_client`?

Is it worth reworking our flags to match up with the new set of independent configurations? Or is it really just a "me problem" and I should figure out how to map the old config onto the new "hybrid mode"?

ldmberman: What do you think about renaming it to "compute_own_vdf", on by default when no VDF servers are specified, off by default otherwise?

JamesPiechota:
Sounds good

apps/arweave/include/ar_blacklist_middleware.hrl
```diff
@@ -44,10 +44,14 @@ end).
 			{get_vdf_session, maps:get(get_vdf_session, LimitByIP, 120)};
 		[<<"vdf3">>, <<"session">>] ->
 			{get_vdf_session, maps:get(get_vdf_session, LimitByIP, 120)};
+		[<<"vdf4">>, <<"session">>] ->
```
JamesPiechota, line 47: is /vdf4 a replacement for /vdf3?

Since /vdf3 was only released as part of the pre-release, will be able to remove it once "everyone" has upgraded to the final 2.7.4 release?

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -417,34 +417,21 @@ init([]) ->
...
-		case ar_config:use_remote_vdf_server() of
+		case ar_config:compute_own_vdf() of
 			false ->
-				State;
+				State#state{ autocompute = false };
 			true ->
-				resolve_remote_server_raw_peers(),
-				gen_server:cast(?MODULE, check_external_vdf_server_input),
-				State#state{ autocompute = false }
+				State
```
JamesPiechota, line 431: I think we can replace this with:

```
State2 = State#state{ autocompute = ar_config:compute_own_vdf() },
```
`autocompute` defaults to `true` and is only set to false if compute_own_vdf() is false. So I think ogically it's the same.

apps/arweave/src/ar_peers.erl
```diff
@@ -300,22 +304,31 @@ discover_peers() ->
 	end.
 
 %% @doc Resolve the domain name of the given peer (if the given peer is an IP address)
-%% and cache it. Return {ok, Peer} | {error, Reason}.
+%% and cache it. Invalidate the cache after ?STORE_RESOLVED_DOMAIN_S seconds.
+%%
+%% Return {ok, Peer} | {error, Reason}.
 resolve_and_cache_peer(RawPeer, Type) ->
 	case ar_util:safe_parse_peer(RawPeer) of
```
JamesPiechota, line 311: It looks like this will always try to resolve first, then it inspects the cache. Or am I mis-reading?

Shoudl we check the cache before calling `ar_util:safe_pars_peer`?

apps/arweave/src/ar_nonce_limiter_server_worker.erl
```diff
@@ -36,46 +39,43 @@ start_link(Name, RawPeer) ->
...
-handle_info({event, nonce_limiter, _Event}, #state{ peer = undefined } = State) ->
-	{noreply, State};
 handle_info({event, nonce_limiter, {computed_output, Args}}, State) ->
-	#state{ peer = Peer, pause_until = Timestamp, format = Format } = State,
-	{SessionKey, StepNumber, Output, _PartitionUpperBound} = Args,
-	CurrentStepNumber = ar_nonce_limiter:get_current_step_number(),
-	case os:system_time(second) < Timestamp of
-		true ->
+	#state{ raw_peer = RawPeer } = State,
+	case ar_peers:resolve_and_cache_peer(RawPeer, vdf_client_peer) of
```
JamesPiechota, line 72: since we always check the cache and re-resolve here, can we remove the periodic `re_resolve_peer_domain`?

apps/arweave/include/ar_blacklist_middleware.hrl
JamesPiechota: yeah I was trying to think through it. I think your current approach is easiest (vdf3 + vdff4). We'll just try to remember to remove vdf3 sometime in the future...

apps/arweave/src/ar_nonce_limiter_server_worker.erl
```diff
@@ -36,46 +39,43 @@ start_link(Name, RawPeer) ->
 init(RawPeer) ->
 	process_flag(trap_exit, true),
 	ok = ar_events:subscribe(nonce_limiter),
-	gen_server:cast(self(), resolve_raw_peer),
+	case ar_config:is_public_vdf_server() of
```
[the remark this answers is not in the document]
ldmberman: There is a peculiar edge case - if the client peers are configured to pull, they will push our pushes back. We would still try to push every once in a while, however I do not want to make the two timers dependent. I mean, the cache expiration timer and the re-try push timer.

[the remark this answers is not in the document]
JamesPiechota: ok. Not sure I full understand - but that's on me! Resolving.

JamesPiechota (approved):
LGTM!

PR #581, opened 2024-06-01
Add /recent endpoint with recent blocks and forks

JamesPiechota:
The `blocks` field stores the last 18 blocks that have been received including their timestamp. The timestamps of the most recent 5 are hidden and when a miner mines a block it uses the timestamp of the first time that same block is shared  to it. These 2 things are done so that we don't leak information about which blocks a miner has mined.

The `forks` field lists all recently received forks - the forks are persisted to disk as well. A fork is identified by:
`block_ids`: a llist of block ids for all abandoned blocks
`id`: a hash of the `block_ids` that serves as the fork's id
`height`: the height of the first abandoned block
`timestamp`: the timestamp (in seconds) of when the fork was detected

### DIFF
--- a/apps/arweave/src/ar_block_cache.erl
+++ b/apps/arweave/src/ar_block_cache.erl
@@ -8,7 +8,7 @@
 		get_longest_chain_cache/1,
 		get_block_and_status/2, remove/2, get_checkpoint_block/1, prune/2,
 		get_by_solution_hash/5, is_known_solution_hash/2,
-		get_siblings/2, get_fork_blocks/2]).
+		get_siblings/2, get_fork_blocks/2, update_timestamp/3]).
 
 -include_lib("arweave/include/ar.hrl").
 -include_lib("eunit/include/eunit.hrl").
@@ -454,6 +454,29 @@ get_siblings(Tab, B) ->
 				[],
 				Children
 			)
+	end.
+
+update_timestamp(Tab, H, ReceiveTimestamp) ->
+	case ets:lookup(Tab, {block, H}) of
+		[{_, {B, Status, Timestamp, Children}}] ->
+			case B#block.receive_timestamp of
+				undefined ->
+					insert(Tab, {
+							{block, H},
+							{
+								B#block{receive_timestamp = ReceiveTimestamp},
+								Status,
+								Timestamp,
+								Children
+							}
+						}, false);
+				_ ->
+					ok
+			end;
+		[] ->
+			?LOG_ERROR([
+				{event, ignored_block_missing_from_cache}, {block, ar_util:encode(H)}]),
+			not_found
 	end.
 
 %%%===================================================================
--- a/apps/arweave/src/ar_block_cache.erl
+++ b/apps/arweave/src/ar_block_cache.erl
@@ -474,8 +474,6 @@ update_timestamp(Tab, H, ReceiveTimestamp) ->
 					ok
 			end;
 		[] ->
-			?LOG_ERROR([
-				{event, ignored_block_missing_from_cache}, {block, ar_util:encode(H)}]),
 			not_found
 	end.
 
--- a/apps/arweave/src/ar_block_cache.erl
+++ b/apps/arweave/src/ar_block_cache.erl
@@ -215,6 +215,16 @@ get(Tab, H) ->
 			B
 	end.
 
+%% @doc Get the block and its status from cache.
+%% Returns not_found if the block is not in cache.
+get_block_and_status(Tab, H) ->
+	case ets:lookup(Tab, {block, H}) of
+		[] ->
+			not_found;
+		[{_, {B, Status, Timestamp, _Children}}] ->
+			{B, {Status, Timestamp}}
+	end.
+
 %% @doc Get a {block, previous blocks, status} tuple for the earliest block from
 %% the longest chain, which has not been validated yet. The previous blocks are
 %% sorted from newest to oldest. The last one is a block from the current fork.
@@ -296,16 +306,6 @@ tx_id(#tx{ id = ID }) ->
 	ID;
 tx_id(TXID) ->
 	TXID.
-
-%% @doc Get the block and its status from cache.
-%% Returns not_found if the block is not in cache.
-get_block_and_status(Tab, H) ->
-	case ets:lookup(Tab, {block, H}) of
-		[] ->
-			not_found;
-		[{_, {B, Status, Timestamp, _Children}}] ->
-			{B, {Status, Timestamp}}
-	end.
 
 %% @doc Mark the given block as the tip block. Mark the previous blocks as on-chain.
 %% Mark the on-chain blocks from other forks as validated. Raises invalid_tip if
@@ -458,6 +458,8 @@ get_siblings(Tab, B) ->
 
 update_timestamp(Tab, H, ReceiveTimestamp) ->
 	case ets:lookup(Tab, {block, H}) of
+		[] ->
+			not_found;
 		[{_, {B, Status, Timestamp, Children}}] ->
 			case B#block.receive_timestamp of
 				undefined ->
@@ -472,9 +474,7 @@ update_timestamp(Tab, H, ReceiveTimestamp) ->
 						}, false);
 				_ ->
 					ok
-			end;
-		[] ->
-			not_found
+			end
 	end.
 
 %%%===================================================================
--- a/apps/arweave/src/ar_chain_stats.erl
+++ b/apps/arweave/src/ar_chain_stats.erl
@@ -25,26 +25,23 @@ log_fork(Orphans, ForkRootB) ->
 %% @doc Returns all forks that have been logged since the given start time
 %% (system time in seconds)
 get_forks(StartTime) ->
-    gen_server:call(?MODULE, {get_forks, StartTime}).
+    case catch gen_server:call(?MODULE, {get_forks, StartTime}) of
+		{'EXIT', {timeout, {gen_server, call, _}}} ->
+			{error, timeout};
+		Reply ->
+			Reply
+	end.
 
 %%%===================================================================
 %%% Generic server callbacks.
 %%%===================================================================
 init([]) ->
     process_flag(trap_exit, true),
-    Columns = [
-        {"default", []},
-        {"forks", []}
-    ],
-    Names = [
-        ?MODULE,
-        forks
-    ],
-    ok = ar_kv:open(filename:join(?ROCKS_DB_DIR, "ar_chain_stats_db"), Columns, [], Names),
+	ok = ar_kv:open(filename:join(?ROCKS_DB_DIR, "forks_db"), forks_db),
     {ok, #{}}.
 
 handle_call({get_forks, StartTime}, _From, State) ->
-    {ok, ForksMap} = ar_kv:get_range(forks, integer_to_binary(StartTime * 1000)),
+    {ok, ForksMap} = ar_kv:get_range(forks_db, <<(StartTime * 1000):64>>),
     %% Sort forks by their key (the timestamp when they were detected)
     SortedForks = lists:sort(maps:to_list(ForksMap)),
     Forks = [binary_to_term(Fork) || {_Timestamp, Fork} <- SortedForks],
@@ -69,7 +66,7 @@ terminate(_Reason, _state) ->
 %%%===================================================================
 log_fork(Orphans, ForkRootB, ForkTime) ->
     Fork = create_fork(Orphans, ForkRootB, ForkTime),
-    ar_kv:put(forks, integer_to_binary(ForkTime), term_to_binary(Fork)),
+    ar_kv:put(forks_db, <<ForkTime:64>>, term_to_binary(Fork)),
     record_fork_depth(Orphans),
     ok.
 
@@ -100,10 +97,33 @@ record_fork_depth([H | Orphans], N) ->
 %%%===================================================================
 forks_test_() ->
     [
-		{timeout, 30, fun test_forks/0}
+        {timeout, 30, fun test_fork_time/0},
+        {timeout, 30, fun test_forks/0}
 	].
 
+test_fork_time() ->
+    clear_forks_db(),
+    ForkRootB1 = #block{ height = 1 },
+
+    Orphans1 = [<<"a">>],
+    ExpectedFork1 = #fork{
+        id = crypto:hash(sha256, list_to_binary(Orphans1)),
+        height = 2,
+        block_ids = Orphans1
+    },
+    Orphans2 = [<<"b">>, <<"c">>],
+    ExpectedFork2 = #fork{
+        id = crypto:hash(sha256, list_to_binary(Orphans2)),
+        height = 2,
+        block_ids = Orphans2
+    },
+
+    gen_server:cast(?MODULE, {log_fork, Orphans1, ForkRootB1, 2}),
+    gen_server:cast(?MODULE, {log_fork, Orphans2, ForkRootB1, 11}),
+    assert_forks_equal([ExpectedFork1, ExpectedFork2], get_forks(0)).
+
 test_forks() ->
+    clear_forks_db(),
     StartTime = os:system_time(seconds),
     ForkRootB1 = #block{ height = 1 },
     ForkRootB2= #block{ height = 2 },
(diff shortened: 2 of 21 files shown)


apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2358,6 +2321,7 @@ handle_block_announcement(#block_announcement{ indep_hash = H, previous_block =
 		solution_hash = SolutionH }, Req) ->
 	case ar_ignore_registry:member(H) of
 		true ->
+			ar_block_cache:update_timestamp(block_cache, H, erlang:timestamp()),
```
ldmberman, line 2324: Why do we update the timestamp here? We've seen the block before so shouldn't we have the relevant timestamp assigned already?

apps/arweave/src/ar_info.erl
```diff
@@ -0,0 +1,94 @@
...
+get_recent_blocks(CurrentHeight) ->
+    lists:foldl(
+        fun({H, _WeaveSize, _TXRoot}, Acc) ->
+            Acc ++ [#{
+                <<"id">> => ar_util:encode(H),
+                <<"received">> => get_block_timestamp(H, length(Acc))
+            }]
+        end,
+        [],
+        lists:sublist(ar_block_index:get_list(CurrentHeight), ?CHECKPOINT_DEPTH)
```
ldmberman, line 67: ar_block_index:get_list is very slow as it fetches the entire block index - we may use e.g. [ar_node:get_block_anchors](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_node.erl#L183C1-L183C18) to quickly fetch the recent block hashes.

apps/arweave/src/ar_chain_stats.erl
```diff
@@ -0,0 +1,176 @@
...
+    gen_server:call(?MODULE, {get_forks, StartTime}).
+
+%%%===================================================================
+%%% Generic server callbacks.
+%%%===================================================================
+init([]) ->
+    process_flag(trap_exit, true),
+    Columns = [
+        {"default", []},
+        {"forks", []}
```
ldmberman, line 37: Do we need to use a custom RocksDB column? The feature has a bug where the data in the non-default columns would not survive the first non-graceful shutdown (some problems with WAL population, perhaps).

apps/arweave/src/ar_chain_stats.erl
```diff
@@ -0,0 +1,176 @@
...
+log_fork([], _ForkRootB) ->
+    %% No fork to log
+    ok;
+log_fork(Orphans, ForkRootB) ->
+    gen_server:cast(?MODULE, {log_fork, Orphans, ForkRootB, os:system_time(millisecond)}).
+
+%% @doc Returns all forks that have been logged since the given start time
+%% (system time in seconds)
+get_forks(StartTime) ->
+    gen_server:call(?MODULE, {get_forks, StartTime}).
```
ldmberman, line 28: I'd handle the timeout gracefully here, I believe, we'll hit it every once in a while.

apps/arweave/src/ar_info.erl
```diff
@@ -0,0 +1,94 @@
...
+                id = ID, height = Height, timestamp = Timestamp, block_ids = BlockIDs} = Fork,
+            Acc ++ [#{
+                <<"id">> => ar_util:encode(ID),
+                <<"height">> => Height,
+                <<"timestamp">> => Timestamp div 1000,
+                <<"blocks">> => [ ar_util:encode(BlockID) || BlockID <- BlockIDs ]
+            }]
+        end,
+        [],
+        ar_chain_stats:get_forks(0)
```
ldmberman, line 83: Do we fetch all forks here, not just a recent subset?

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -1086,8 +1086,7 @@ get_info(Peer, Type) ->
 	case get_info(Peer) of
 		info_unavailable -> info_unavailable;
 		Info ->
-			{Type, X} = lists:keyfind(Type, 1, Info),
-			X
+			maps:get(atom_to_binary(Type), Info)
```
ldmberman, line 1089: The previous implementation handled the case when an expected key would not be sent (plus the absence of the release field treated as release=0), here we will have badarg, if I see it right.

apps/arweave/src/ar_block_cache.erl
```diff
@@ -456,6 +456,27 @@ get_siblings(Tab, B) ->
 			)
 	end.
 
+update_timestamp(Tab, H, ReceiveTimestamp) ->
+	case ets:lookup(Tab, {block, H}) of
+		[{_, {B, Status, Timestamp, Children}}] ->
+			case B#block.receive_timestamp of
+				undefined ->
+					insert(Tab, {
```
ldmberman, line 464: It's not thread-safe to update this table outside the ar_node_worker process.

apps/arweave/src/ar_chain_stats.erl
JamesPiechota: Oh. Ok. So I should create a separate rocksdb file for each column? My plan is to add a few more columns in the ar_chai_stats module to track some other data needed for separate bugs/requests

ldmberman: I would do one db per table, like we have it e.g., in [ar_storage](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_storage.erl#L1030), do you see major downsides to this approach?

JamesPiechota: Nope! I was just trying to follow conventions and wasn't sure whether ar_data_sync or ar_storage was the "right" way. I'll do one file per table!

apps/arweave/src/ar_http_iface_client.erl
JamesPiechota: what about if I change to `maps:get(atom_to_binary(Type), Info, info_unavailable)`?

I believe that's what happened previously if the expected key was not sent.

In that case the only situation that isn't handled is the special handling for a missing `release`. Which perhaps I can handle in `ar_peers.erl` (the only place where `get_info(Peer, release)` is called?

My motivation in changing this to a map was to cut down on some of the boilerplate that we had where we converted JSON -> JSonStruct -> Map -> value.

Let me build a commit to show you the proposed fix.

JamesPiechota: Proposed fix: https://github.com/ArweaveTeam/arweave/pull/581/commits/319984a85aba444a1492271f22f53ba8010f8898

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: Mostly to handle the fact that we only want to record the timestamp the first time the block is received from a peer so that we don't record the timestamp when the block was first mined. So in the event that this node mined a block it'll be added to the ignore registry even though the received timestamp hasn't been set. Instead we always check the timestamp whenever we receive an announcement or post, and update it if it hasn't already been set.

JamesPiechota: since we do an ETS lookup by a specific key, I'm assuming the performance impact of the extra check whenever we handle a post or announcement is minimal . But perhaps that's a bad assumption?

apps/arweave/src/ar_block_cache.erl
JamesPiechota: ah, oof. Okay I'll rework it

apps/arweave/src/ar_http_iface_middleware.erl
ldmberman: Are there any other places where we call update_timestamp except the two I see here under ar_ignore_registry:member(H) == true? Does it mean we only set the timestamp the second time we receive the block?

apps/arweave/src/ar_http_iface_client.erl
ldmberman: Makes sense, looks good, thank you!

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: We also set the timestamp the first time a block is pre-validated.

The value gets set in the Block here: https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_block_pre_validator.erl#L49

And the block gets cached when the `{block, {new` event is handled, sent here: https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_block_pre_validator.erl#L746

JamesPiechota: ugh, I suppose I also need to handle the first time that a node fetches a block it had previously mined: https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_poller.erl#L146

JamesPiechota: Basically: it's the handling of self-mined blocks that causes all the trouble since we *don't* want the timestamp to be set to the momen the block was mined since that will make it possible to determine which nodes mined which blocks. And near as I can tell there are 3 places were we receive a block from a peer and need to check the timestamp:
- block announcement
- POST block
- GET block via polling
-

JamesPiechota: Atually never mind, I don't think polling is relevant for this case. If the node has already mined the block it won't poll it from a peer, and if it hasn't mined the block then the timestamp will be handled via pre-validation when it polls it.

apps/arweave/src/ar_block_cache.erl
JamesPiechota: I moved the update into ar_node_worker: https://github.com/ArweaveTeam/arweave/pull/581/commits/e6b2dd18b55d769aa62d2bfb187db946c5d3b48c

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: So yeah, as far as I can tell:

Previous behavior:
1. receive_timestamp is *not* set when a node mines a block
2. receive_timestamp is set when a newly received block passes pre-validation

Desired behavior:
1. always set receive_timestamp the first time a node receives a block from a peer
2. To do this we need to handle the case where a block exists in a node's block_cache because it was the one who mined it. In this case we need to update the receive_timestamp when it receives the block either via `POST /block` or `POST /block_announcement` even though that block make already exist in the ar_ignore_registry

Let me know if you I've missed something or you think it should be implemented differently

apps/arweave/src/ar_info.erl
JamesPiechota: Now we do! Update to fetch just a recent subset: https://github.com/ArweaveTeam/arweave/pull/581/commits/b63151dc1abd41661302bb5fce2f810f7a5cd7dc

apps/arweave/src/ar_chain_stats.erl
JamesPiechota: Updated: https://github.com/ArweaveTeam/arweave/pull/581/commits/b63151dc1abd41661302bb5fce2f810f7a5cd7dc

apps/arweave/src/ar_chain_stats.erl
JamesPiechota: Updated: https://github.com/ArweaveTeam/arweave/pull/581/commits/b63151dc1abd41661302bb5fce2f810f7a5cd7dc

apps/arweave/src/ar_info.erl
JamesPiechota: I may update this logic again when I switch to using get_block_anchors - but I'll do that in a separate commit

apps/arweave/src/ar_chain_stats.erl
```diff
@@ -0,0 +1,173 @@
...
+
+terminate(_Reason, _state) ->
+    ok.
+
+%%%===================================================================
+%%% Private functions.
+%%%===================================================================
+log_fork(Orphans, ForkRootB, ForkTime) ->
+    Fork = create_fork(Orphans, ForkRootB, ForkTime),
+    ar_kv:put(forks_db, integer_to_binary(ForkTime), term_to_binary(Fork)),
```
ldmberman, line 69: integer_to_binary(2) > integer_to_binary(11); I think, we need << ForkTime:64 >> here

apps/arweave/src/ar_info.erl
```diff
@@ -0,0 +1,99 @@
...
+        %%   "received": <received_timestamp>"
+        %% }
+        <<"blocks">> => get_recent_blocks(ar_node:get_height()),
+        %% #{
+        %%   "id": <hash_of_block_ids>,
+        %%   "height": <height_of_first_orphaned_block>,
+        %%   "timestamp": <timestamp_of_when_fork_was_abandoned>
+        %%   "blocks": [<block_id>, <block_id>, ...]
+        %% }
+        <<"forks">> => get_recent_forks()
```
ldmberman, line 55: The {error, _} response will still crash the handler process

JamesPiechota: Got it. I haven't written the tests for that bit yet. I'll do that next and then fix.

JamesPiechota: I think I got it this time? https://github.com/ArweaveTeam/arweave/pull/581/commits/0542b4e1254d1d50b751c3b1589fb995e9b7f3a8

apps/arweave/src/ar_chain_stats.erl
JamesPiechota: Yikes good catch. I added a test and fixex: https://github.com/ArweaveTeam/arweave/pull/581/commits/0542b4e1254d1d50b751c3b1589fb995e9b7f3a8

apps/arweave/src/ar_info.erl
JamesPiechota: Updated: https://github.com/ArweaveTeam/arweave/pull/581/commits/4764260fed3c794b8e5c17574edb43476ae51717

apps/arweave/src/ar_info.erl
```diff
@@ -0,0 +1,100 @@
...
+    CutOffTime = os:system_time(seconds) - ?RECENT_FORKS_AGE,
+    case ar_chain_stats:get_forks(CutOffTime) of
+        {error, _} -> error;
+        Forks ->
+            lists:foldl(
+                fun(Fork, Acc) ->
+                    #fork{ 
+                        id = ID, height = Height, timestamp = Timestamp, 
+                        block_ids = BlockIDs} = Fork,
+                    Acc ++ [#{
```
ldmberman, line 80: 20 times is not a lot, but I would avoid some unnecessary reductions here (e.g., might matter slightly under load) and prepend `[#{...} | Acc]`, with lists:reverse if we want this ordering.

apps/arweave/src/ar_node.erl
```diff
@@ -181,8 +181,12 @@ get_current_usd_to_ar_rate() ->
 %% the hashes of the recent blocks that can be used in transactions as anchors.
 %% @end
 get_block_anchors() ->
-	[{block_anchors, BlockAnchors}] = ets:lookup(node_state, block_anchors),
```
ldmberman, line 184: Could we return an atom here (e.g., not_joined) to avoid bugs?

ldmberman: Hm, I think, it makes sense to return 403 not_joined from the new endpoint until the node has been initialized, like in most other endpoints?

apps/arweave/src/ar_http_iface_middleware.erl
ldmberman: Ah, yes, I remember now, thank you for the detailed explanation!

apps/arweave/src/ar_node.erl
JamesPiechota: Updated! https://github.com/ArweaveTeam/arweave/pull/581/commits/9e18942770c521b4dcb50e31cc00641a2cdea790

apps/arweave/src/ar_info.erl
JamesPiechota: Updated! https://github.com/ArweaveTeam/arweave/pull/581/commits/9e18942770c521b4dcb50e31cc00641a2cdea790

JamesPiechota:
@ldmberman Added fix to your comment about recent blocks ordering (I can't find the comment anymore though...). https://github.com/ArweaveTeam/arweave/pull/581/commits/1a2d357299ca23b2a4753f272eec9589e0a16581

6th time's the charm ;)

ldmberman (approved):
LGTM, thank you!

PR #590, opened 2024-07-16
Feature/packing difficulty

ldmberman:


apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+		unsigned char* subChunk = paddedChunk + offset;
+		unsigned char* encryptedSubChunk = (unsigned char*)malloc(subChunkSize);
+
+		// 3 bytes is sufficient to represent offsets up to at most MAX_CHUNK_SIZE.
+		int offsetByteSize = 3;
+		unsigned char offsetBytes[offsetByteSize];
+		for (int k = 0; k < offsetByteSize; k++) {
+			offsetBytes[k] = ((offset + subChunkSize) >> (8 * (offsetByteSize - 1 - k))) & 0xFF;
+		}
+		SHA256_CTX sha256;
```
JamesPiechota, line 388: Is this a correct understanding of this block?
1. It computes a feistel hash which is used as the encryption key for the sub-chunk
2. the feistel hash is generated from the packing address (i.e. `inputData`) and the offset of the sub-chunk
3. this guarantees that the encryption key for each sub-chunk is unique but deterministic

If that's correct maybe we could add something like that as a comment? e.g.

```
// Sub-chunk encryption key is the feistel hash of the input data and the sub-chunk offset
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+	uint32_t offset = 0;
+	unsigned char key[PACKING_KEY_SIZE];
+	for (int i = 0; i < subChunkCount; i++) {
+		unsigned char* subChunk = paddedChunk + offset;
+		unsigned char* encryptedSubChunk = (unsigned char*)malloc(subChunkSize);
+
+		// 3 bytes is sufficient to represent offsets up to at most MAX_CHUNK_SIZE.
+		int offsetByteSize = 3;
+		unsigned char offsetBytes[offsetByteSize];
+		for (int k = 0; k < offsetByteSize; k++) {
```
JamesPiechota, line 385:
```suggestion
		// Byte string representation of the sub-chunk offset: i * subChunkSize
		for (int k = 0; k < offsetByteSize; k++) {
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+			offsetBytes[k] = ((offset + subChunkSize) >> (8 * (offsetByteSize - 1 - k))) & 0xFF;
+		}
+		SHA256_CTX sha256;
+		SHA256_Init(&sha256);
+		SHA256_Update(&sha256, inputDataPtr->data, inputDataPtr->size);
+		SHA256_Update(&sha256, offsetBytes, offsetByteSize);
+		SHA256_Final(key, &sha256);
+
+		for (int j = 0; j < iterations; j++) {
```
JamesPiechota, line 394:
```suggestion
		// Sequentially encrypt each sub-chunk 'iterations' times.
		// The encrypted output of each iteration is the input for the following iteration.
		for (int j = 0; j < iterations; j++) {
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+
+	ERL_NIF_TERM encryptedChunkTerm;
+	unsigned char* encryptedChunk = enif_make_new_binary(envPtr, MAX_CHUNK_SIZE,
+			&encryptedChunkTerm);
+	// Both MAX_CHUNK_SIZE and subChunkCount are multiples of 64 so all sub-chunks
+	// are of the same size.
+	uint32_t subChunkSize = MAX_CHUNK_SIZE / subChunkCount;
+	uint32_t offset = 0;
+	unsigned char key[PACKING_KEY_SIZE];
+	for (int i = 0; i < subChunkCount; i++) {
```
JamesPiechota, line 378:
```suggestion
	// Encrypt each sub-chunk independently and then concatenate the encrypted sub-chunks to yield encrypted composite chunk
	for (int i = 0; i < subChunkCount; i++) {
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+	ERL_NIF_TERM decryptedChunkTerm;
+	unsigned char* decryptedChunk = enif_make_new_binary(envPtr, outChunkLen,
+			&decryptedChunkTerm);
+	unsigned char* decryptedSubChunk;
+	// Both MAX_CHUNK_SIZE and subChunkCount are multiples of 64 so all sub-chunks
+	// are of the same size.
+	uint32_t subChunkSize = outChunkLen / subChunkCount;
+	uint32_t offset = 0;
+	unsigned char key[PACKING_KEY_SIZE];
+	for (int i = 0; i < subChunkCount; i++) {
```
JamesPiechota, line 425:
```suggestion
	// Decrypt each sub-chunk independently and then concatenate the decrypted sub-chunks to yield encrypted composite chunk
	for (int i = 0; i < subChunkCount; i++) {
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+		randomx_vm *vmPtr, ErlNifBinary *inputDataPtr, ErlNifBinary *inputChunkPtr,
+		const int outChunkLen, const int subChunkCount, const int iterations,
+		const int randomxRoundCount, const int jitEnabled,
+		const int largePagesEnabled, const int hardwareAESEnabled) {
+
+	ERL_NIF_TERM decryptedChunkTerm;
+	unsigned char* decryptedChunk = enif_make_new_binary(envPtr, outChunkLen,
+			&decryptedChunkTerm);
+	unsigned char* decryptedSubChunk;
+	// Both MAX_CHUNK_SIZE and subChunkCount are multiples of 64 so all sub-chunks
```
JamesPiechota, line 420:
```suggestion
	// Both outChunkLen and subChunkCount are multiples of 64 so all sub-chunks
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -910,7 +1026,463 @@ static ERL_NIF_TERM randomx_reencrypt_chunk_nif(
...
+		return enif_make_badarg(envPtr);
+	}
+	if (!enif_get_int(envPtr, argv[6], &randomxRoundCount)) {
+		return enif_make_badarg(envPtr);
+	}
+	if (!enif_get_int(envPtr, argv[7], &iterations) ||
+		iterations < 1) {
+		return enif_make_badarg(envPtr);
+	}
+	if (!enif_get_int(envPtr, argv[8], &subChunkCount) ||
```
JamesPiechota, line 1075: Do we also need to test that `subChunkCount` is a multiple of 64? e.g. I believe a subChunkCount of 2 would pass these checks but would then run afoul of this comment `// Both MAX_CHUNK_SIZE and subChunkCount are multiples of 64`

JamesPiechota: Or, maybe that comment just needs to be updated - since we don't care if subChunkCount is a multiple of 64 so long as all sub chunks are the same size and the sub-chunks are multiple of 64

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+		unsigned char* subChunk = inputChunkPtr->data + offset;
+		decryptedSubChunk = (unsigned char*)malloc(subChunkSize);
+
+		// 3 bytes is sufficient to represent offsets up to at most MAX_CHUNK_SIZE.
+		int offsetByteSize = 3;
+		unsigned char offsetBytes[offsetByteSize];
+		for (int k = 0; k < offsetByteSize; k++) {
+			offsetBytes[k] = ((offset + subChunkSize) >> (8 * (offsetByteSize - 1 - k))) & 0xFF;
+		}
+		SHA256_CTX sha256;
```
JamesPiechota, line 435:
```suggestion
		// Sub-chunk encryption key is the feistel hash of the input data and the sub-chunk offset
		SHA256_CTX sha256;
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+		SHA256_Init(&sha256);
+		SHA256_Update(&sha256, inputDataPtr->data, inputDataPtr->size);
+		SHA256_Update(&sha256, offsetBytes, offsetByteSize);
+		SHA256_Final(key, &sha256);
+
+		for (int j = 0; j < iterations; j++) {
```
JamesPiechota, line 441:
```suggestion
		// Sequentially encrypt each sub-chunk 'iterations' times.
		// The encrypted output of each iteration is the input for the following iteration.
		for (int j = 0; j < iterations; j++) {
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -337,6 +353,106 @@ static ERL_NIF_TERM encrypt_chunk(ErlNifEnv* envPtr,
...
+	uint32_t offset = 0;
+	unsigned char key[PACKING_KEY_SIZE];
+	for (int i = 0; i < subChunkCount; i++) {
+		unsigned char* subChunk = inputChunkPtr->data + offset;
+		decryptedSubChunk = (unsigned char*)malloc(subChunkSize);
+
+		// 3 bytes is sufficient to represent offsets up to at most MAX_CHUNK_SIZE.
+		int offsetByteSize = 3;
+		unsigned char offsetBytes[offsetByteSize];
+		for (int k = 0; k < offsetByteSize; k++) {
```
JamesPiechota, line 432:
```suggestion
		// Byte string representation of the sub-chunk offset: i * subChunkSize
		for (int k = 0; k < offsetByteSize; k++) {
```

apps/arweave/c_src/ar_mine_randomx.c
```diff
@@ -910,7 +1026,463 @@ static ERL_NIF_TERM randomx_reencrypt_chunk_nif(
...
+			memcpy(subChunk, decryptedSubChunk, subChunkSize);
+		}
+	}
+	free(subChunk);
+	destroy_vm(statePtr, vmPtr);
+
+	return ok_tuple(envPtr, decryptedSubChunkTerm);
+}
+
+static ERL_NIF_TERM randomx_reencrypt_legacy_to_composite_chunk_nif(
```
JamesPiechota, line 1279: The `reencrypt` code that I added back in April 2023 was a mistake, IMO. I'd added it under the assumption that doing the decrypt+encrypt in a single NIF call would improve performance. But once implemented and I could run performance tests the impact was negligible / within the error tolerance of the benchmark: https://digitalhistor-8tx7838.slack.com/archives/C04U7EH3ETF/p1680824283020639

Given that maintaining reencrypt is adding several permutations to the code paths with no real performance gains, my suggestion would be to *not* add them in the new code and instead just use decrypt+encrypt. It will also mean fewer code paths that we have to test, validate, and maintain.

If we still want to go ahead with adding separate reencrypt calls, it might be worth refactoring the code paths to reduce duplication.

apps/arweave/include/ar.hrl
```diff
@@ -310,6 +310,23 @@
 %% Maximum size of a single data chunk, in bytes.
 -define(DATA_CHUNK_SIZE, (256 * 1024)).
 
+%% The maximum allowed packing difficulty.
+-define(MAX_PACKING_DIFFICULTY, 32).
+
+%% The number of sub-chunks in a compositely packed chunk with the packing difficulty
+%% between 1 and 32 incl. The composite packing with the packing difficulty 1 matches
+%% approximately the non-composite 2.6 packing in terms of computational costs.
+-define(PACKING_DIFFICULTY_ONE_SUB_CHUNK_COUNT, 32).
```
JamesPiechota, line 319: The number of sub-chunks is the same for every packing difficulty, right? i.e. this isn't specifically about "Packing Difficulty 1" right? Or have I misunderstood?

I.e. could this be called `COMPOSITE_PACKING_SUB_CHUNK_COUNT` or something and still be accurate?

JamesPiechota: Hm, okay I see elsewhere we use the term "atomic sub-chunk" or "basic atomic sub-chunk" - are those different from a "packing difficulty 1 sub chunk"?

if they're all the same thing, what do you think about just labeling them all "sub chunk" and dropping the qualifiers - I don't think we have the concept of a sub-chunk anywhere else in the code yet, right?

apps/arweave/src/ar.erl
```diff
@@ -98,12 +98,13 @@ show_help() ->
 					"a particular data range. The data and metadata related to the module "
 					"are stored in a dedicated folder "
 					"([data_dir]/storage_modules/storage_module_[partition_number]_[packing]/"
-					") where packing is either a mining address or \"unpacked\"."
-					" Example: storage_module 0,En2eqsVJARnTVOSh723PBXAKGmKgrGSjQ2YIGwE_ZRI. "
+					") where packing is either a mining address + : + packing difficulty "
```
JamesPiechota, line 101:
```suggestion
					") where packing is either <mining address>:<packing difficulty> "
```

apps/arweave/include/ar.hrl
```diff
@@ -310,6 +310,23 @@
...
+
+%% The size of a unit sub-chunk in the compositely packed chunk with the packing difficulty
+%% between 1 and 32 incl.
+-define(PACKING_DIFFICULTY_ONE_SUB_CHUNK_SIZE,
+		(?DATA_CHUNK_SIZE div ?PACKING_DIFFICULTY_ONE_SUB_CHUNK_COUNT)).
+
+%% The number of RandomX rounds used for a single iteration of packing of a single sub-chunk
+%% during the composite packing with the packing difficulty between 1 and 32 incl.
+-define(PACKING_DIFFICULTY_ONE_ROUND_COUNT, 10).
```
JamesPiechota, line 328: Why do we use 10 here, but down [here](https://github.com/ArweaveTeam/arweave/pull/590/files#diff-f39d3764650cee3f08a44af8852d72bb770d091e7cdfb3c7539efa092b030569R19-R25) we use 8?

apps/arweave/src/ar_bench_packing.erl
```diff
@@ -13,7 +13,9 @@ run_benchmark_from_cli(Args) ->
 	LargePages = list_to_integer(get_flag_value(Args, "large_pages", "1")),
 	HardwareAES = list_to_integer(get_flag_value(Args, "hw_aes", "1")),
 	VDF = list_to_integer(get_flag_value(Args, "vdf", "0")),
-	run_benchmark(Test, JIT, LargePages, HardwareAES, VDF).
+	Iterations = list_to_integer(get_flag_value(Args, "iterations", "1")),
+	Rounds = list_to_integer(get_flag_value(Args, "rounds", "8")),
```
JamesPiechota, line 17: Should the default rounds be 10? 

Or actually maybe [it is 8](https://github.com/ArweaveTeam/arweave/pull/590/files#r1706283377)? I think I'm missing something

apps/arweave/c_src/ar_mine_randomx.c
ldmberman: It is a sha2 hash (not a feistel)

apps/arweave/c_src/ar_mine_randomx.c
ldmberman: Hm, yeah, good question. I see you had 1-2% gain there which does not sound like a lot but considering how much packing miners need to do overall and how expensive packing is, I'd say it's worth it. Although the higher the packing difficulty, the lower the gain will be.

apps/arweave/src/ar_bench_packing.erl
ldmberman: We need to fix it here.

apps/arweave/c_src/ar_mine_randomx.c
JamesPiechota: haha, my ignorance is showing. I was going off of pattern matching with the `feistel_hash()` function: https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/c_src/feistel_msgsize_key_cipher.cpp#L7-L13

The code blocks look pretty similar - is the main difference that a feistel hash forces a 32-length? (or is it just a misnaming of the `feistel_hash(...)` function and I should ignore?)

apps/arweave/c_src/ar_mine_randomx.c
JamesPiechota: Got it. Let me revive the benchmark and re-run the tests. If gains still hold I can look into refacoring the duplicate code paths. If my earlier reported gains were wrong, I can pull out the reencrypt implementations.

apps/arweave/src/ar_bench_packing.erl
JamesPiechota: Okay great. I'll look at cleaning up the benchmarks as part of re-running the re-encrypt tests.

apps/arweave/include/ar.hrl
```diff
@@ -359,7 +376,8 @@
 	%% of the corresponding transaction. Proofs the inclusion of the chunk
 	%% in the corresponding "data_root" under a particular offset.
 	data_path = <<>>,
-	chunk = <<>>
+	chunk = <<>>,
```
JamesPiechota, line 379:
```suggestion
	%% When packing difficulty is 0 chunk stores a full packed chunk
	%% When packing difficulty >= 1, chunk stores a 8192-byte packed sub-chunk
	chunk = <<>>,
```

apps/arweave/include/ar.hrl
```diff
@@ -359,7 +376,8 @@
...
+	unpacked_chunk = <<>>
```
JamesPiechota, line 380:
```suggestion
	%% When packing difficulty is 0, unpacked_chunk remains <<>>
	%% When packing difficulty >= 1, unpacked_chunk stores a full unpacked chunk
	unpacked_chunk = <<>>
```

apps/arweave/src/ar_block.erl
```diff
@@ -526,7 +605,9 @@ hash_wallet_list(WalletList) ->
 						(ar_serialize:encode_bin(LastTX, 8))/binary,
 						(ar_serialize:encode_int(Denomination, 8))/binary,
 						MiningPermissionBin/binary >>,
-				crypto:hash(sha384, Preimage)
+				Alg = case Height >= ar_fork:height_2_8() of true -> sha256;
```
JamesPiechota, line 608: I forget the context on this change. Is this required for the packing depth changes, or we taking the opportunity to make this change since we have to do a hard fork anyways?

And the rationale is to reduce the size of the wallet list?

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -243,21 +239,77 @@ may_be_pre_validate_second_chunk_hash(#block{ recall_byte2 = undefined } = B, Pr
...
+		#block{ packing_difficulty = PackingDifficulty } = B, PrevB, Peer)
+			when PackingDifficulty >= 1 ->
+	PoA = B#block.poa,
+	case crypto:hash(sha256, PoA#poa.unpacked_chunk) == B#block.unpacked_chunk_hash
+			%% The unpacked chunk is expected to be 0-padded when smaller than
+			%% ?DATA_CHUNK_SIZE.
+			andalso byte_size(PoA#poa.unpacked_chunk) == ?DATA_CHUNK_SIZE of
+		false ->
+			post_block_reject_warn(B, check_first_unpacked_chunk, Peer),
+			ar_events:send(block, {rejected, invalid_first_unpacked_chunk,
```
JamesPiechota, line 264: Do we need to add `invalid_first_unpacked_chunk` to the `ar_peers.erl` `?BLOC_REJECTION_IGNORE` list?

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -243,21 +239,77 @@ may_be_pre_validate_second_chunk_hash(#block{ recall_byte2 = undefined } = B, Pr
...
+			may_be_pre_validate_second_unpacked_chunk_hash(B, PrevB, Peer)
+	end;
+may_be_pre_validate_first_unpacked_chunk_hash(B, PrevB, Peer) ->
+	#block{ poa = PoA, poa2 = PoA2 } = B,
+	#block{ unpacked_chunk_hash = UnpackedChunkHash,
+			unpacked_chunk2_hash = UnpackedChunk2Hash } = B,
+	case {UnpackedChunkHash, UnpackedChunk2Hash} == {undefined, undefined} of
+		false ->
+			post_block_reject_warn(B, check_first_unpacked_chunk_hash, Peer),
+			ar_events:send(block, {rejected, invalid_first_unpacked_chunk_hash,
```
JamesPiechota, line 277: Dito: Do we need to add `invalid_first_unpacked_chunk_hash` to the `ar_peers.erl` `?BLOC_REJECTION_IGNORE` list?

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -243,21 +239,77 @@ may_be_pre_validate_second_chunk_hash(#block{ recall_byte2 = undefined } = B, Pr
...
+		#block{ recall_byte2 = RecallByte2 } = B, PrevB, Peer)
+			when RecallByte2 /= undefined ->
+	PoA2 = B#block.poa2,
+	case crypto:hash(sha256, PoA2#poa.unpacked_chunk) == B#block.unpacked_chunk2_hash
+			%% The unpacked chunk is expected to be 0-padded when smaller than
+			%% ?DATA_CHUNK_SIZE.
+			andalso byte_size(PoA2#poa.unpacked_chunk) == ?DATA_CHUNK_SIZE of
+		false ->
+			post_block_reject_warn(B, check_second_unpacked_chunk, Peer),
+			ar_events:send(block, {rejected, invalid_second_unpacked_chunk,
```
JamesPiechota, line 296: Dito

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -243,21 +239,77 @@ may_be_pre_validate_second_chunk_hash(#block{ recall_byte2 = undefined } = B, Pr
...
+			invalid;
 		true ->
 			pre_validate_indep_hash(B, PrevB, Peer)
+	end;
+may_be_pre_validate_second_unpacked_chunk_hash(B, PrevB, Peer) ->
+	#block{ poa2 = PoA2 } = B,
+	case B#block.unpacked_chunk2_hash == undefined of
+		false ->
+			post_block_reject_warn(B, check_second_unpacked_chunk_hash, Peer),
+			ar_events:send(block, {rejected, invalid_second_unpacked_chunk_hash,
```
JamesPiechota, line 307: Ditto

apps/arweave/src/ar_block_pre_validator.erl
JamesPiechota: Oops, I hadn't refreshed. I see it now!

apps/arweave/src/ar_block_pre_validator.erl
JamesPiechota: Oops, I hadn't refreshed. I see it now!

apps/arweave/src/ar_block_pre_validator.erl
JamesPiechota: Oops, I hadn't refreshed. I see it now!

apps/arweave/src/ar_block_pre_validator.erl
JamesPiechota: Oops, I hadn't refreshed. I see it now!

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -1219,7 +1222,22 @@ handle_cm_partition_table_response({ok, {{<<"200">>, _}, _, Body, _, _}}) ->
...
+									ar_util:decode(EncodedAddr),
+									0
+								},
+								{ok, [DecodedPartition | Acc]};
+							{[
+								{<<"bucket">>, Bucket},
+								{<<"bucketsize">>, BucketSize},
+								{<<"addr">>, EncodedAddr},
+								{<<"pdiff">>, PackingDifficulty}
+							]} when is_integer(PackingDifficulty), PackingDifficulty >= 0,
```
JamesPiechota, line 1234: Should this check >= 1? It looks like we only serialize to `{pdiff, xxx}` if the difficlty is >= 1, so if we get a 0 here it might indicate some weird error?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2332,62 +2344,22 @@ handle_block_announcement(#block_announcement{ indep_hash = H, previous_block =
 			case ar_node:get_block_shadow_from_cache(PrevH) of
 				not_found ->
 					{412, #{}, <<>>, Req};
-				#block{ height = Height } ->
+				#block{} ->
 					Indices = collect_missing_tx_indices(Prefixes),
-					IsSolutionHashKnown =
```
JamesPiechota, line 2337: Why are we removing all this code?

apps/arweave/src/ar_block.erl
```diff
@@ -484,6 +524,45 @@ get_recall_range(H0, PartitionNumber, PartitionUpperBound) ->
...
+			PackingDifficulty == 0
+	end.
+
+validate_packing_difficulty(PackingDifficulty) ->
+	PackingDifficulty >= 0 andalso PackingDifficulty =< ?MAX_PACKING_DIFFICULTY.
+
+get_max_nonce(0) ->
+	get_max_nonce2((?RECALL_RANGE_SIZE) div ?DATA_CHUNK_SIZE);
+get_max_nonce(PackingDifficulty) ->
+	AdjustedRecallRangeSize = ?RECALL_RANGE_SIZE div PackingDifficulty,
```
JamesPiechota, line 549:
```suggestion
	AdjustedRecallRangeSize = get_recall_range(PackingDifficulty),
```

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -510,17 +566,29 @@ pre_validate_cumulative_difficulty(B, PrevB, SolutionResigned, Peer) ->
 					gen_server:cast(?MODULE, {enqueue, {B, PrevB, true, Peer}}),
 					enqueued;
 				false ->
-					pre_validate_quick_pow(B, PrevB, false, Peer)
+					pre_validate_packing_difficulty(B, PrevB, false, Peer)
```
JamesPiechota, line 569: Is this ever called with `SolutionResigned == true`?

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -34,6 +34,9 @@
 
 encode_packing({spora_2_6, Addr}) ->
 	"spora_2_6_" ++ binary_to_list(ar_util:encode(Addr));
+encode_packing({composite, Addr, PackingDifficulty}) ->
+	"composite_" ++ binary_to_list(ar_util:encode(Addr)) ++ ":"
```
JamesPiechota, line 38: Flagging this discussion here (was started in the other repo). @vird pointed out that `:` is used for address checksum and it will cause a problem on windows (eventually). So perhaps we can select a different delimeter?

Slack discussion here: https://digitalhistor-8tx7838.slack.com/archives/C04Q93X02AJ/p1723576491561449

Maybe we use `.` as the delimiter?

apps/arweave/c_src/ar_mine_randomx.c
ldmberman: The block computes sha256(<< packing_key_without_sub_chunk_offset, sub_chunk_offset >>).

apps/arweave/include/ar.hrl
ldmberman: The latter (8) is not used, I removed it.

apps/arweave/src/ar_chunk_storage.erl
ldmberman: I like `.`!

apps/arweave/src/ar_http_iface_middleware.erl
ldmberman: This one might be questionable, but I felt like this feature is more trouble maintaining than it is actually useful. The feature is about the validators optionally informing the block sender they have the chunks so that some bandwidth is saved while the validator reads and packs the chunk(s) themselves.

Thinking about it now, I believe the feature might be in fact hurtful in the long run especially given the composite packing specifics because it may make miners rely excessively on validators extracting the chunks and if such validators disappear over time the network might not notice it and might not react sufficiently quickly. In other words, I would rather make the network spend some bandwidth but require all miners to set up the chunk building correctly from the start.

apps/arweave/src/ar_block_pre_validator.erl
ldmberman: Oh, good catch! This is wrong, fixed

apps/arweave/c_src/ar_mine_randomx.c
JamesPiechota: Re-running the benchmarks showed about a 0.5% peformance improvement for reencrypt. We'll keep the reencrypt functions for now, but I'll remove `randomx_reencrypt_legacy_to_composite_chunk_nif` since we'll be using a different randomx library for legacy and composite which makes it hard to put both in the same nif

PR #596, opened 2024-07-29
Mining performance optimizations

JamesPiechota:
With this PR we've confirmed that a single-node miner can mine the full weave (56 partitions as of today) at about 95% efficiency. We ran the tests on a 128c EPYC with SMT disabled, but Arweave only ran at 17-25% CPU utilization - this implies a much smaller CPU might also be able to mine the full weave.

When running on a large core CPU it's important to limit some of the threading. This can be done by using the `hashing_threads`, `max_vdf_validation_thread_count`, `max_vdf_last_step_validation_thread_count`, and `packing_rate` flags or by passing a flag directly to the Erlang VM 

An example using the application flags:
```
./bin/start hashing_threads 24 max_vdf_validation_thread_count 4 max_vdf_last_step_validation_thread_count 4 packing_rate 100 <other options>
```

An example using the VM flag:
```
./bin/start +S 24:24 -- <other options>
```

Note: the above restrictions may nor may not be needed - and the values used may change. You'll want to run tests on your miner to find the best options to use. The above examples were used to mostly restrict the 128-core CPU to uses 24-cores for hashing. without this restriction the thread contention from 128 threads hurt performance.

Main changes in this PR:
- Expose the Erlang VM flags from the command line. All options before a `--` separator are passed to the Erlang VM, all arguments after it are passed to Arweave. If there is no `--` then all arguments are passed to Arweave.
- Aggregate updates to `ar_mining_stats` rather than incrementing the counters with each operation. At 56 partitions the constant incrementing created an ets table bottleneck which hurt performance. Now hash and chunk updates are aggregated so that far fewer ETS operations are needed.
- Bug fix in the caching of `ar_chunk_storage` file handles. Prior to this fix the chunk storage files were opened and closed on every read because the key used in the cache differed from the key used in the lookup.
- Fix `data-doctor bench` which had broken in an earlier release
- Improve the node shutdown time by removing a call to `process_flag(trap_exit, true),`
- Slight tweak to the Ideal read/hash rate reported in the mining performance report - previously the tip partition was included in the ideal rate even though it is not minable.

### DIFF
--- a/apps/arweave/src/ar_block.erl
+++ b/apps/arweave/src/ar_block.erl
@@ -3,7 +3,7 @@
 -export([block_field_size_limit/1, verify_timestamp/2,
 		get_max_timestamp_deviation/0, verify_last_retarget/2, verify_weave_size/3,
 		verify_cumulative_diff/2, verify_block_hash_list_merkle/2, compute_hash_list_merkle/1,
-		compute_h0/4, compute_h1/3, compute_h2/3, compute_solution_h/2,
+		compute_h0/4, compute_h0/5, compute_h1/3, compute_h2/3, compute_solution_h/2,
 		indep_hash/1, indep_hash/2, indep_hash2/2,
 		generate_signed_hash/1, verify_signature/3,
 		generate_block_data_segment/1, generate_block_data_segment/2,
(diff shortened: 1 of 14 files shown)


ldmberman (approved):
Great work!

kslazarev:
Great PR. Improve hashrate +5 % on 3-node coordinate mining instance (data size - 167,1 TiB - 15.5 kH/s)

kslazarev:
@JamesPiechota "128c EPYC with SMT disabled" which EPYC model? 
I want to compare with consumer models (3900x, 5900x, zen2 threadrippers / epycs).

https://docs.arweave.org/developers/mining/hardware#hashing
This page needs to update with this PR?

JamesPiechota:
@kslazarev your best bet would be to ask in the discord channel: https://discord.com/invite/SMqeXW3ck7

User Vidiot configured the node and has offered to answer questions

JamesPiechota:
And if you do get some more benchmark results I'd love to hear them!

PR #610, opened 2024-08-27
Preparation to bump the RandomX dataset size to 4096MiB

JamesPiechota:
1. Add a bunch of tests for the randomx nifs as well as the nif wrappers
2. Remove a bunch of obsolete C code
3. remove the `reencrypt_legacy_composite` NIF since we won't be able to do that across RandomX dataset sizes
4. Add support for Light vs. Fast mode for all randomx NIFs
5. Have the difference ar_mine_randomx modes handled via the state (e.g. non-debug vs. debug, light vs. fast). This will make it easier later when we add a 3rd mode (512MiB vs. 4096MiB)

### DIFF
--- a/apps/arweave/c_src/Makefile
+++ b/apps/arweave/c_src/Makefile
@@ -75,12 +75,7 @@ link_verbose_0 = @echo " LD    " $(@F);
 link_verbose = $(link_verbose_$(V))
 
 ALL_LIB_SOURCES := $(wildcard $(C_SRC_DIR)/*.c $(C_SRC_DIR)/*.cpp)
-ENIF_SOURCES := $(wildcard $(C_SRC_DIR)/ar_*.c $(C_SRC_DIR)/ar_*.cpp)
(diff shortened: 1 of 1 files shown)


apps/arweave/src/ar_randomx_state.erl
```diff
@@ -1,6 +1,6 @@
 -module(ar_randomx_state).
 
--export([init/2, init/4, start/0, start_block_polling/0, reset/0, hash/2,
+-export([init/2, init/4, start/0, start_block_polling/0, reset/0,
```
ldmberman, line 3: I would remove this module completely because we do not need it any more. The `ar_randomx_state:get_key_block` query in `ar_http_iface_middleware` is totally unnecessary and easy to remove too.

JamesPiechota: Sounds good! Can I do it in the next PR?

ldmberman (approved):

apps/arweave/src/ar_randomx_state.erl
ldmberman: Off course

PR #614, opened 2024-09-06
Increase RandomX Dataset size to 4GiB

JamesPiechota:
Increase the RandomX dataset size to 4GiB.

The main motivation for this change is to increase the packing process' ASIC-resistance. There is minimal if any impact on CPU performance. With this change Arweave nodes will need to allocate 4GiB of memory just for the RandomX dataset, however for most miners this a small fraction of their available RAM. For lightweight validator nodes there is an option to run RandomX in "light" mode which allocates a much smaller cache and computes the dataset on demand.

Main Changes:
- We now build 2 versions of the randomx static library, one with the old 512MiB dataset and one with the new 4GiB dataset
- Each static library is linked to a different NIF shared library, and redundant NIFs are registered - some which call the 512 MiB version of RandomX and some which call the 4GiB version
- Some of the complexity is mitigated by the fact that we are rolling out the 4GiB dataset only for the new `composite` packing. Chunks packed with the legacy `spora_2_6` packing will continue to use the 512 MIB dataset in order to ensure backwards compatibility (i.e. existing miners do not need to repack)
- The H0 calculation using in mining is also split between 512MiB and 4GiB implementations based on the packing address being mined.


ldmberman:
I would suggest we use https://github.com/ArweaveTeam/RandomX/tree/master2 - I rebased it on 102f8acf90a7649ada410de5499a7ec62e49e1da and included all our historical commits to the library (including your new commits) to replace 913873c13a2dffb7c4188c39b4eb188f912f523e. There is still a diff for some weird reason, but it does not look meaningful - e.g., some documentation is reverted (the master2 branch matches the current upstream master), `RANDOMX_ARGON_SALT TEXTEQU <"RandomX\x03">` from configuration.asm is restored in master2 plus some other minor changes - worth double-checking the diff if you decide to proceed with using the new branch.

ldmberman (approved):

PR #621, opened 2024-09-23
Arweave Should Not Build Before Being Stopped

humaite:
This commit fixes an issue related to `./bin/stop` script and `./bin/arweave.env`. When stopping Arweave application in dev environment, a full release is always created. It is not clean and could probably leads to errors (even if was not the case previously).

The compilation is now made only when SCRIPT_ACTION variable is not set to "stop" or if `./bin/arweave-dev` does not exist.

### DIFF
--- a/bin/arweave.env
+++ b/bin/arweave.env
@@ -18,7 +18,13 @@ else
     # extended start script to run from our current directory (which is necessary for relative
     # paths to work correctly and not be relative to the rebar3 _build release directory)
     echo Building dependencies for target ${ARWEAVE_BUILD_TARGET:-default}...
-    (cd ${PARENT_DIR} && ./ar-rebar3 ${ARWEAVE_BUILD_TARGET:-default} release)
+
+    # sometimes, arweave does not need to be build, for example
+    # when someone is stopping a node.
+    if ! [ "${SKIP_BUILD}" ]; then
+        (cd ${PARENT_DIR} && ./ar-rebar3 ${ARWEAVE_BUILD_TARGET:-default} release)
+    fi
+
     export ARWEAVE="${SCRIPT_DIR}/arweave-dev"
     export ARWEAVE_COMMAND="console"
 fi
--- a/bin/stop
+++ b/bin/stop
@@ -5,6 +5,7 @@ set -e
 SCRIPT_DIR="$(dirname "$0")"
 
 # Sets $ARWEAVE and $ARWEAVE_* variables
+SKIP_BUILD=yes
 source $SCRIPT_DIR/arweave.env
 
 $ARWEAVE stop


bin/arweave.env
```diff
@@ -12,13 +12,24 @@ if [ -f "${SCRIPT_DIR}/arweave" ]; then
...
-    # extended start script to run from our current directory (which is necessary for relative
-    # paths to work correctly and not be relative to the rebar3 _build release directory)
-    echo Building dependencies...
-    (cd ${PARENT_DIR} && ./ar-rebar3 ${ARWEAVE_BUILD_TARGET:-default} release)
+
+    # Arweave will be recompiled ONLY when the SCRIPT_ACTION variable
+    # is not set to stop. A full recompilation is not needed when the
+    # application was already started. It is required if arweave-dev
+    # does not exist though.
+    if [ "${SCRIPT_ACTION}" != "stop" ] || ! [ -f "${SCRIPT_DIR}/arweave-dev" ]; then
```
JamesPiechota, line 20: Rather than SCRIPT_ACTION... maybe some other more generic arg like "SKIP_BUILD" or something? That way we only have to set the parameter in scripts that need it. There are a lot of scripts that source arweave.env and currently don't set SCRIPT_ACTION

also if you can find a better way to do the arweave.env initialization. I don't love the current approach of setting environment variables, and since it's called via `source` we can't esily pass in command line arguments (which is why we have the kludgy "set an env var and then check it inside arweave.env")

The key thing is making sure all scripts which run arweave (e.g. arweave-server, start, stop, benchmark-xxx, data-doctor, shell, test, et....) all are launched by default under the same environment

We had some hairy bugs before where the two scripts (e.g. start vs. arweave-server) were launched with different flags which caused some hard to debug performance issues.

JamesPiechota: yeah a single entry point might be better, but also agree not for this PR. Main issue with a single entry point is just the one extra layer of nested flags. E.g. the `benchmark-xxx` and `data-doctor` commands already have the definition of a subcommand and then flags specific to that subcommand. So we'd essentially need to have command, sub-commands, and sub-sub-commands which get hairy.

Probably best to defer that change until we can revisit how we luanch all this stuff in general (e.g. if we're moving towards a form of UI, maybe the commandline complexity becomes less of an issue) So for now, no change. We just have to make sure each of the multiple endpoints are launched in a consistent state (as they are now)

bin/stop
```diff
@@ -3,8 +3,10 @@
 set -e
 
 SCRIPT_DIR="$(dirname "$0")"
+SCRIPT_ACTION="stop"
 
 # Sets $ARWEAVE and $ARWEAVE_* variables
+export SKIP_BUILD=yes
```
JamesPiechota, line 9: I don't think you need the `export`. I'm always fuzzy on that, but if you export skip_build I think it might persist in the shell environment after you run the script? If so, then we don't want that. 

e.g. SCRIPT_DIR and SCRIPT_ACTION aren't exported, but are still picked up by the `source arweave.env` line which is all we need for now.

JamesPiechota (approved):

PR #625, opened 2024-10-03
Header Synchronization Improvement

humaite:
Added a way to know if a peer is up or down
in `ar_peers` by calling `ar_peers:connected_peer/1`
and `ar_peers:disconnected_peer/1` functions. To
know if a peer is activate, one can call
`ar_peers:is_connected_peer/1` function. The full
list of active peers (present in the ranking) can
be returned by calling `ar_peers:get_peers/1`
function.

    % returns the list of all connected peers.
    % this list is highly unstable, it will depend
    % on the gun HTTP connection
    ar_peers:get_peers(connected)

All connected peers are now being tagged with a
timestamp with one connection is successful. This
timestamp can be retrieved using `ar_peers:get_connection_timestamp/1`
function. To filter peers based on their timestamps,
`ar_peers:get_peers/1` can be used like this:

    % returns the list of peers connected during the
    % last 2 days
    ar_peers:get_peers({current, 2}).

Arweave can be started with a new option, controlling
the nodes returned by `/peers`. Only `current` value
can be filtered at this time of writing, and only
a number of positive days can be configured.

    peers_list_filter current
    peers_list_filter current:365

The default peers filtering is set to 30 days: `{current, 30}`.

"tags" are used as simple key in `ar_peers`
ets table, prefixed with the `ar_tags` atom. They
are saved in the `peers` file maintained with
`ar_storage` function.

Modified `ar_http_iface_client:get_block_shadow/2`
function to add more parameters, in particular
to modify default timeout and the value used to
pick a peer from peer list.

Modified `ar_header_sync` module to only select
active peer when synchronizing headers.

Fixed a typo in `bin/console`.

### DIFF
--- a/apps/arweave/include/ar_config.hrl
+++ b/apps/arweave/include/ar_config.hrl
@@ -106,6 +106,9 @@
 %% The default number of chunks fetched from disk at a time during in-place repacking.
 -define(DEFAULT_REPACK_BATCH_SIZE, 100).
 
+%% default filtering value for the peer list (30days)
+-define(CURRENT_PEERS_LIST_FILTER, 30*60*60*24).
+
 %% The default rocksdb databases flush interval, 30 minutes.
 -define(DEFAULT_ROCKSDB_FLUSH_INTERVAL_S, 1800).
 %% The default rocksdb WAL sync interval, 1 minute.
--- a/apps/arweave/src/ar_http.erl
+++ b/apps/arweave/src/ar_http.erl
@@ -147,10 +147,12 @@ handle_info({gun_up, PID, _Protocol}, #state{ status_by_pid = StatusByPID } = St
 			[gen_server:reply(ReplyTo, {ok, PID}) || {ReplyTo, _} <- PendingRequests],
 			StatusByPID2 = maps:put(PID, {connected, MonitorRef, Peer}, StatusByPID),
 			prometheus_gauge:inc(outbound_connections),
+			ar_peers:connected_peer(Peer),
 			{noreply, State#state{ status_by_pid = StatusByPID2 }};
 		{connected, _MonitorRef, Peer} ->
 			?LOG_WARNING([{event, gun_up_pid_already_exists},
 					{peer, ar_util:format_peer(Peer)}]),
+			ar_peers:connected_peer(Peer),
 			{noreply, State}
 	end;
 
@@ -179,6 +181,7 @@ handle_info({gun_error, PID, Reason},
 					prometheus_gauge:dec(outbound_connections),
 					ok
 			end,
+			ar_peers:disconnected_peer(Peer),
 			gun:shutdown(PID),
 			?LOG_DEBUG([{event, connection_error}, {reason, io_lib:format("~p", [Reason])}]),
 			{noreply, State#state{ status_by_pid = StatusByPID2, pid_by_peer = PIDByPeer2 }}
@@ -208,6 +211,7 @@ handle_info({gun_down, PID, Protocol, Reason, _KilledStreams, _UnprocessedStream
 					prometheus_gauge:dec(outbound_connections),
 					ok
 			end,
+			ar_peers:disconnected_peer(Peer),
 			{noreply, State#state{ status_by_pid = StatusByPID2, pid_by_peer = PIDByPeer2 }}
 	end;
 
@@ -226,6 +230,7 @@ handle_info({'DOWN', _Ref, process, PID, Reason},
 					prometheus_gauge:dec(outbound_connections),
 					ok
 			end,
+			ar_peers:disconnected_peer(Peer),
 			{noreply, State#state{ status_by_pid = StatusByPID2, pid_by_peer = PIDByPeer2 }}
 	end;
 
--- a/apps/arweave/src/ar_http_iface_client.erl
+++ b/apps/arweave/src/ar_http_iface_client.erl
@@ -4,19 +4,22 @@
 
 -module(ar_http_iface_client).
 
--export([send_block_json/3, send_block_binary/3, send_block_binary/4, send_tx_json/3,
-		send_tx_binary/3, send_block_announcement/2, get_block_shadow/2, get_block_shadow/3,
-		get_block/3, get_tx/2, get_txs/2, get_tx_from_remote_peer/2, get_tx_data/2,
-		get_wallet_list_chunk/2, get_wallet_list_chunk/3, get_wallet_list/2,
-		add_peer/1, get_info/1, get_info/2, get_peers/1, get_time/2, get_height/1,
-		get_block_index/3, get_sync_record/1, get_sync_record/3,
-		get_chunk_binary/3, get_mempool/1, get_sync_buckets/1,
-		get_recent_hash_list/1, get_recent_hash_list_diff/2, get_reward_history/3,
-		get_block_time_history/3,
-		push_nonce_limiter_update/3, get_vdf_update/1, get_vdf_session/1,
-		get_previous_vdf_session/1, get_cm_partition_table/1, cm_h1_send/2, cm_h2_send/2,
-		cm_publish_send/2, get_jobs/2, post_partial_solution/2,
-		get_pool_cm_jobs/2, post_pool_cm_jobs/2, post_cm_partition_table_to_pool/2]).
+-export([send_block_json/3, send_block_binary/3, send_block_binary/4,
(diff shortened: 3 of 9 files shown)


ldmberman:
I suggest we do not immediately deactivate peers on communication failure but allow them some slack within a certain time window. The motivation is to stay connected to very useful peers. I would also consider using the existing [issue_warning](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_peers.erl#L274C21-L274C26) mechanism.

apps/arweave/src/ar_http.erl
```diff
@@ -179,6 +181,7 @@ handle_info({gun_error, PID, Reason},
 					prometheus_gauge:dec(outbound_connections),
 					ok
 			end,
+			ar_peers:inactivate_peer(Peer),
```
ldmberman, line 184: We should be getting here on a regular connection shutdown (occurs after a [minute](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_http.erl#L249)) too.

ldmberman:
The peers we have established connections with are peers we attempted to make requests to in the first place; it's not like we are establishing connections with every peer we know (so that filtering by active connections would be useful) but we are simply reusing connections we have already established for efficiency. Node being temporarily down or time-outing excessively, on the other hand, might be a useful metric to track.

Consequently we should not purge peers we are not making requests to, we should purge peers we cannot connect to for a certain amount of time.

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -802,7 +802,7 @@ priority(B, Peer) ->
 	{B#block.height, get_peer_score(Peer)}.
 
 get_peer_score(Peer) ->
-	get_peer_score(Peer, ar_peers:get_peers(lifetime), 0).
```
JamesPiechota, line 805: This *might* need to stay as `lifetime`. Or rather: I think we need to differentiate between 3 cases (so far):

1. peer list and ranking that we use to prioritize block validation and gossiping. Currently `lifetime`
2. peer list and ranking that we use for syncing chunks. Currently `current` (via `ar_data_discovery`)
3. Peer list and ranking that we use for downloading headers. Currently `lifetime`

My gut feel is that we should probably do something like:
1. keep at `lifetime` for now
2. Keep at `current` for now *or* update to something like `days, 30`
3. `days, 30`


@ldmberman what do you think? 

I think there's an argument to be made that all values could purge peers after X days, however the `lifetime` ranking is different than the `current` ranking and I don't think we want to change that yet. `lifetime` considers more data point and so is harder to game which is important for stuff like gossiping and validation order. `current` biases towards recent data and so is important for performance-critical behavior like syncing chunks and header data

But I'm not sure what's best in general. One option is we have 2 thresholds - e.g.
`lifetime` does ranking based on all data points and purges after 90 days 
`current` does the EMA ranking as it does now (biasing towards recent data points) and purges after 30 days.

In this way we still only have two peer lists: lifetime and current.

In all cases we probably want the time window to be a `-define` constant rather than hardcoded value.

ldmberman:
We do collect this information; active peers end up higher on the list whereas inactive end up lower. We can introduce rocksdb any time, off course; I would not do it earlier than we really need it, so far the data volume is small.

apps/arweave/include/ar_config.hrl
```diff
@@ -198,6 +200,7 @@
 	pool_server_address = not_set,
 	pool_api_key = not_set,
 	pool_worker_name = not_set,
+	peers_list_filter = ?DEFAULT_PEERS_LIST_FILTER,
```
JamesPiechota, line 203: I was thinking that this config would actually be a way to set the number of days that would be filtered whenever using the `current` peers list. Alternatively it could be a way to set the days of filtering for each specific peers list (e.g. it could be a multi-use flag that lest you do something like `peers_list_filter lifetime:365 peers_list_filter current:30` so you could set different thresholds or different lists)

As implemented it looks like it only applies to the `/peers` endpoint? I'm not sure that that's that useful since it won't let us configure the peers list actually used by this node when syncing chunks and/or blockchain data.

I suggest we either:
1. remove the option and just use `DEFAULT_PEERS_LIST_FILTER` everywhere. And in that case I'd suggest renaming it to something like `CURRENT_PEERS_LIST_FILTER` to make clear that it only applies to the `current` list and not the `lifetime` or `connected` lists
2. or change the flag so that it lets you configure the filter value for any of the lists (e.g. the multi-use flag example above)


I'm fine with either approach.

apps/arweave/src/ar_data_discovery.erl
```diff
@@ -204,9 +205,8 @@ pick_peers(Peers, PeerLen, N) ->
 
 collect_peers() ->
 	N = ?DATA_DISCOVERY_COLLECT_PEERS_COUNT,
-	%% rank peers by their current rating since we care about their recent throughput performance
-	collect_peers(lists:sublist(ar_peers:get_peers(current), N)).
-
+	%% rank peers by their current rating since we care out their recent throughput performance
+	collect_peers(lists:sublist(ar_peers:get_peers(?DEFAULT_PEERS_LIST_FILTER), N)).
```
JamesPiechota, line 209: Adding to my comment above, I'd suggest we have distinct peers lists that we use everywhere similar to what we hadd before. E.g. `get_peers(lifetime)` and `get_peers(current)` and then use the a constant or a config value when building those lists (rather than passing the value in when making the query)

the idea is to limit the number of peer lists for now and not open up the possibility for different areas of code to filter based on different day amounts (atleast not add that flexibility until we need it)

Also if you want to rename the peers lists that's fine too. `lifetime` and `current` aren't totally right any more. 

`lifetime` is probably more accurately named something like `consensus` or `trust`
and `current` maybe more accurately named something like `performance` (although `current` does also kinda work)

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -122,28 +125,48 @@ get_block(Peer, H, TXIndices) ->
...
 
-%% @doc Retreive a block shadow by hash or height from the given peer.
-get_block_shadow(ID, Peer, Encoding) ->
+%%--------------------------------------------------------------------
+%% @doc Retrieve a block shadow by hash or height from the given peer.
+%% @end
+%%--------------------------------------------------------------------
+get_block_shadow(ID, Peer, Encoding, Opts) ->
+	ConnectTimeout = maps:get(connect_timeout, Opts, 500),
+	Timeout = maps:get(timeout, Opts, 30),
```
JamesPiechota, line 161: could `timeout` be `30 * 1000` rather than 30 just to keep the unit the same as for `connect_timeout`?

apps/arweave/src/ar_peers.erl
```diff
@@ -149,15 +152,49 @@
...
+%%      (e.g. when syncing chunks).
+%%
+%%      `connected' returns connected `lifetime' peers.
+%%
+%%      `{timestamp, Seconds}' filters `current' list based on their
+%%      last connection, where `Seconds' is the number of seconds
+%%      since the last connection.
+%%
+%%      `{current, Days}' filters `current' list based on
+%%      their last connection where `Days' is a number of days.
```
JamesPiechota, line 177: Since for now we only have 2 lists that we care about `lifetime` and `current` I suggest we only implement support for those two lists when querying `get_peers`. As we identify more rankings that are necesary, we can extend the list of supported rankings - but I think for now we only have the 2?

apps/arweave/src/ar_peers.erl
```diff
@@ -904,13 +950,123 @@ store_peers() ->
...
-			ar_storage:write_term(peers, {Total, Records})
+			Tags = ets:foldl(fun ({{ar_tags, _, _, _}, _} = Tag, Acc) ->
+						[Tag|Acc];
+					     (_, Acc) -> Acc
+					end, [], ?MODULE),
+			?LOG_INFO([{event, store_peers}
+				   , {total, Total}
+				   , {records, length(Records)}
+				   , {tags, length(Tags)}]),
+			ar_storage:write_term(peers, {Total, Records, Tags})
```
JamesPiechota, line 961: How are the tags associated with each individual peer? It looks like the tags are being separated from the peers into one long list?

JamesPiechota: Oops, I see. Nvm.

JamesPiechota:
LGTM!

PR #626, opened 2024-10-09
Ensure ar_global_sync_record is notified when a chunk is removed from the disk pool

JamesPiechota:


apps/arweave/src/ar_sync_record.erl
```diff
@@ -606,7 +606,13 @@ emit_add_range(Start, End, ID, StoreID) ->
 	ar_events:send(sync_record, {add_range, Start, End, ID, StoreID}).
 
 emit_remove_range(Start, End, StoreID) ->
-	ar_events:send(sync_record, {remove_range, Start, End, StoreID}).
+	ar_events:send(sync_record, {remove_range, Start, End, StoreID}),
+	case ar_sync_record:is_recorded(Start+1, ar_data_sync) of
```
ldmberman, line 610: The interval may consist of more than one chunk. The problem with a generic solution like this is that we then need to filter out synced sub-intervals. That's why I think it is easier to make a decision an abstraction level above. We have three cases - blocklisting, disk pool cleanup, and chain reorg. When we blocklist, we know by design the entire blocklisted interval is going to be removed from everywhere so we emit global_remove_range. In the disk pool case, we do cleanup chunk by chunk so we can make one is_recoded query there per chunk. In the chain reorg case we also know by design the interval needs to be removed globally so we throw global_cut.

PR #627, opened 2024-10-09
Fix p3 missing pattern

humaite:
This fix an issue with p3 protocol. When the call returns timeout, the message is not correctly matched and crash the process.


JamesPiechota (approved):
LGTM!

PR #628, opened 2024-10-10
Ensure the unpacked chunk is passed from  the H2 CM peer to the H1 CM peer

JamesPiechota:
1. refactor the code for reading the PoA from disk (and then going to peers if it can't be found)
2. after finding a solution, ensure that the H2 CM peer reads the PoA - including the unpacked_chunk - before passing it off to the H1 CM peer

### DIFF
--- a/apps/arweave/src/ar_mining_server.erl
+++ b/apps/arweave/src/ar_mining_server.erl
@@ -639,7 +639,7 @@ prepare_solution(proofs, Candidate, Solution) ->
 prepare_solution(poa1, Candidate, Solution) ->
 	#mining_solution{
 		poa1 = CurrentPoA1, recall_byte1 = RecallByte1,
-		mining_address = MiningAddress, packing_difficulty = PackingDifficulty, 
+		mining_address = MiningAddress, packing_difficulty = PackingDifficulty
 	} = Solution,
 	#mining_candidate{ chunk1 = Chunk1, nonce = Nonce } = Candidate,
 
--- a/apps/arweave/src/ar_mining_server.erl
+++ b/apps/arweave/src/ar_mining_server.erl
@@ -699,7 +699,7 @@ prepare_solution(poa1, Candidate, Solution) ->
 
 prepare_solution(poa2, Candidate, Solution) ->
 	#mining_solution{ poa2 = CurrentPoA2 } = Solution,
-	#mining_candidate{chunk2 = Chunk2 } = Candidate,
+	#mining_candidate{ chunk2 = Chunk2 } = Candidate,
 
 	case prepare_poa(poa2, Candidate, CurrentPoA2) of
 		{ok, PoA2} ->
--- a/apps/arweave/src/ar_mining_server.erl
+++ b/apps/arweave/src/ar_mining_server.erl
@@ -698,31 +698,19 @@ prepare_solution(poa1, Candidate, Solution) ->
 	end;
 
 prepare_solution(poa2, Candidate, Solution) ->
-	#mining_solution{
-		poa2 = CurrentPoA2,
-		packing_difficulty = PackingDifficulty,
-		mining_address = MiningAddress, partition_number = PartitionNumber,
-		recall_byte2 = RecallByte2 } = Solution,
-	#mining_candidate{
-		chunk2 = Chunk2, h0 = H0, nonce = Nonce,
-		partition_upper_bound = PartitionUpperBound } = Candidate,
+	#mining_solution{ poa2 = CurrentPoA2 } = Solution,
+	#mining_candidate{ chunk2 = Chunk2 } = Candidate,
 
 	case prepare_poa(poa2, Candidate, CurrentPoA2) of
 		{ok, PoA2} ->
 			prepare_solution(poa1, Candidate, Solution#mining_solution{ poa2 = PoA2 });
-		{error, Error} ->
-			Chunk2Binary = case Chunk2 of
-				not_set ->
-					<<>>;
-				_ ->
-					Chunk2
-			end,
+		{error, _Error} ->
 			%% If we are a coordinated miner and not an exit node - the exit
 			%% node will fetch the proofs.
 			may_be_leave_it_to_exit_peer(
 				prepare_solution(poa1, Candidate,
 						Solution#mining_solution{
-								poa2 = #poa{ chunk = Chunk2Binary } }))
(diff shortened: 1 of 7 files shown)


apps/arweave/test/ar_coordinated_mining_tests.erl
```diff
@@ -353,14 +361,36 @@ wait_for_each_node(
...
+	end.
+
+wait_for_cross_node(Miners, ValidatorNode, CurrentHeight, ExpectedPartitions) ->
+	wait_for_cross_node(
+		Miners, ValidatorNode, CurrentHeight, sets:from_list(ExpectedPartitions), 20).
+
+wait_for_cross_node(_Miners, _ValidatorNode, _CurrentHeight, _ExpectedPartitions, 0) ->
+	?assert(false, "Timed out waiting for a cross-node solution");
+wait_for_cross_node(_Miners, _ValidatorNode, _CurrentHeight, _ExpectedPartitions, 0) ->
+	?assert(false, "Cross-node solutions can only have 2 partitions.");
```
ldmberman, line 381: A check for ExpectedPartitions' length missing?

apps/arweave/src/ar_mining_server.erl
```diff
@@ -691,38 +685,14 @@ prepare_solution(poa1, Candidate, Solution) ->
...
-					ModuleIDs = [ar_storage_module:id(Module) || Module <- Modules],
-					?LOG_ERROR([{event, failed_to_find_poa_proofs_locally},
-							{error, io_lib:format("~p", [Error])},
-							{tags, [solution_proofs]},
-							{recall_byte, RecallByte1},
-							{packing_difficulty, PackingDifficulty},
-							{modules_covering_recall_byte, ModuleIDs}]),
-					ar:console("WARNING: we have mined a block but did not find the PoA1 proofs "
-							"locally - searching the peers...~n"),
-					case fetch_poa_from_peers(RecallByte1, PackingDifficulty) of
```
ldmberman, line 705: Why do we remove fetch_poa_from_peers from here?

JamesPiechota: it all happens in prepare_poa now up here: https://github.com/ArweaveTeam/arweave/pull/628/files#diff-ea8e957c63acb7497b0ed4f9201b40f18e5d5b4fa371e67b4cff05850e4df854R646

We had 3 spots where we potentially needed to read or fetch poa when preparing a solution poa1, preparing a solution poa2, and before sending an H2 candidate to a CM peer. Previously the 3rd case was missing the read and fetch. So the intent was to pull the logic together into a single function (prepare_poa) and then call that function from each of the 3 spots that need to read/fetch the poa.

that was the intent at least...

ldmberman: Ah, yes, I see, thank you

apps/arweave/src/ar_mining_server.erl
```diff
@@ -732,49 +702,83 @@ prepare_solution(poa2, Candidate, Solution) ->
 	#mining_candidate{
 		chunk2 = Chunk2, h0 = H0, nonce = Nonce,
 		partition_upper_bound = PartitionUpperBound } = Candidate,
+
+	case prepare_poa(poa2, Candidate, CurrentPoA2) of
+		{ok, PoA2} ->
+			prepare_solution(poa1, Candidate, Solution#mining_solution{ poa2 = PoA2 });
+		{error, Error} ->
+			Chunk2Binary = case Chunk2 of
+				not_set ->
```
ldmberman, line 711: It is not a realistic case that we do not have the second chunk and leave it to the exit peer to fetch it, does it?

JamesPiechota: Yeah you're right, that case statement was preserving the existing logic but fixing a crash. the previous `prepare_solution(poa2` call didn't check whether Chunk2 was set or not and just sent it directly to the exit peer here: https://github.com/ArweaveTeam/arweave/pull/628/files#diff-ea8e957c63acb7497b0ed4f9201b40f18e5d5b4fa371e67b4cff05850e4df854L773

Unfortunately when Chunk2 was `not_set` it caused a crash during `ar_serialize`. So the above change just fixes that `ar_serialize` crash, but doesn't change the overall logic of deferring to the exit peer (which I agree, doesn't buy us much if it's the Chunk data that's missing rather than proof data)

That said I didn't fully investigate why `chunk2` was `not_set` in the first place... so unclear to me how we get to this code path anyways. The `case` statement "fix" was just in response to the stack traces shared by miners which were showing a crash in `ar_serialize` due to the poa.chunk value being `not_set` rather than a binary.

apps/arweave/src/ar_mining_server.erl
```diff
@@ -639,23 +638,18 @@ prepare_solution(proofs, Candidate, Solution) ->
...
+	} = Solution,
+	#mining_candidate{ chunk1 = Chunk1, nonce = Nonce } = Candidate,
+
+	case prepare_poa(poa1, Candidate, CurrentPoA1) of
+		{ok, PoA1} ->
+			Solution#mining_solution{ poa1 = PoA1 };
+		{error, Error} ->
+			case Chunk1 of
+				not_set ->
+					Packing = ar_block:get_packing(PackingDifficulty, MiningAddress),
```
ldmberman, line 652: The PR did not introduce it (I did [here](https://github.com/ArweaveTeam/arweave/commit/18e7eacb731781007ae9efb299d67c21f4923bb7#diff-ea8e957c63acb7497b0ed4f9201b40f18e5d5b4fa371e67b4cff05850e4df854R654),) but it does not look quite right - we should wrap the `Solution#mining_solution{ poa1 = #poa{ chunk = SubChunk } }` below in `may_be_leave_it_to_exit_peer` to make sure it errors our on exit peer because there are no proofs, only the raw chunk. Should we fix it here?

JamesPiechota: Like this?

https://github.com/ArweaveTeam/arweave/pull/628/commits/eaeffd695b764623f046013b71e49034bd8f87cf

ldmberman: Yes, thank you!

apps/arweave/src/ar_mining_server.erl
ldmberman: Hm, interesting

PR #630, opened 2024-10-11
Improve the speed of doctor-dump and allow omitting txs

JamesPiechota:


humaite (commented):
Except "style", it looks good to me. Perhaps creating a tool a bit more flexible to extract blocks and/or txs could be even better, with a way to extract a range for each of them:

```sh
# dump both blocks and txs
data-doctor dump min_height max_height data_dir output_dir

# dump only a block
data-doctor dump-block min_height max_height data_dir output_dir

# dump only a tx min_height max_height data_dir output_dir
data-doctor dump-tx min_height max_height data_dir output_dir
```

the syntax stays the same, but we can now select what kind of data we want to dump to verify/export.

JamesPiechota:
re: style. yeah agree this is not a well-polished tool. Just an emergency tool for internal use (and I'm the only one who's ever actually used it). Once we have an external use case that we need to solve, we should give some more thought to the UX. but for now it's really just a dev tool.

other changes made

PR #633, opened 2024-10-20
Implement configurable requesting of packed chunks from peers

shizzard:
@JamesPiechota WIP version implemented

### DIFF
--- a/apps/arweave/src/ar_data_sync_worker.erl
+++ b/apps/arweave/src/ar_data_sync_worker.erl
@@ -16,7 +16,8 @@
 -include_lib("arweave/include/ar_data_sync.hrl").
 
 -record(state, {
-	name = undefined
+	name = undefined,
+	request_packed_chunks = false
 }).
 
  %% # of messages to cast to ar_data_sync at once. Each message carries at least 1 chunk worth
@@ -36,7 +37,11 @@ start_link(Name) ->
 %%%===================================================================
 
 init(Name) ->
-	{ok, #state{ name = Name }}.
+	{ok, Config} = application:get_env(arweave, config),
+	{ok, #state{
+		name = Name,
+		request_packed_chunks = Config#config.data_sync_request_packed_chunks
+	}}.
 
 handle_call(Request, _From, State) ->
 	?LOG_WARNING([{event, unhandled_call}, {module, ?MODULE}, {request, Request}]),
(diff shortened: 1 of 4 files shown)


apps/arweave/src/ar_data_sync_worker.erl
```diff
@@ -267,3 +267,10 @@ sync_range({Start, End, Peer, TargetStoreID, RetryCount} = Args) ->
 					end
 			end
 	end.
+
+get_target_packing(StoreID) ->
+	{ok, Config} = application:get_env(arweave, config),
```
JamesPiechota, line 272: PR looks good! I'd suggest loading the config once during `init` and storing it in the #state. I'm not sure how significant an impact the get_env call will have (may be minor), but during heavy syncing we might expect this call to happen hundreds of times a second (and billions of times overall) so it may add up

shizzard: The actual `ar_storage_module:get_packing(StoreID)` us O(n) as it traverses the list. Is this will become a problem, this is potential candidate for optimization.

JamesPiechota (approved):
LGTM!

shizzard:
The way to configure the distributed packing.
Vocabulary:
- primary node: the node collecting the packed chunks;
- secondary node: the node packing chunks for the primary node.

Primary node must have the following configuration options:
- `local_peer <secondary-node-host>:<secondary-node-port>`: should be added for each secondary node available;
- `sync_from_local_peers_only`: only allows primary node to sync chunks from the list of local peers.

Secondary node must have the following configuration options:
- `enable pack_served_chunks`: enables secondary node to pack the requested chunks with the address and composite depth requested by the peer.

After starting the secondary nodes with the option described above, reconfigure the primary node.
You can verify the actual requested packing by sniffing the traffic on the secondary node, looking for the `GET /chunk2/<chink-id>` requests and the `x-packing` header:

```
GET /chunk2/4566605654 HTTP/1.1
host: 127.0.0.1:1985
X-Block-Format: 3
X-Network: arweave.N.1
X-Version: 8
x-bucket-based-offset: true
x-p2p-port: 1995
x-packing: composite_P6SbdirPgsRBnSz8SBMQd5R1kQaZDoLbwisb9I1wqzk.2  <--
x-release: 72
```

PR #635, opened 2024-10-24
Repack in place complete console message now includes storage module name

vird:
Proposal
Untested (because it's way difficult to test that comparing to implementation)

### DIFF
--- a/apps/arweave/src/ar_chunk_storage.erl
+++ b/apps/arweave/src/ar_chunk_storage.erl
@@ -759,15 +759,15 @@ chunk_offset_list_to_map(ChunkOffsets) ->
 repack(Cursor, RightBound, Packing, StoreID) ->
 	case ar_sync_record:get_next_synced_interval(Cursor, RightBound, ?MODULE, StoreID) of
 		not_found ->
-			Label = ar_storage_module:label_by_id(StoreID),
-			ar:console("~n~nRepacking of storage module ~s is complete! "
+			ar:console("~n~nRepacking of ~s is complete! "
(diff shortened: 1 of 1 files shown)


apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -759,14 +759,15 @@ chunk_offset_list_to_map(ChunkOffsets) ->
 repack(Cursor, RightBound, Packing, StoreID) ->
 	case ar_sync_record:get_next_synced_interval(Cursor, RightBound, ?MODULE, StoreID) of
 		not_found ->
-			ar:console("~n~nRepacking is complete! We suggest you stop the node, rename "
+			Label = ar_storage_module:label_by_id(StoreID),
+			ar:console("~n~nRepacking of storage module ~s is complete! "
+					"We suggest you stop the node, rename "
 					"the storage module folder to reflect the new packing, and start the "
-					"node with the new storage module.~n", []),
+					"node with the new storage module.~n", [Label]),
```
JamesPiechota, line 766: I thin we want to output the `StoreID` rather than the label. The Label is the obfuscated storage module name (with the mining address hidden), but since the console message is only visible to the miner, it's probably more helpful to include the real storage module name (non obfuscated). Like what is included in the LOG_INFO: https://github.com/ArweaveTeam/arweave/pull/635/files#diff-70860f94e3504d0cdd89e58a65a090fadb6a75eabc58f0a89aeae829109e8dbbR768

JamesPiechota (approved):
LGTM!

PR #631, opened 2024-10-14
Initial commit for ar_p3 rework

shizzard:
Implement the `ar_p3_ledger` gen_server.
Implement the `ar_p3_sup` supervisor.

PR #636, opened 2024-10-25
fix: fix mining chunk cache

JamesPiechota:
https://github.com/ArweaveTeam/arweave-dev/issues/666

### DIFF
--- a/apps/arweave/src/ar_mining_server.erl
+++ b/apps/arweave/src/ar_mining_server.erl
@@ -31,6 +31,8 @@
 	merkle_rebase_threshold		= infinity,
 	is_pool_client				= false,
 	allow_composite_packing		= false,
+	% For now we only support a single packing format when mining, this is the packing
+	% difficulty of that format.
 	packing_difficulty			= 0
 }).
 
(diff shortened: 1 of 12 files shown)


apps/arweave/src/ar_mining_server.erl
```diff
@@ -391,56 +409,38 @@ update_cache_limits(State) ->
...
-
-	%% We shard the chunk cache across every active worker. Only workers that mine a partition
-	%% included in the current weave are active.
-	NewCacheLimit = max(1, OverallCacheLimit div NumActivePartitions),
+	{MinimumCacheLimitMiB, OverallCacheLimitMiB, PartitionCacheLimit, VDFQueueLimit,
+		GarbageCollectionFrequency} =
+			calculate_cache_limits(NumActivePartitions,  State#state.packing_difficulty),
 
-	case NewCacheLimit == State#state.chunk_cache_limit of
+	case PartitionCacheLimit == State#state.chunk_cache_limit of
```
shizzard, line 416: These parts of the code are easily extracted into sharp functions like
```erlang
maybe_update_cache_limit(PartitionCacheLimit, #state{chunk_cache_limit = PartitionCacheLimit} = S0) -> S0;
maybe_update_cache_limit(PartitionCacheLimit, S0) ->
    ...
```
This will greatly reduce the nesting within a function and reduce function scope.

apps/arweave/src/ar_mining_server.erl
```diff
@@ -391,56 +409,38 @@ update_cache_limits(State) ->
...
-						"Consider changing the 'mining_server_chunk_cache_size_limit' option.",
-						[IdealCacheLimit]);
+						"Consider changing the 'mining_cache_size_mb' option.",
+						[MinimumCacheLimitMiB]);
 				false -> ok
 			end,
-			GarbageCollectionFrequency = 4 * VDFQueueLimit * 1000,
+
 			GCRef =
 				case State#state.gc_frequency_ms == undefined of
```
shizzard, line 445: The logic is not clear here.
We're checking if the `gc_frequency_ms` is defined in the process state, but not checking if the `gc_process_ref` is set (the actual marker that the timer is set here). So there are a possible scenario where we will have two timers set at the same time.
I would recommend saving the actual timer reference in the state and extract the timer initialization function into a separate function. The example may be found in the [P3 ledger implementation](https://github.com/ArweaveTeam/arweave/pull/631/files#diff-cc13c07b3a4f51899a3b95183c92361b26ece97879ac1df7fde8a5482a809452R431) (`handle_continue(?msg_continue_start_shutdown_timer, ...)` @ `ar_p3_ledger.erl` line 431):
- ensure you cancel the existing timer if the timer exists and still running;
- ensure to put the unique marker into timer message ()this is already done here);

When hitting the timer message handler, just do the related work and ensure to call the extracted timer initialization function once again, that will do all the needed things.

shizzard (commented):
No major problems with the code (didn't test it though)

apps/arweave/src/ar_mining_server.erl
JamesPiechota: Done! https://github.com/ArweaveTeam/arweave/pull/636/commits/1a8a2ce25b993da9e0cec6282ed00abbf28da073

apps/arweave/src/ar_mining_server.erl
JamesPiechota: Got it - Done! https://github.com/ArweaveTeam/arweave/pull/636/commits/cb46fb0912839e365c9f23784b659c3b93e2ad25

PR #637, opened 2024-10-27
ar_data_sync: Skip unpacking if the chunk was downloaded from the local peer

shizzard:


apps/arweave/src/ar_data_sync.erl
```diff
@@ -1078,6 +1079,7 @@ handle_cast({store_fetched_chunk, Peer, Byte, Proof} = Cast, State) ->
 	Offset = SeekByte - BlockStartOffset,
 	ValidateDataPathRuleset = ar_poa:get_data_path_validation_ruleset(BlockStartOffset,
 			get_merkle_rebase_threshold()),
+	IsLocalPeer = lists:member(Peer, State#sync_data_state.local_peers),
```
JamesPiechota, line 1082: Nice - elegant!

A couple points:
1. It's possible to query chunks from local_peers that *aren't* packed to our target packing format. In that case we don't need to validate, but we do still need to repack. Will that still happen in this flow?
2. When we are querying from a local_peer, ideally we'd skip the validate_proof call entirely.

shizzard: Basically, we need to call for `validate_proof` here exactly to check if the chunk is packed as we want, or we need to repack it. The function itself just takes the proof and does some calculations on it, so no long calls or other delays involved, should be fast.

JamesPiechota: I don't think validate_proof checks if the chunk is packed correctly. I believe it just checks if it's packed or unpacked - if it's unpacked it can validate the chunk directly, if it's packed to any format it returns `need_unpacking`. It is `pack_and_store_chunk` that then checks if a valid chunk is packed correctly ad packs it as needed.

So maybe the only change needed is to swap `store_chunk` to `pack_and_store_chunk` - however in that case it might still be worth avoiding `validate_proof` mostly because I don't believe it's necessary. Even though performance may not be a concern, it adds a layer to the logic flow that needs to be  understood and maintained (e.g. the next person to see the code will wonder why validate_proof is called even though the result is ignored, and may spend time digging into the code, etc... alternatively it's possible someone adds a side effect to validate_proof or changes its return value which then impacts the "no validation needed" case, etc...)

Also, given the complexity of this logic, it would be good to also run some live tests (perhaps using debug output to confirm the correct behavior).

i.e.

1. sync from local peers where client requests chunks packed correctly. Chunk should be written without any repacking
2. sync from local peers where client requests `any` chunks, chunks should be repacked before being written.

shizzard: @JamesPiechota yes, I articulated it incorrectly, it does not check the packing, but it checks the proof, as the name implies :)
The thing is that to get all the proper arguments for storing the chunk I will need to duplicate half of the code that is doing the proof validation now. In ideal circumstances you're probably right, the proof validation is not needed, but in this case I think this is the best solution.
I will do all the tests, of course.

apps/arweave/src/ar_data_sync.erl
```diff
@@ -1089,6 +1093,8 @@ handle_cast({store_fetched_chunk, Peer, Byte, Proof} = Cast, State) ->
 				true ->
 					decrement_chunk_cache_size(),
 					{noreply, State};
+				false when IsLocalPeer ->
+					process_valid_fetched_chunk(ChunkArgs, Args, State);
```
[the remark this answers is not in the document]
JamesPiechota: awesome! 🤞

shizzard:
@JamesPiechota I did implement the requested packing of the served chunks for local peers only.

I did test all the possible combinations of packer and miner configurations, including:
- for packer: 
  - setting `local_peer` address and not setting it (to check the `local_peers` filtering for the packer);
  - setting the `pack_served_chunks` and not setting it (to check on-demand packing);
- for miner:
  - setting `sync_from_local_peers_only` to ensure syncing only from local packer;
  - setting `data_sync_request_packed_chunks`: to test on-demand packing);
  - setting `local_peer` address and not setting it (to check the `local_peers` filtering for requesting packed chunks);

I didn't find any node misbehaving.

JamesPiechota:
I'm rerunning the auto tests. `ar_packing_tests` failed - in the past that hasn't been that flaky, so it may indicate a real issue, but re-running just in case.

JamesPiechota:
> @JamesPiechota I did implement the requested packing of the served chunks for local peers only.
> 
> I did test all the possible combinations of packer and miner configurations, including:
> 
> * for packer:
>   
>   * setting `local_peer` address and not setting it (to check the `local_peers` filtering for the packer);
>   * setting the `pack_served_chunks` and not setting it (to check on-demand packing);
> * for miner:
>   
>   * setting `sync_from_local_peers_only` to ensure syncing only from local packer;
>   * setting `data_sync_request_packed_chunks`: to test on-demand packing);
>   * setting `local_peer` address and not setting it (to check the `local_peers` filtering for requesting packed chunks);
> 
> I didn't find any node misbehaving.

Those tests look good! Were you able to confirm when chunk repacking actually occurs on both packer and miner? I think last time we only checked that the requests were formatted correctly, and missed the fact that even with correctly formatted requests, the miner had to do some unpacking to validate.

shizzard:
> Were you able to confirm when chunk repacking actually occurs on both packer and miner? I think last time we only checked that the requests were formatted correctly, and missed the fact that even with correctly formatted requests, the miner had to do some unpacking to validate.

No, how can I actually check that the chunk was packed correctly?

JamesPiechota:
> > Were you able to confirm when chunk repacking actually occurs on both packer and miner? I think last time we only checked that the requests were formatted correctly, and missed the fact that even with correctly formatted requests, the miner had to do some unpacking to validate.
> 
> No, how can I actually check that the chunk was packed correctly?

Easiest way is probably to monitor the metrics. Every time a node does a pack, unpack, or repack, the `packing_requests` metric is incremented. You can use this [Grafana board](https://arweave.grafana.net/d/cvSjnX-Vk/packing?orgId=1&from=now-6h&to=now&var-node=127.0.0.1:9100&var-cores=8) to monitor packing behavior, or also manually check the `/metrics` endpoint and look at the `packing_requests` value. If you want to use the grafana board we'll have to add your IP:PORT to our prometheus server, If you DM me those values I can make the addition.

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2008,21 +2008,26 @@ handle_get_chunk(OffsetBinary, Req, Encoding) ->
...
-									true ->
-										ok = ar_semaphore:acquire(get_and_pack_chunk,
-												infinity),
-										{RequestedPacking, ok}
+								IsPackServedChunks = lists:member(pack_served_chunks, Config#config.enable),
+								Peer = cowboy_req:peer(Req),
+								IsLocalPeerAddr = lists:foldl(fun
+									(_, true) -> true;
+									(LocalPeer, _) -> is_matching_addr(Peer, LocalPeer)
+								end, false, Config#config.local_peers),
```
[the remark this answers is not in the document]
shizzard: Actually not, that's the problem here. 
Cowboy uses standard Erlang address tuple: `{{O1, O2, O3, O4}, Port}`, while local peers are parsed into five-tuple: `{O1, O2, O3, O4, Port}`.
Probably should change this, but I'm afraid I can suddenly break something. Should I try?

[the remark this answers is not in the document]
JamesPiechota: @shizzard unfortunately it looks `ar_util:safe_parse_peer` and `ar_util:parse_peer` are used to parse all config peers and they always dump to the 5 element tuple. I agree probably too risky to update that now, but can you log a github issue? Seems like something we should rationalize with cowboy.

FWIW it looks like the parse_peer code is from 7 years ago (!)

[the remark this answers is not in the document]
shizzard: @JamesPiechota #640

apps/arweave/src/ar_serialize.erl
```diff
@@ -2120,6 +2120,8 @@ encode_packing(spora_2_5, _Strict) ->
 encode_packing(unpacked, _Strict) ->
 	"unpacked".
 
+decode_packing(<<"any">>, _Error) ->
```
JamesPiechota, line 2123: I think the `any` case was omitted before because `any` isn't a valid packing format. You'll see above under `encode_packing` that the `any` case is only handled when `Strict` is `false`. The idea is that we only want to handle encoding or decoding `any` if we're explicitly in a code block that doesn't need a valid packing format. If however we're in a code block that expects a valid packing format, `any` or `<<"any">>` should cause an error.

Im not sure what the implication are of supportin `<<"any">>` in all decode paths. Could be fine, but worth double checking that all of the code paths that call decode_packing either explicitly allow for `any` or have another chekc that validates that the packing format is correct.

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2008,21 +2008,26 @@ handle_get_chunk(OffsetBinary, Req, Encoding) ->
...
+										{none, {reply, {403, #{}, <<>>, Req}}}
 								end
 						end,
 					case CheckRecords of
 						{reply, Reply} ->
 							Reply;
 						ok ->
 							Args = #{ packing => ReadPacking,
-									bucket_based_offset => IsBucketBasedOffset },
+									bucket_based_offset => IsBucketBasedOffset, pack => true },
```
JamesPiechota, line 2030: why are we requesting the chunk to be repacked here? This is outside the `pack_served_chunks` check, right?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -2820,6 +2827,11 @@ process_valid_fetched_chunk(ChunkArgs, Args, State) ->
 					%% The chunk has been synced by another job already.
 					decrement_chunk_cache_size(),
 					{noreply, State};
+				false when Packing =/= unpacked ->
+					true = AbsoluteEndOffset == AbsoluteTXStartOffset + ChunkEndOffset,
+					pack_and_store_chunk({DataRoot, AbsoluteEndOffset, TXPath, TXRoot,
+							DataPath, Packing, ChunkEndOffset, ChunkSize, Chunk,
+							none, none, none}, State);
```
JamesPiechota, line 2834: Do you know why we need this extra case? Looks like the only difference is passin gin `none` instead of `UnpackedChunk`?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2008,21 +2008,26 @@ handle_get_chunk(OffsetBinary, Req, Encoding) ->
...
+								IsLocalPeerAddr = lists:foldl(fun
+									(_, true) -> true;
+									(LocalPeer, _) -> is_matching_addr(Peer, LocalPeer)
+								end, false, Config#config.local_peers),
+								case {IsPackServedChunks, IsLocalPeerAddr} of
+									{true, true} ->
+										ok = ar_semaphore:acquire(get_and_pack_chunk, infinity),
+										{RequestedPacking, ok};
+									{_, _} ->
+										{none, {reply, {403, #{}, <<>>, Req}}}
```
JamesPiechota, line 2022: I think this should be a 404 to match the old behavior: https://github.com/ArweaveTeam/arweave/pull/637/files#diff-0b7084531612365a22d52032a422eb664adc91b34cefb5f8d9499584072aa489L2013

We could have something like {true, false} be a 403

But I think best to just always 404 if the server can't or is unwilling to serve the chunk in the requested format

shizzard: Ah yes, I did this change to distinct this particular response from others during the tests. Will revert.

apps/arweave/src/ar_data_sync.erl
shizzard: This is a dirty fix for the problem, I will refactor this to make it more clear.
But the reason for having this clause here is the wrong naming of the variables inside the `ChunkArgs`. It actually can hold `Packing` + `PackedChunk`, just the original one. So we need to manually set `UnpackedChunk` to `none` here.

apps/arweave/src/ar_http_iface_middleware.erl
shizzard: Actually no, the CheckRecords is returned from the `pack_served_chunks` clause, and repacking the chunk requres this flag.

PR #638, opened 2024-11-03
add verify mode

JamesPiechota:


apps/arweave/src/ar_data_sync.erl
```diff
@@ -721,7 +730,7 @@ handle_cast({join, RecentBI}, State) ->
 			PreviousWeaveSize = element(2, hd(CurrentBI)),
 			{ok, OrphanedDataRoots} = remove_orphaned_data(State, Offset, PreviousWeaveSize),
 			{ok, Config} = application:get_env(arweave, config),
-			[gen_server:cast(list_to_atom("ar_data_sync_" ++ ar_storage_module:label(Module)),
+			[gen_server:cast(name(Module),
```
ldmberman, line 733: name/1 accepts StoreID so we need name(ar_storage_module:id(Module))

apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -408,22 +408,27 @@ apply_external_update(Update, Peer) ->
...
-	State =
-		case ar_node:is_joined() of
-			true ->
-				Blocks = get_blocks(),
-				handle_initialized(Blocks, #state{});
-			_ ->
-				#state{}
-		end,
-	case ar_config:use_remote_vdf_server() and not ar_config:compute_own_vdf() of
+	case ar_config:verify() of
```
ldmberman, line 411: I think, it makes more sense to check for auto_join() here

apps/arweave/src/ar_nonce_limiter_server_sup.erl
```diff
@@ -21,7 +21,7 @@ start_link() ->
 %% ===================================================================
 
 init([]) ->
-	case ar_config:is_vdf_server() of
+	case ar_config:is_vdf_server() andalso not ar_config:verify() of
```
ldmberman, line 24: I think, it makes more sense to check for auto_join()

apps/arweave/src/ar_peers.erl
```diff
@@ -375,13 +375,20 @@ resolve_and_cache_peer(RawPeer, Type) ->
...
 init([]) ->
-	%% Trap exit to avoid corrupting any open files on quit.
-	process_flag(trap_exit, true),
-	ok = ar_events:subscribe(block),
-	load_peers(),
-	gen_server:cast(?MODULE, rank_peers),
-	gen_server:cast(?MODULE, ping_peers),
-	timer:apply_interval(?GET_MORE_PEERS_FREQUENCY_MS, ?MODULE, discover_peers, []),
+	{ok, Config} = application:get_env(arweave, config),
+	case Config#config.verify of
```
ldmberman, line 379: Why do we turn it off?

JamesPiechota: Mostly for performance reasons. Pinging the peers seems to be one of the slower parts during node launch. That said... not completely sure this is the right approach. I slacked you with more context.

apps/arweave/src/ar_nonce_limiter_server_sup.erl
JamesPiechota: Latest commit removes this check and instead uses the `set_dependent_flags` process to disable all vdf features.

apps/arweave/src/ar_nonce_limiter.erl
JamesPiechota: Latest commit removes this check and instead uses the `set_dependent_flags` process to disable all vdf features.

apps/arweave/src/ar_data_sync.erl
JamesPiechota: thanks!

apps/arweave/src/ar_verify_chunks.erl
```diff
@@ -0,0 +1,544 @@
...
+				{true, _Proof} ->
+					State
+			end;
+		Error ->
+			invalidate_chunk(
+				read_data_path_error, AbsoluteOffset, ChunkSize, [{reason, Error}], State)
+	end.
+
+verify_chunk_storage(Offset, _ChunkSize, {End, Start}, State) 
+		when Offset >= Start andalso Offset =< End ->
```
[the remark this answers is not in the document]
JamesPiechota: Adding: `Offset - ?DATA_CHUNK_SIZE >= Start` (since if Offset <= End, Offset - DATA_CHUNKS_SIZE will be too)

Thanks!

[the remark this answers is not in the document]
JamesPiechota: https://github.com/ArweaveTeam/arweave/pull/638/commits/f796b4815b18b9670860fdd00d1797c75a6bc57b

apps/arweave/src/ar_data_sync.erl
JamesPiechota: https://github.com/ArweaveTeam/arweave/pull/638/commits/f796b4815b18b9670860fdd00d1797c75a6bc57b

apps/arweave/src/ar_verify_chunks.erl
```diff
@@ -0,0 +1,544 @@
...
+	State2.
+
+verify_chunks(not_found, _Intervals, State) ->
+	State#state{ cursor = State#state.end_offset };
+verify_chunks({End, _Start}, _Intervals, #state{cursor = Cursor} = State) when Cursor >= End ->
+	State;
+verify_chunks({IntervalEnd, IntervalStart}, Intervals, State) ->
+	#state{cursor = Cursor, store_id = StoreID} = State,
+	Cursor2 = max(IntervalStart, Cursor),
+	ChunkData = ar_data_sync:get_chunk_by_byte({chunks_index, StoreID}, Cursor2+1),
```
ldmberman, line 110: get_chunk_by_byte accepts an offset >= Start, < End so passing Cursor2 should work

JamesPiechota: Hm, it turns out that removing the `+1` caused the script to hang. Not sure why, but I re-added the +1

ldmberman: Ah, I was wrong, sorry, it needs > Start, =< End

PR #647, opened 2024-11-25
Fix High Load Average on Transaction Interface

humaite:
When a server has the full transaction locally, and uses spinning disk, the amount of requests from other nodes is too high, and leads to a bottleneck on rocksdb database.

This commit adds a semaphore to control the amount of requests on `ar_http_iface_middleware:handle_get_tx/3` function, to avoid a node to be overflooded by requests.

see: https://github.com/ArweaveTeam/arweave-dev/issues/704

### DIFF
--- a/apps/arweave/include/ar_config.hrl
+++ b/apps/arweave/include/ar_config.hrl
@@ -67,6 +67,7 @@
 -define(MAX_PARALLEL_POST_CHUNK_REQUESTS, 100).
 -define(MAX_PARALLEL_GET_SYNC_RECORD_REQUESTS, 10).
 -define(MAX_PARALLEL_REWARD_HISTORY_REQUESTS, 1).
+-define(MAX_PARALLEL_GET_TX_REQUESTS, 20).
 
 %% The number of parallel tx validation processes.
 -define(MAX_PARALLEL_POST_TX_REQUESTS, 20).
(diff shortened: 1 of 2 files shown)


apps/arweave/include/ar_config.hrl
```diff
@@ -67,6 +67,7 @@
 -define(MAX_PARALLEL_POST_CHUNK_REQUESTS, 100).
 -define(MAX_PARALLEL_GET_SYNC_RECORD_REQUESTS, 10).
 -define(MAX_PARALLEL_REWARD_HISTORY_REQUESTS, 1).
+-define(MAX_PARALLEL_GET_TX_REQUESTS, 1).
```
JamesPiechota, line 70: What do you think about setting the default a little higher? It looks like our nodes that were underload from `GET /tx/{hash}` or `GET /tx/{hash}/status` requests were getting hit with like 50 per second. But the ones that weren't falling over could still get hit with up to 5 or 10 requests per second in short bursts. The latency on those requests was pretty short so maybe it doesn't really matter.

What about a default of 5 or 10? Do you think bumping it to 10 would make the full-chain nodes fallover?

JamesPiechota: (PS: I changed the PR base to `release/performance-2.8`)

PR #650, opened 2024-11-25
`ar_data_sync` skip local peer verification

shizzard:


apps/arweave/src/ar_data_sync.erl
```diff
@@ -1871,8 +1878,12 @@ get_tx_offset_data_in_range2(TXOffsetIndex, TXIndex, Start, End) ->
 get_tx_data(Start, End, Chunks) when Start >= End ->
 	{ok, iolist_to_binary(Chunks)};
 get_tx_data(Start, End, Chunks) ->
-	case get_chunk(Start + 1, #{ pack => true, packing => unpacked,
-			bucket_based_offset => false }) of
+	{ok, Config} = application:get_env(arweave, config),
+	case get_chunk(Start + 1, #{
+		pack => lists:member(pack_served_chunks, Config#config.enable),
```
JamesPiechota, line 1883: I couldn't trace through all code paths leading to `get_tx_data`. Some look like they start at an HTTP handler - makes sense that we'd only repack when the option is set for those.

But it looks like a bunch come from `ar_storage:read_tx` which I think is used by a lot of internal processes. I wonder if those processes might depend on the chunk being repacked? Were you able to confirm one way or another?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2008,21 +2008,24 @@ handle_get_chunk(OffsetBinary, Req, Encoding) ->
...
 							{{true, _}, _StoreID} ->
 								{ok, Config} = application:get_env(arweave, config),
-								case lists:member(pack_served_chunks, Config#config.enable) of
-									false ->
-										{none, {reply, {404, #{}, <<>>, Req}}};
+								IsPackServedChunks = lists:member(pack_served_chunks, Config#config.enable),
+								Peer = ar_http_util:arweave_peer(Req),
+								IsLocalPeerAddr = lists:member(Peer, Config#config.local_peers),
+
+								case IsPackServedChunks orelse IsLocalPeerAddr of
```
JamesPiechota, line 2015: Should this be `andalso`? I forget where we landed

Like I could see it working like this:
- server sets `pack_served_chunks`, in which case it only repacks chunks for members of its `local_peers`
- client sets `request_packed_chunks` and `sync_from_local_peers_only` in which case it requests packed chunks from its `local_peers` and does not validate it them before storing.

apps/arweave/src/ar_data_sync.erl
shizzard: No, but I'll check what I can.

apps/arweave/src/ar_http_iface_middleware.erl
shizzard: We decided that we pack served chunks for local peers even when the option is not set.

apps/arweave/src/ar_data_sync.erl
JamesPiechota: worst case you can revert this to `pack => true` - I don't think there's much of a risk there (and probably no new risk?)

apps/arweave/src/ar_http_iface_middleware.erl
JamesPiechota: I think we should always require the option to be set. Unless you and Lev discussed otherwise?

i.e. there are a lot of scenarios where local_peers are set for other reasons (e.g. reduce rate limiting between CM cluster peers) and in those cases we probably don't want to also pack served chunks unless it's desired (e.g. we wouldn't want a CM Exit Node to get bogged down repacking accidentally)

shizzard: We discussed this few times:
- CM exit node will unlikely have any data at all;
- local_peers are considered trusted and therefore any requested packing is served.

JamesPiechota: Just chatted on slack. Current approach as implemented works! clients will only request packed chunks if they have `fetch_packed_chunks` set, but if they do request a packed chunk, the local_peer will repack.

PR #651, opened 2024-11-27
Add node introspection options

shizzard:


JamesPiechota (approved):
Looks good to me @shizzard . I'm not sure I understand the module name `user_default` though? If it's only only supposed to be used when debugging or introspection... maybe something like `introspection.erl` or something? As it is, the module sounds important, like it's a core part of the normal node execution.

in any case: since I don't really know how i will be used, I could be off base on the naming! I defer to you what's best.

shizzard:
> Looks good to me @shizzard . I'm not sure I understand the module name `user_default` though? If it's only only supposed to be used when debugging or introspection... maybe something like `introspection.erl` or something? As it is, the module sounds important, like it's a core part of the normal node execution.
> 
> in any case: since I don't really know how i will be used, I could be off base on the naming! I defer to you what's best.

@JamesPiechota the name is indeed important: [docs](https://www.erlang.org/doc/apps/erts/erl_cmd.html#configuration).
I'll check the `shell_default` though, I missed that entirely. If those are mutually interchangeable, we can go with `shell_default`.

PR #653, opened 2024-11-28
Optimize ar_tx_emitter module

humaite:
Only current (active) peers should be able to receive the
chunks, others will generate unecessary timeouts.
    
When chunks are propagated across the network, the to be
acknowledged are the trusted peer. On the previous version,
the trusted peer where put at the end of the list.
    
Cleaned up ar_tx_emitter_sup, converting old tuple to map.
    
see: https://github.com/ArweaveTeam/arweave-dev/issues/720

### DIFF
--- a/apps/arweave/src/ar_tx_emitter.erl
+++ b/apps/arweave/src/ar_tx_emitter.erl
@@ -42,7 +42,10 @@ start_link(Name, Workers) ->
 
 init(Workers) ->
 	gen_server:cast(?MODULE, process_chunk),
-	{ok, #state{ workers = queue:from_list(Workers), currently_emitting = sets:new() }}.
+	State = #state{ workers = queue:from_list(Workers)
+		      , currently_emitting = sets:new()
+		      },
+	{ok, State}.
 
 handle_call(Request, _From, State) ->
 	?LOG_WARNING([{event, unhandled_call}, {module, ?MODULE}, {request, Request}]),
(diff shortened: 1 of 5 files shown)


apps/arweave/src/ar_tx_emitter.erl
```diff
@@ -42,26 +42,42 @@ start_link(Name, Workers) ->
...
+
+	% only current (active) peers should be used, using lifetime
+	% peers will create unecessary timeouts. The first to
+	% contact are the trusted peers
 	TrustedPeers = ar_peers:get_trusted_peers(),
-	Peers = (ar_peers:get_peers(lifetime) -- TrustedPeers) ++ TrustedPeers,
-	{ok, Config} = application:get_env(arweave, config),
-	{Q2, Emitting2} = emit(ar_mempool:get_propagation_queue(), Q, Emitting, Peers,
-			Config#config.max_propagation_peers, ?CHUNK_SIZE),
+	CurrentPeers = ar_peers:get_peers(current) -- TrustedPeers,
```
JamesPiechota, line 61: Everything other than this change is refactoring, right? i.e. we shouldn't expect any functional change other than this line?

apps/arweave/src/ar_tx_emitter_worker.erl
```diff
@@ -35,27 +35,11 @@ handle_cast({emit, TXID, Peer, ReplyTo}, State) ->
 			ok;
 		TX ->
 			StartedAt = erlang:timestamp(),
-			TrustedPeers = ar_peers:get_trusted_peers(),
-			PropagatedTX = tx_to_propagated_tx(TX, Peer, TrustedPeers),
-			Release = ar_peers:get_peer_release(Peer),
-			SendFun =
-				case Release >= 42 of
-					true ->
-						fun() ->
```
JamesPiechota, line 44: Any idea why the old code wrapped this in a `fun`? Your refactored version is much cleaner - but I can't for the life of me figure out why the old version added the extra `fun` indirection. Can you think of any reason? Perhaps just "historical" - i.e. maybe there used to be a reason, but it was refactored out or otherwise changed at some point in the past and the reason disappeared.

JamesPiechota (approved):
Looks great! I added 2 comments, but they're just questions. I don't see any needed code changes.

apps/arweave/src/ar_tx_emitter_worker.erl
JamesPiechota: yep agree!

PR #659, opened 2024-12-02
Fix Gun Message Leaks

humaite:
`ar_http` client was using internal functions
from inet (`inet:start_timer/1` and `inet:stop_timer/1`). Those functions are wrapper around `timer` module and send a message to the current active process doing the http request. Unfortunately, this is not compatible with `gun:await/3è function and breaks its loop. Gun messages are after send to the active process.

see: https://github.com/ArweaveTeam/arweave-dev/issues/717

### DIFF
--- a/apps/arweave/src/ar_http.erl
+++ b/apps/arweave/src/ar_http.erl
@@ -38,8 +38,7 @@ unblock_peer_connections() ->
 
 req(#{ peer := {_, _} } = Args) ->
 	req(Args, false);
-req(Args) ->
-	#{ peer := Peer } = Args,
+req(#{ peer := Peer } = Args) ->
 	{ok, Config} = application:get_env(arweave, config),
 	case Config#config.port == element(5, Peer) of
 		true ->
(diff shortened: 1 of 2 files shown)


apps/arweave/src/ar_http.erl
```diff
@@ -301,17 +300,21 @@ method_to_list(_) ->
 	"unknown".
 
 request(PID, Args) ->
-	Timer = inet:start_timer(maps:get(timeout, Args, ?HTTP_REQUEST_SEND_TIMEOUT)),
+	Timeout = maps:get(timeout, Args, ?HTTP_REQUEST_SEND_TIMEOUT),
```
JamesPiechota, line 303: I realize I'm not sure how the gun request/response works. We'll often pass in a `timeout => Timeout` option when making an HTTP request. I see that that gets pass into the `request2` call ... which is where I assume it was used. But I guess the gun works, waiting on the response is handled separately, so we also need to pass the timeout into the response wait? So that we correctly timeout if no response is handled in time?

Previously we used the inet_timer, but now we're just using the same `timeout => Timeout` option?

If I've understood correctly: sounds good!

JamesPiechota (approved):

apps/arweave/src/ar_http.erl
JamesPiechota: Got it! LGTM. Merge when ready.

PR #660, opened 2024-12-02
Implement naive clamping of the chunk cache size for mining worker

shizzard:


apps/arweave/src/ar_mining_worker.erl
```diff
@@ -13,24 +13,24 @@
...
-	sub_chunk_cache_size		= #{},
-	sub_chunk_cache_limit		= 0,
-	vdf_queue_limit				= 0,
-	latest_vdf_step_number		= 0,
-	is_pool_client				= false,
-	h1_hashes					= #{},
-	h2_hashes					= #{}
+    name = not_set :: atom(),
+    partition_number = not_set :: non_neg_integer() | not_set,
+    diff_pair = not_set, % :: ?DIFF_PAIR()
```
shizzard, line 18: @vird what is the diff pair here?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -752,11 +752,9 @@ update_sub_chunk_cache_size(0, _SessionKey, State) ->
 	State;
 update_sub_chunk_cache_size(Delta, SessionKey, State) ->
 	CacheSize = maps:get(SessionKey, State#state.sub_chunk_cache_size, 0),
-	prometheus_gauge:inc(mining_server_chunk_cache_size,
-			[State#state.partition_number], Delta),
-	State#state{
-		sub_chunk_cache_size = maps:put(SessionKey, CacheSize + Delta,
-		State#state.sub_chunk_cache_size) }.
+	NewCacheSize = ar_util:clamp_lower(CacheSize + Delta, 0),
```
shizzard, line 755: The key change: do not let the sub_chunk_cache_size go below zero, as this does not make any sense.

ldmberman: Aren't we fighting the symptom? Perhaps, let's leave a comment this is a temporary measure until we figure out why we subtract more than we add, if that is the case.

apps/arweave/src/ar_mining_worker.erl
ldmberman: A pair of mining difficulties, the first element is the mining difficulty for one-chunk solutions, the second element - for two-chunk solutions.

apps/arweave/src/ar_util.erl
```diff
@@ -381,3 +381,23 @@ assert_file_exists_and_readable(FilePath) ->
...
+%% If the value is greater than the upper bound, return the upper bound.
+clamp_upper(Value, UpperBound) when Value > UpperBound ->
+	UpperBound;
+clamp_upper(Value, _UpperBound) ->
+	Value.
+
+%% @doc Clamp a value to a lower and upper bound.
+%% If the value is less than the lower bound, return the lower bound.
+%% If the value is greater than the upper bound, return the upper bound.
+clamp(Value, LowerBound, UpperBound) ->
```
JamesPiechota, line 402: FYI: `ar_util:between/3` also does this: https://github.com/ArweaveTeam/arweave/pull/660/files#diff-1f26621f4ea9ef47e071afce0f4ce9fc3e6bacc76089ac9037c00e0c07a5ce9eR46

Can you add a comment to `ar_util:between` that people should start using ar_util:clamp instead?

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: @shizzard what about adding a log_warning message whenever CacheSize+Delta is below zero? We do want to fix the OOMs as quick as possible and clamping may help there, but I'm a little worried we may also lose this valuable signal that something is wrong. Which may make tracking down the true root cause harder.

JamesPiechota: I slacked with some more information that might help with reproducing the negative cache size (and so might help get at the root cause)

shizzard: > what about adding a log_warning message

Thats a very good idea, will do.

apps/arweave/src/ar_util.erl
shizzard: Ah, didn't see the existing one, okay.

PR #661, opened 2024-12-05
Fix missing clause in ar_http_iface_middleware

humaite:
This PR also adds specifications and improve documentation to help doing a more deeper investigations in the future.

### DIFF
--- a/apps/arweave/src/ar_http_iface_middleware.erl
+++ b/apps/arweave/src/ar_http_iface_middleware.erl
@@ -190,7 +190,7 @@ handle(<<"GET">>, [<<"recent">>], Req, _Pid) ->
 		true ->
 			{200, #{}, ar_serialize:jsonify(ar_info:get_recent()), Req}
 	end;
-	
+
 handle(<<"GET">>, [<<"is_tx_blacklisted">>, EncodedTXID], Req, _Pid) ->
 	case ar_util:safe_decode(EncodedTXID) of
 		{error, invalid} ->
@@ -2007,6 +2007,15 @@ handle_get_chunk(OffsetBinary, Req, Encoding) ->
 							{{true, Packing}, _StoreID} when RequestedPacking == any ->
 								ok = ar_semaphore:acquire(get_chunk, infinity),
 								{Packing, ok};
+							{true, _StoreID} ->
+								{ok, Config} = application:get_env(arweave, config),
+								case lists:member(pack_served_chunks, Config#config.enable) of
+									false ->
+										{none, {reply, {404, #{}, <<>>, Req}}};
+									true ->
+										ok = ar_semaphore:acquire(get_and_pack_chunk,
+												infinity),
+								{RequestedPacking, ok};
 							{{true, _}, _StoreID} ->
 								{ok, Config} = application:get_env(arweave, config),
 								case lists:member(pack_served_chunks, Config#config.enable) of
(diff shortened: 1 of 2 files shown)


apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2007,6 +2007,15 @@ handle_get_chunk(OffsetBinary, Req, Encoding) ->
 							{{true, Packing}, _StoreID} when RequestedPacking == any ->
 								ok = ar_semaphore:acquire(get_chunk, infinity),
 								{Packing, ok};
+							{true, _StoreID} ->
```
JamesPiechota, line 2010: I think this missing clause was addressed in master. I suggest we go with the master version for now (since it's undergone some testing already)

apps/arweave/src/ar_storage_module.erl
```diff
@@ -193,29 +193,139 @@ get_size(ID, [Module | Modules]) ->
...
+	Packing :: atom().
+
 get(Offset, Packing) ->
 	{ok, Config} = application:get_env(arweave, config),
 	get(Offset, Packing, Config#config.storage_modules, not_found).
 
-%% @doc Return the list of all configured storage modules covering the given Offset.
+%%--------------------------------------------------------------------
+%% @doc Return the list of all configured storage modules covering the
+%% given Offset. This offset is defined by the size of the storage
```
JamesPiechota, line 228: I think in this case the offset is an absolute offset of a byte in the weave. I.e. I'd remove the second sentence "This offset is defined by the size of the storage module (in bytes) and its bucket id (e.g. 51)."

apps/arweave/src/ar_storage_module.erl
```diff
@@ -193,29 +193,139 @@ get_size(ID, [Module | Modules]) ->
...
+%% module (in bytes) and its bucket id (e.g. 51).
+%%
+%% @see get_all/2
+%%
+%% == Examples ==
```
JamesPiechota, line 233: ditto above, I'm not sure the example is quite right. The `Offset` parameter is "simply" the absolute offset of a byte in the weave (i.e. a value from 0 to 269830660858102 as of block 1607197)

apps/arweave/src/ar_storage_module.erl
```diff
@@ -193,29 +193,139 @@ get_size(ID, [Module | Modules]) ->
...
+%%--------------------------------------------------------------------
+%% @doc Return the list of configured storage modules whose ranges
+%% intersect the given interval.
+%%
+%% @see get_all/1
+%%
+%% == Examples ==
+%%
+%% ```
+%% StartOffset = 51*3600000000001.
```
JamesPiechota, line 268: ditto comment above. The offsets aren't generally defined in terms of partition boundaries or storage module boundaries. They are generally offsets of bytes or chunks in the weave.

apps/arweave/src/ar_storage_module.erl
```diff
@@ -193,29 +193,139 @@ get_size(ID, [Module | Modules]) ->
...
 
-%% @doc Return true if the given Offset belongs to at least one storage module.
+%%--------------------------------------------------------------------
+%% @doc Return true if the given Offset belongs to at least one
+%% storage module.
+%%
+%% == Examples ==
+%%
+%% ```
+%% ar_storage_module:has_any(51*3600000000001).
```
JamesPiechota, line 295: ditto above. I'd recommend dropping the `51*XXX` syntax as it *might* be confusing. i.e. one common use of these queries is to, for example, find the storage module containing the recall_byte. The recall_byte is a random value taken from across the full weave that determines the range of chunks to read and hash. When the recall_byte is calculated it si ignorant of partition or storage_module boundaries.

sometimes these methods are also used to respond to a GET /chunk request, in that case a client has provided the offset of a chunk they are interested in, and the server needs to find which storage module to consult to read the chunk. In that case the chunk offset is also not defined in terms of storag modules (e.g. because the client may not know the storage moculde configuration of the server)

PR #665, opened 2024-12-15
refactor: move several replica_2_9 functions into ar_replica_2_9

JamesPiechota:


apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -333,16 +348,17 @@ handle_cast(do_prepare_replica_2_9, State) ->
 			false ->
 				false
 		end,
+	%% XXX: SubChunkStart2 will always be equal to SubChunkStart, right?
```
ldmberman, line 351: I kept it from the sub-chunk based version where the increments were of the sub-chunk size. Also, I kept the two-element cursor format such that it will be slightly easier to re-introduce sub-chunk based iterations if we want it (e.g., might be helpful to support low-memory setups.)

JamesPiechota: Got it! I'll update the comment.

apps/arweave/src/ar_replica_2_9.erl
```diff
@@ -0,0 +1,473 @@
...
+-spec get_entropy_count() -> non_neg_integer().
+get_entropy_count() ->
+	Count = ?PARTITION_SIZE div ?REPLICA_2_9_ENTROPY_SIZE,
+	false = ?PARTITION_SIZE rem ?REPLICA_2_9_ENTROPY_SIZE == 0,
+	%% Add some extra entropies so that the total count is multiple of 32. There are
+    %% 32 sub-chunks per chunk, and each sub-chunk is enciphered with a different entropy. 
+    %% Ensuring we have a multiple of 32 entropies proves very convenient for chunk-by-chunk
+    %% syncing.
+    %% 
+    %% Some entropies will be slightly underused.
```
ldmberman, line 149: This note is outdated. Every bit of the generated entropy is used now.

apps/arweave/src/ar_replica_2_9.erl
```diff
@@ -0,0 +1,473 @@
...
+    %% index for the first sub-chunk in the chunk, for each subsequent sub-chunk the
+    %% expected index is incremented by 1.
+    %% 
+    %% The sector size determines the number of entropy indices. During tests the sector
+    %% size is 3*262144, so the total number of entropy indices is 3*262144 / 8192 = 96 (one
+    %% for each sub-chunk in each sector).
+    
+    %% In tests the strict data split threshold is 262144 * 3, before that offset chunks
+    %% were not padded. So each provided end offset is taken as is. After the threshold each
+    %% offset is padded to a multiple of ?DATA_CHUNK_SIZE (i.e. 262144).
```
ldmberman, line 376: After the threshold is padded to threshold + a multiple of ?DATA_CHUNK_SIZE, to be precise, before determining the entropy bucket (also involves padding.)

JamesPiechota: Updated! https://github.com/ArweaveTeam/arweave/pull/665/commits/eeff94bd8ea1c2d02d219610dee41970ab2d3f54

apps/arweave/src/ar_replica_2_9.erl
JamesPiechota: Updated! https://github.com/ArweaveTeam/arweave/pull/665/commits/eeff94bd8ea1c2d02d219610dee41970ab2d3f54

PR #680, opened 2024-12-30
Test/e2e 2 20241228

JamesPiechota:


apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -809,15 +810,26 @@ generate_entropy_keys(RewardAddr, Offset, SubChunkStart) ->
...
 		Entropies, PaddedEndOffset, SubChunkStartOffset,
-		Partition, Keys, RewardAddr, N, WaitN) ->
+		RangeEnd, Keys, RewardAddr, N, WaitN) ->
 	case take_and_combine_entropy_slices(Entropies) of
 		{<<>>, []} ->
+			%% We've finished processing all the entropies, wait for the writes to complete.
 			wait_store_entropy_processes(WaitN),
 			{ok, N};
 		{ChunkEntropy, Rest} ->
-			true = ar_replica_2_9:get_entropy_partition(PaddedEndOffset) == Partition,
```
ldmberman, line 810: Could you explain this change? The way I see it the original assertion should hold because store_entropy only takes the entropies corresponding to a single partition in the input.

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -389,7 +389,8 @@ handle_cast(do_prepare_replica_2_9, State) ->
 				%% storage modules are commonly set up with.
 				PaddedEndOffset2 = shift_replica_2_9_entropy_offset(
 					PaddedEndOffset, -SliceIndex),
-				store_entropy(Entropies, PaddedEndOffset2, SubChunkStart, Partition,
+				PartitionEnd = (Partition + 1) * ?PARTITION_SIZE,
```
ldmberman, line 392: We should pad PartitionEnd it because we compare it with the padded offset. We should do the same for the CheckRangeEnd check above considering the updates.

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -389,7 +389,8 @@ handle_cast(do_prepare_replica_2_9, State) ->
 				%% storage modules are commonly set up with.
```
ldmberman, line 389: Suggesting a change: `since the 2.9 partition is slightly bigger than the recall partitition...` -> `since the storage module may be configured to be smaller than the partition.`

apps/arweave/src/ar_chunk_storage.erl
JamesPiechota: Yep! The intent of the assertion is the same (unless Ive made a mistake)

Previously Partition was calculated from PartitionEnd and then passed into the function and, I believe, only used in the assertion. 

Now I pass in PartitionEnd directly so I can use it in a guard to stop the recursion. So the assertion *should* still be asserting the same logic (one partition of entropy) but it's doing it using PartitionEnd instead of Partition.

ldmberman: Ah, right, thank you. What do you think about keeping the previous assertion too?

JamesPiechota: Like re add the Partition argument?

ldmberman: Ah, my bad, nevermind, I see RangeEnd replaced Partition so the assertion is changed accordingly.

JamesPiechota: Thanks!

apps/arweave/src/ar_chunk_storage.erl
JamesPiechota: Updated! https://github.com/ArweaveTeam/arweave/pull/680/commits/6811ceae2fa0632e5fe830b34ebb1fd85585af55

apps/arweave/src/ar_chunk_storage.erl
JamesPiechota: Updated! https://github.com/ArweaveTeam/arweave/pull/680/commits/6811ceae2fa0632e5fe830b34ebb1fd85585af55

PR #697, opened 2025-01-29
Chunk cache reimplementation

shizzard:
Chunk cache is reimplemented in a separate module.
Mining worker utilizes new chunk cache.


apps/arweave/src/ar_mining_worker.erl
```diff
@@ -229,14 +264,14 @@ handle_info({garbage_collect, StartTime, GCResult}, State) ->
...
-handle_info({fetched_last_moment_proof, _}, State) ->
+handle_info(?MSG_FETCHED_LAST_MOMENT_PROOF(_), State) ->
     %% This is a no-op to handle "slow" response from peers that were queried by `fetch_poa_from_peers`
     %% Only the first peer to respond with a PoA will be handled, all other responses will fall through to here
     %% an be ignored.
 	{noreply, State};
 
 handle_info(Message, State) ->
-	?LOG_WARNING([{event, unhandled_info}, {module, ?MODULE}, {message, Message}]),
+	?LOG_ERROR([{event, unhandled_info}, {module, ?MODULE}, {message, Message}]),
```
shizzard, line 274: Revert

apps/arweave/include/ar_chunk_cache.hrl
```diff
@@ -0,0 +1,12 @@
+-record(ar_chunk_cache_group, {
```
JamesPiechota, line 1: Does this need a

```
-ifndef(AR_CHUNK_CACHE_HRL).
-define(AR_CHUNK_CACHE_HRL, true).
```

apps/arweave/src/ar_chunk_cache.erl
```diff
@@ -0,0 +1,495 @@
...
+    chunk_cache = ChunkCache0,
+    chunk_cache_size_bytes = ChunkCacheSize0,
+    reserved_chunk_cache_bytes = ReservedChunkCacheBytes0
+  } = Group0) ->
+    case maps:find(ChunkId, ChunkCache0) of
+      {ok, ?CACHE_VALUE(Chunk0, _Meta)} ->
+        {ok, Group0#ar_chunk_cache_group{
+          chunk_cache = maps:put(ChunkId, {Chunk, ChunkMeta}, ChunkCache0),
+          chunk_cache_size_bytes = ChunkCacheSize0 + byte_size(Chunk) - byte_size(Chunk0),
+          reserved_chunk_cache_bytes = max(0, ReservedChunkCacheBytes0 - byte_size(Chunk) + byte_size(Chunk0))
```
JamesPiechota, line 232: Any idea why the `max(0, xxx)` is needed? Is this because we still don't know how the cache size can go negative?

apps/arweave/src/ar_chunk_cache.erl
```diff
@@ -0,0 +1,495 @@
...
+  drop_chunk/3, chunk_exists/3
+]).
+
+
+
+-define(CACHE_GROUPS_LIMIT, 4).
+
+
+
+-define(CACHE_VALUE(Chunk, Meta), {Chunk, Meta}).
```
JamesPiechota, line 17: We may need to think of some different names. Currently we use `Chunk` pretty broadly to represent a full chunk of data (typically 256KiB), but here I think it's being used to represent either a cull Chunk (under spora_2_6) or a SubChunk (under replica_2_9 and composite). I think naming the module ar_chunk_cache still works, but I might suggest changing all the functions and variables to SubChunk... or maybe some other name? I had to keep checking myself as I read through to remind myself the `Chunk` values *weren't* full 256 KiB chunks, but were actually SubChunks

Ditto with `Meta`, we already have the concept of chunk metadata (e.g. in [ar_data_sync](https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_data_sync.erl#L293 and related modules) and it refers to all the proof-related metadata stored next to chunks.

In this case since it seems to store metadata specific to the mining process (e.g. chunk1/chunk2, h1/h2), maybe `MiningMeta` or `MiningContext` or, going back to the [whitepaper](https://nm5d5uvmhjw363xf3idhjrd6nllqqrk76kkjbfpavyfmf7zzddxa.arweave.net/azo-0qw6bb9u5doGdMR-atcIRV_ylJCV4K4Kwv85GO4), maybe `SPOAMeta` / `SPOAContext` or just `SPOA`

JamesPiechota: hm, we could also rename the module to `ar_mining_cache`. Now that I think about it, I think that might be better. We have other places in the code that cache chunks (e.g. `ar_packing_server` when packing or `ar_data_sync` when queueing up chunks to write to disk, etc...), so a more specific name in this cace might help avoid terminology overlap.

apps/arweave/src/ar_chunk_cache.erl
```diff
@@ -0,0 +1,495 @@
...
+
+drop_chunk(GroupId, ChunkId, Cache0) ->
+  case take_chunk(GroupId, ChunkId, Cache0) of
+    {ok, _, Cache1} -> {ok, Cache1};
+    {error, chunk_not_found} -> {ok, Cache0};
+    {error, Reason} -> {error, Reason}
+  end.
+
+
+-spec chunk_exists(GroupId :: term(), ChunkId :: term(), Cache0 :: #ar_chunk_cache{}) ->
```
JamesPiechota, line 202: I don't think this is used anywhere (maybe used to be?)

apps/arweave/src/ar_chunk_cache.erl
```diff
@@ -0,0 +1,495 @@
...
+      error -> {error, chunk_not_found}
+    end
+  end, Cache0).
+
+
+
+-spec drop_chunk(GroupId :: term(), ChunkId :: term(), Cache0 :: #ar_chunk_cache{}) ->
+  {ok, Cache1 :: #ar_chunk_cache{}} | {error, Reason :: term()}.
+
+drop_chunk(GroupId, ChunkId, Cache0) ->
```
JamesPiechota, line 194: I don't think this is used anywhere (maybe used to be?)

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -13,29 +13,43 @@
...
 }).
 
 -define(TASK_CHECK_FREQUENCY_MS, 200).
 -define(STATUS_CHECK_FREQUENCY_MS, 5000).
 
+%%%===================================================================
+%%% Messages
+%%%===================================================================
+
+-define(MSG_RESET_DIFFICULTY(DiffPair), {reset_difficulty, DiffPair}).
```
JamesPiechota, line 40: I think the reason we went with `reset` rather than `reset_difficulty` is that the active session list is also cleared vs. `set_difficulty` that only sets the dificulty.

i.e. `reset` is called when there's a mining session reset and `set_difficulty` is called when it's only the difficult that's being updated without reseting the mining session

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -112,31 +138,41 @@ init({Partition, PackingDifficulty}) ->
 	Name = name(Partition, PackingDifficulty),
 	?LOG_DEBUG([{event, mining_debug_worker_started},
 		{worker, Name}, {pid, self()}, {partition, Partition}]),
-	gen_server:cast(self(), handle_task),
-	gen_server:cast(self(), check_worker_status),
-	prometheus_gauge:set(mining_server_chunk_cache_size, [Partition], 0),
-	{ok, #state{ name = Name, partition_number = Partition,
-			is_pool_client = ar_pool:is_client(), packing_difficulty = PackingDifficulty }}.
+	ChuckCache = ar_chunk_cache:new(),
```
JamesPiechota, line 141:
```suggestion
	ChunkCache = ar_chunk_cache:new(),
```

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -112,31 +138,41 @@ init({Partition, PackingDifficulty}) ->
...
 
 handle_call(Request, _From, State) ->
 	?LOG_WARNING([{event, unhandled_call}, {module, ?MODULE}, {request, Request}]),
 	{reply, ok, State}.
 
-handle_cast({set_difficulty, DiffPair}, State) ->
-	{noreply, State#state{ diff_pair = DiffPair }};
+handle_cast(?MSG_SET_CACHE_LIMITS(SubChunkCacheLimitBytes, VDFQueueLimit), State) ->
+	{noreply, State#state{
+		%% TODO: Convert to bytes
```
JamesPiechota, line 160: This TODO is done, right?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -333,24 +370,24 @@ process_sub_chunk(chunk1, Candidate, SubChunk, State) ->
 process_sub_chunk(chunk2, Candidate, SubChunk, State) ->
 	#mining_candidate{ session_key = SessionKey } = Candidate,
 	Candidate2 = Candidate#mining_candidate{ chunk2 = SubChunk },
-	case cycle_sub_chunk_cache(Candidate2, {chunk2, SubChunk}, State) of
-		{{chunk1, Chunk1, H1}, State2} ->
+	case cycle_sub_chunk_cache(Candidate2, SubChunk, #{chunk2 => true}, State) of
+		{{<<>>, #{chunk1 := true, h1 := H1}}, State2} ->
+			ar_mining_hash:compute_h2(self(), Candidate2#mining_candidate{ h1 = H1 }),
+			%% Decrement 1 for chunk2:
```
JamesPiechota, line 376: The `Decrement xxx` comments may no longer be helpful, they are from when we managed the chunk cache size manually. I believe it's not handled transparently by ar_chunk_cache via the `chunk_cache_size_bytes` and `reserved_chunk_cache_size_bytes` fields

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -750,33 +758,17 @@ remove_tasks(SessionKey, TaskQueue) ->
...
-try_to_reserve_cache_space(SessionKey, State) ->
-	#state{ packing_difficulty = PackingDifficulty } = State,
-	case total_cache_size(State) =< State#state.sub_chunk_cache_limit of
-		true ->
-			%% reserve for both h1 and h2
-			{true, update_sub_chunk_cache_size(
-					2 * ar_block:get_nonces_per_recall_range(PackingDifficulty),
-					SessionKey, State)};
-		false ->
+try_to_reserve_cache_space(SessionKey, #state{
```
JamesPiechota, line 761: The old way of managing the cache size was confusing and should not be emulated. 

However I'm not totally sure I understand the new way. Some context:

The old `try_to_reserve_cache_space` reserved `2 * recall_range_size` under the assumption we would read both recall ranges on each step for each partition. This rarely happens (which is where `do_not_cache` comes in), but even when it happens only 1 out of every 2 chunks would actually be *cached*. 

i.e. if the miner read chunk2 first, it would cache it and and wait until chunk1 was read. At which poitn it would hash chunk1 and then hash chunk2 - chunk1 was never actually stored in the cache.

This means that although we reserved `2 * recall_range_size` optimistically we'd still only end up storgin `recall_range_size` worth of chunks.

However the reason we reserved `2 * recall_range_size` is because even if a chunk isn't cached in the cache, it still exists in the erlang binary memory and takes up memory. So we reserve 2xrecall_range_size to make sure the miner doesn't run out of memory.

i.e. if the user says I want to allocate a total of 10GiB to the mining cache, we want to make sure we never read more than 10GiB of chunks concurrently even if we're only storing 5 GiB of those chunks in the mining cache.

I can't quite trace through the new logic to know if this promise still holds. I.e. If we set a chunk cache limit of 10 GiB, will the new process ensure that only 10 GiB of mining chunks will be read at a time or does it only ensure that only 10 GiB of mining chunks will be *cached* at a time (and perhaps 20 GiB will actually be held in memory)?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -333,24 +370,24 @@ process_sub_chunk(chunk1, Candidate, SubChunk, State) ->
...
+		{{<<>>, #{chunk1 := true, h1 := H1}}, State2} ->
+			ar_mining_hash:compute_h2(self(), Candidate2#mining_candidate{ h1 = H1 }),
+			%% Decrement 1 for chunk2:
+			%% we're computing h2 for a peer so chunk1 was not previously read or cached
+			%% on this node
+			State2;
+		{{Chunk, #{chunk1 := true, h1 := H1}}, State2} ->
 			ar_mining_hash:compute_h2(
-				self(), Candidate2#mining_candidate{ chunk1 = Chunk1, h1 = H1 }),
+				self(), Candidate2#mining_candidate{ chunk1 = Chunk, h1 = H1 }),
```
JamesPiechota, line 382: See comment [below](https://github.com/ArweaveTeam/arweave/pull/697/files#r1949811181)

The reason we decremented 2 here even though we had only removed a single chunk from the cache, is we knew that we had finished hashing both chunk1 and chunk2 so they would shortly be garbage collected and purged from memory (even though only chunk1 had ever been added to the cache)

Does that scenario translate to the new logic?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -427,40 +470,34 @@ handle_task({computed_h0, Candidate, _ExtraArgs}, State) ->
...
 			Range2Exists = ar_mining_io:read_recall_range(
 					chunk2, self(), Candidate3, RecallRange2Start),
 			case Range2Exists of
-				true -> 
+				true ->
 					State;
 				false ->
 					%% Release just the Range2 cache space we reserved with
 					%% try_to_reserve_cache_space/2
-					State2 = update_sub_chunk_cache_size(
```
JamesPiechota, line 441: Related to the comment [below](https://github.com/ArweaveTeam/arweave/pull/697/files#r1949811181) and [above](https://github.com/ArweaveTeam/arweave/pull/697/files#r1949813511)

Part of the reason we updated teh chunk cache size here is to make sure that the miner doesn't intiate too many concurrent reads.

e.g. let's say the miner is running slow for a few seconds - maybe some automated OS process is causing all reads from disks to take a lot longer than they should - maybe several second to complete a read. If during that period several VDF steps come in, we may not want the miner to initiate several reads from disk. because if it does and then all those reads come in at the same time, it may cause the miner to OOM.

Instead we "reserve" chunk cache space by incrementing the size *before* the chunks are read form disk. This makes sure that we won't start processing new VDF steps until those reads are completed and the chunks purged from memory.

I don't full understand the new logic, but I feel like we should have an `ar_chunk_cache:reserve` or similar here.

JamesPiechota: Ah I see the reserve is earlier. All good. But the reserve is currently only reserving 1 recall range. I think it should be reserving 2x recall range?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -478,12 +515,12 @@ handle_task({computed_h1, Candidate, _ExtraArgs}, State) ->
 					ok
 			end,
 			{ok, Config} = application:get_env(arweave, config),
-			case cycle_sub_chunk_cache(Candidate, {chunk1, Chunk1, H1}, State2) of
+			case cycle_sub_chunk_cache(Candidate, Chunk1, #{chunk1 => true, h1 => H1}, State2) of
```
JamesPiechota, line 518: I think this now needs an `{error, Error}` case clause

apps/arweave/src/ar_chunk_cache.erl
shizzard: Well, no, it is a reserved size. 
The chunk cache size is reported as a sum of all cache groups (or sessions), and each session size is a sum of the actual data size plus reserved size. 
So the `max/2` here is to ensure the reserved size will not go below zero, thus allowing the cache overflow.

apps/arweave/src/ar_chunk_cache.erl
shizzard: No, it isn't, this function is here for API consistency.

apps/arweave/src/ar_chunk_cache.erl
shizzard: No, it isn't, this function is here for API consistency.

apps/arweave/src/ar_mining_worker.erl
shizzard: Got it.

apps/arweave/src/ar_chunk_cache.erl
JamesPiechota: I think I understand. By design the reserved space shouldn't go negative - the fact that we had a negative cache size in the old implementation was due to a bug that we may not have tracked down yet.

Does the new implementation expect there to be a negative reserved size? Or is it just a hold over from the old bug? If it's a hold over it might be worth logging something when it goes negative, cause it could mean we still have an issue.

shizzard: No, that's expected behavior: you reserved X bytes, you push some data into it, but at some point the reserved space is running out, but you still can have some space left in the cache.

apps/arweave/src/ar_chunk_cache.erl
JamesPiechota: Hm, okay binary cache works... but the design is pretty specific to the mining needs (E.g. concept of groups and "reserved" size), so I'd still lean towards naming it something specific to mining until we have another use case for it (at which point we can rename), but I don't yet have a clear idea where else it could be used without being reworded.

shizzard: Okay.,

apps/arweave/src/ar_chunk_cache.erl
JamesPiechota: My preference would be to remove all unused code. My experience is that unused code just gets in the way - i.e. more time spent understanding it, greater chance that it doesn't do what it's supposed to (since it isn't being used, etc..)

apps/arweave/src/ar_chunk_cache.erl
JamesPiechota: Hm, okay then we'll just need to make sure this new design works with the old intent which was partially to make sure the node didn't initiate too many concurrent reads at once

I.e. the old cache would essentially pre-allocate 2 full recall ranges of chunks and then decrement the cache size as the chunks were used and purged. So we never used the cache without first reserving space, and never used beyond what was reserved. The purpose this served was to make sure we don't kick off too many concurrent reads and end up blowing out memeory. e.g. if the cache is 400 MiB and a recall range is 100 MiB, we should only be able to initiate four recall ranges reads at a time - even if something happens and there ends up being a long delay between initiating a read and receiving te chunks.. During that delay period the cache is actually empty, but because we reserved the the space for all of them before we started, the node will know not ot initiate any more reads until those reads complete (i.e. the pre-alocated space in the cache is already at the limit)

Under that model the "reserved size" = "cache size" and neither *should* go negative

apps/arweave/src/ar_mining_worker.erl
shizzard: There are two answers, and none of those is right (...fully).
Short answer: no, the promise is not held.
Long answer: We can reserve 2x recall range, but this technique will not guarantee anything anyway.

In fact, this promise is not held today, as the chunks are send beyond mining process boundaries.
In order to guarantee that this promise will hold, we need to ensure that (given we're on the cache limit right now) we do not read a new piece of data (chunk, subchunk, you name it) while older data is not freed from memory. It is not sufficient to remove the binary from the cache and trigger the garbage collect routine: other processes might still have a valid reference to that binary, and it will not be freed. For example, if we have a CPU time shortage, it will cause memory bloat, and none of the implementations (previous, and this new one) is protecting us from this.

The only way to make this work is to utilize old erlanger trickery, that is used when working with large binary blobs: I call it binary guarding.
The idea is simple if you understand how large binaries work (I believe you do, we discussed it before), but just to make explanation full: large binaries are stored on the common heap, and are refcounted; it is necessary to remove all the references to that binary for it to be freed. So with this binary guarding we're limiting the number of references to the extreme low: just one.
When the binary blob "enters" the VM, we're allowing a single short-living process to touch it, thus creating a single reference. This process is guarding this binary blob, not allowing other processes to touch it. If any other process needs to do something with this large binary, it calls the guard process, sending him a closure to operate on the binary, getting back the result of the closure reduction. If you need to send a binary from one process to another, you send the guard process pid instead of the actual binary reference.
In this case freeing the memory is very simple and fast: you kill the guard process, its memory gets reclaimed, and the binary blob refcount immediately drops to zero, freeing the common heap space as well. No global garbage collection needed in this case.

This approach is very elaborate in our case, but I believe this is the only way to ensure we're actually hard-limiting the cache size. In all other cases we're only soft-limiting it, and it may (and sometimes will) grow beyond the configured size.
So if we will reserve 2x recall range size, we will reduce the chance of such event, but we are also effectively reducing the number of chunks that we can read by two.

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -577,22 +473,47 @@ handle_task({computed_h2, Candidate, _ExtraArgs}, State) ->
...
 			PoA2 = case ar_mining_server:prepare_poa(poa2, Candidate, #poa{}) of
-				{ok, PoA} ->
-					PoA;
+				{ok, PoA} -> PoA;
 				{error, _Error} ->
 					%% Fallback. This will probably fail later, but prepare_poa/3 should
 					%% have already printed several errors so we'll continue just in case.
+					%% df: Is this the right fallback?..
 					#poa{ chunk = Chunk2 }
 			end,
```
shizzard, line 495: Perhaps, should move the PoA part into `ar_coordination` here. This is the only place in whole mining worker that touches the PoA logic.

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -112,31 +137,40 @@ init({Partition, PackingDifficulty}) ->
...
 
 handle_call(Request, _From, State) ->
 	?LOG_WARNING([{event, unhandled_call}, {module, ?MODULE}, {request, Request}]),
 	{reply, ok, State}.
 
-handle_cast({set_difficulty, DiffPair}, State) ->
-	{noreply, State#state{ diff_pair = DiffPair }};
+handle_cast(?MSG_SET_CACHE_LIMITS(SubchunkCacheLimitBytes, VDFQueueLimit), State) ->
+	{noreply, State#state{
+		%% TODO: Convert to bytes
```
JamesPiechota, line 158:
```suggestion
```

apps/arweave/src/ar_mining_io.erl
```diff
@@ -122,6 +127,15 @@ handle_call({read_recall_range, WhichChunk, Worker, Candidate, RecallRangeStart}
 	end,
 	{reply, ThreadFound, State};
 
+handle_call({is_recall_range_readable, Candidate, RecallRangeStart}, _From, State) ->
```
JamesPiechota, line 130: If possible would be good to dedupe the duplicate code between this handler and `read_recall_range`

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -2,40 +2,53 @@
...
+	latest_vdf_step_number = 0,
+	is_pool_client = false,
+	h1_hashes = #{},
+	h2_hashes = #{}
 }).
 
 -define(TASK_CHECK_FREQUENCY_MS, 200).
 -define(STATUS_CHECK_FREQUENCY_MS, 5000).
 
+-define(VDF_STEP_CACHE_KEY(CacheRef, Nonce), {CacheRef, Nonce}).
```
JamesPiechota, line 33: is this the Mining Cache Key? If so I'd maybe change the name? Or are we tracking cache entries by vdf step explicitly?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -112,31 +137,40 @@ init({Partition, PackingDifficulty}) ->
 	Name = name(Partition, PackingDifficulty),
 	?LOG_DEBUG([{event, mining_debug_worker_started},
 		{worker, Name}, {pid, self()}, {partition, Partition}]),
-	gen_server:cast(self(), handle_task),
-	gen_server:cast(self(), check_worker_status),
-	prometheus_gauge:set(mining_server_chunk_cache_size, [Partition], 0),
-	{ok, #state{ name = Name, partition_number = Partition,
-			is_pool_client = ar_pool:is_client(), packing_difficulty = PackingDifficulty }}.
+	ChuckCache = ar_mining_cache:new(),
```
JamesPiechota, line 140:
```suggestion
	ChunkCache = ar_mining_cache:new(),
```

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -112,31 +137,40 @@ init({Partition, PackingDifficulty}) ->
...
+	gen_server:cast(self(), ?MSG_CHECK_WORKER_STATUS),
+	{ok, report_chunk_cache_metrics(State0)}.
 
 handle_call(Request, _From, State) ->
 	?LOG_WARNING([{event, unhandled_call}, {module, ?MODULE}, {request, Request}]),
 	{reply, ok, State}.
 
-handle_cast({set_difficulty, DiffPair}, State) ->
-	{noreply, State#state{ diff_pair = DiffPair }};
+handle_cast(?MSG_SET_CACHE_LIMITS(SubchunkCacheLimitBytes, VDFQueueLimit), State) ->
```
JamesPiechota, line 156:
```suggestion
handle_cast(?MSG_SET_CACHE_LIMITS(ChunkCacheLimitBytes, VDFQueueLimit), State) ->
```

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -199,24 +232,23 @@ handle_cast(handle_task, #state{ task_queue = Q } = State) ->
 			end
 	end;
 
-handle_cast({remove_sub_chunks_from_cache, SubChunkCount, Candidate}, State) ->
-	{noreply, remove_sub_chunks_from_cache(Candidate, SubChunkCount, State)};
+handle_cast(?MSG_REMOVE_SUB_CHUNKS_FROM_CACHE(SubchunkCount, Candidate), State) ->
```
JamesPiechota, line 235:
```suggestion
handle_cast(?MSG_REMOVE_SUB_CHUNKS_FROM_CACHE(SubChunkCount, Candidate), State) ->
```

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -199,24 +232,23 @@ handle_cast(handle_task, #state{ task_queue = Q } = State) ->
...
+	{noreply, remove_subchunks_from_cache(Candidate, SubchunkCount, State)};
```
JamesPiechota, line 236:
```suggestion
	{noreply, remove_sub_chunks_from_cache(Candidate, SubChunkCount, State)};
```

JamesPiechota: I think we could go either way with `SubChunk` vs `Subchunk`, but in the rest of the codebase we treat it as two words, so best to keep it hat way for now

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -199,24 +232,23 @@ handle_cast(handle_task, #state{ task_queue = Q } = State) ->
...
 
-handle_cast(check_worker_status, State) ->
+handle_cast(?MSG_CHECK_WORKER_STATUS, State) ->
 	maybe_warn_about_lag(State#state.task_queue, State#state.name),
-	maybe_warn_about_stale_chunks(State),
```
JamesPiechota, line 207: `maybe_warn_about_stale_chunks` was just a stop gap when we didn't know what was causing the chunk cache to go negative, right? It's no longer needed?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -601,6 +545,144 @@ handle_task({compute_h2_for_peer, Candidate, _ExtraArgs}, State) ->
 %%% Private functions.
 %%%===================================================================
 
+process_chunks(WhichChunk, Candidate, RangeStart, ChunkOffsets, State) ->
+	PackingDifficulty = Candidate#mining_candidate.packing_difficulty,
+	SubchunksPerRecallRange = ar_block:get_max_nonce(PackingDifficulty),
```
JamesPiechota, line 550:
```suggestion
	SubChunksPerRecallRange = ar_block:get_max_nonce(PackingDifficulty),
```

ditto for all `Subchunks`

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,281 +277,181 @@ terminate(_Reason, _State) ->
...
-	end.
+%% @doc Handle the `compute_h0` task.
+%% Indicates that the VDF step has been computed.
+handle_task({compute_h0, Candidate, _ExtraArgs}, State) ->
+	#state{
+		latest_vdf_step_number = LatestVDFStepNumber,
+		vdf_queue_limit = VDFQueueLimit
+	} = State,
+	#mining_candidate{ step_number = StepNumber } = Candidate,
+	State2 = report_and_reset_hashes(State),
```
JamesPiechota, line 305: I forget, why did we decide to move the `try_to_reserve_cache_space` out of `compute_h0` and into `computed_h0`?

It does open the possibility of the node computing a bunch of h0s that it then has to throw away (i.e. computes them but then can't process them further cause th cache is full, but by the time the cache has space the previously computed h0s are now too old to use).

I'm not sure how big an impact it will have in practice - so if there's a good reason to delay the cache check until later it's probably fine. But if it's all the same, checking the cache earlier should reduce some amount of unnecessary work. And computing h0 isn't *expensive* but it's not cheap (it's a randomx operation)

JamesPiechota: Okay yeah I think we do have to do the cache check here as this is the only place we check `StepNumber >= LatestVDFStepNumber - VDFQueueLimit`

By the time we make it to `computed_h0` we assume we're going to proceed through the whole pipeline, which means we may end up reading chunks and performing hashes for a too-old step number.

Best is to have the cache check, VDFQueueLimit check, and the `add_delayed_task` line here, in `compute_h0`. That way by the time we reach computed_h0 we know the step number is fresh and the cache has space.

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,281 +277,181 @@ terminate(_Reason, _State) ->
...
+			%% Both recall ranges are readable, so we need to reserve cache space for both.
+			case try_to_reserve_cache_space(2, Candidate2#mining_candidate.session_key, State) of
+				{true, State1} ->
+					%% Read the recall ranges; the result of the read will be reported by the `chunk1` and `chunk2` tasks.
+					ar_mining_io:read_recall_range(chunk1, self(), Candidate2, RecallRange1Start),
+					ar_mining_io:read_recall_range(chunk2, self(), Candidate2, RecallRange2Start),
+					{noreply, State1};
+				false ->
+					%% We don't have enough cache space to read the recall ranges, so we'll try again later.
+					add_delayed_task(self(), computed_h0, Candidate),
```
JamesPiechota, line 339: See above comment. By the time we retry this task, the vdf step may be too old in which case we should abandon it so that the miner can focus on newer VDF steps (the older a VDF step is the greater the likelihood that any solution derived from it will be orphaned/invalid due to another miner mining a block with a higher step number)

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -611,257 +693,187 @@ passes_diff_checks(SolutionHash, IsPoA1, Candidate, State) ->
...
+	case ar_mining_cache:with_cached_value(
+		?VDF_STEP_CACHE_KEY(Candidate#mining_candidate.cache_ref, Nonce),
+		Candidate#mining_candidate.session_key,
+		State#state.chunk_cache,
+		fun(CachedValue) -> {ok, CachedValue#ar_mining_cache_value{chunk2_missing = true}} end
+	) of
+		{ok, ChunkCache1} ->
+			mark_chunk2_missing(Nonce + 1, SubchunksPerRecallRange, Candidate, State#state{ chunk_cache = ChunkCache1 });
+		{error, Reason} ->
+			%% NB: this clause may cause a memory leak, because mining worker will wait for
```
JamesPiechota, line 829: Given the function used here, there's no way this clause can be hit, right? i.e. 

```
fun(CachedValue) -> {ok, CachedValue#ar_mining_cache_value{chunk2_missing = true}} end
```

never returns {error, Reason}

If so, maybe we drop the clause and comment? Mostly just thinking of future engineers going through the code looking for memory leaks

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,281 +277,181 @@ terminate(_Reason, _State) ->
...
+			ar_mining_stats:h1_solution(),
+			%% Update the cache to store h1.
+			case ar_mining_cache:with_cached_value(
+				?VDF_STEP_CACHE_KEY(Candidate#mining_candidate.cache_ref, Candidate#mining_candidate.nonce),
+				Candidate#mining_candidate.session_key,
+				State2#state.chunk_cache,
+				fun(CachedValue) -> {ok, CachedValue#ar_mining_cache_value{h1 = H1}} end
+			) of
+				{ok, ChunkCache1} -> State2#state{ chunk_cache = ChunkCache1 };
+				{error, Reason1} ->
```
JamesPiechota, line 396: Is there any way we can hit `{error, Reason1}`? If not I'd say we remove the clause for now

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,281 +277,181 @@ terminate(_Reason, _State) ->
...
-					ar_mining_hash:compute_h2(
-						self(), Candidate#mining_candidate{ chunk2 = Chunk2 }),
-					%% Decrement 2 for chunk1 and chunk2:
-					%% 1. chunk2 was previously read and cached
-					%% 2. chunk1 that was just read and used to compute H1
-					{noreply, update_sub_chunk_cache_size(-2, SessionKey, State3)}
-			end
+			ar_mining_stats:h1_solution(),
+			%% Update the cache to store h1.
+			case ar_mining_cache:with_cached_value(
```
JamesPiechota, line 389: Do we need this here since we're also doing it down here: https://github.com/ArweaveTeam/arweave/pull/697/files#diff-7d9b88c10de9e64afe9594e33a3e801b5b6a9421c6aa51f4d7e529a26a05fa9eR409

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: Ah I see, it's the `{error, cache_limit_exceeded}`

Previously we let the cache exceed the limit in order to avoid this sort of situation. I.e. by pre-allocating the cache before reading recall ranges we'll generally avoid the worst of the overruns. Occasionally we might overrun slightly, but never by more than 2 recall ranges (I don't think).

Do you have a sense of how often we'll hit the `cache_limit_exceeded` now? previously it was fairly often, if it's still fairly often then I suggest we go back to th eold behavir and allow the cache limit to be exceeded (otherwise the miner will likely see thousands of errors)

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: agree, I don't like having the poa logic here, makes sense to refactor it into ar_coordination or ar_mining_server

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -541,53 +473,65 @@ handle_task({computed_h2, Candidate, _ExtraArgs}, State) ->
...
-			%% *even if* this puts us over the chunk cache limit. This may have to be
-			%% revisited later if we find that this causes unacceptable memory bloat.
-			State2 = update_sub_chunk_cache_size(
-				ar_block:get_nonces_per_recall_range(PackingDifficulty), SessionKey, State),
-			%% First flag all nonces in the range as do_not_cache, then cache the specific
-			%% nonces included in the H1 list. This will make sure we don't cache the chunk2s
-			%% that are read for the missing nonces.
-			State3 = do_not_cache(Candidate3, State2),
-			{noreply, cache_h1_list(Candidate3, H1List, State3)};
+			State2 = mark_chunk2_missing(Candidate3, State),
```
JamesPiechota, line 533: is marking chunk2 missing just a sanity check? It should be missing by default right?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -611,257 +693,187 @@ passes_diff_checks(SolutionHash, IsPoA1, Candidate, State) ->
...
+%% @doc Remove SubchunkCount sub-chunks from the cache starting at
 %% Candidate#mining_candidate.nonce.
-remove_sub_chunks_from_cache(#mining_candidate{ cache_ref = CacheRef }, 0, State)
-		when CacheRef /= not_set ->
+remove_subchunks_from_cache(_Candidate, 0, State) ->
 	State;
-remove_sub_chunks_from_cache(#mining_candidate{ cache_ref = CacheRef } = Candidate,
-		SubChunkCount, State) when CacheRef /= not_set ->
+remove_subchunks_from_cache(#mining_candidate{ cache_ref = CacheRef } = Candidate,
+		SubchunkCount, State) when CacheRef /= not_set ->
```
JamesPiechota, line 840:
```suggestion
		SubChunkCount, State) when CacheRef /= not_set ->
```

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -611,257 +693,187 @@ passes_diff_checks(SolutionHash, IsPoA1, Candidate, State) ->
...
-%% @doc Remove SubChunkCount sub-chunks from the cache starting at
+%% @doc Remove SubchunkCount sub-chunks from the cache starting at
 %% Candidate#mining_candidate.nonce.
-remove_sub_chunks_from_cache(#mining_candidate{ cache_ref = CacheRef }, 0, State)
-		when CacheRef /= not_set ->
+remove_subchunks_from_cache(_Candidate, 0, State) ->
 	State;
-remove_sub_chunks_from_cache(#mining_candidate{ cache_ref = CacheRef } = Candidate,
-		SubChunkCount, State) when CacheRef /= not_set ->
+remove_subchunks_from_cache(#mining_candidate{ cache_ref = CacheRef } = Candidate,
```
JamesPiechota, line 839:
```suggestion
remove_sub_chunks_from_cache(#mining_candidate{ cache_ref = CacheRef } = Candidate,
```

apps/arweave/src/ar_mining_worker.erl
shizzard: We have a session key, which contains multiple VDF sessions, and a VDF key, which contains multiple recall ranges (one or two), several chunks each.

apps/arweave/src/ar_mining_worker.erl
shizzard: Not sure about that, I think I removed the warning we added earlier. If we don't need it I'll remove it, but I need to recalibrate the logic.

apps/arweave/src/ar_mining_worker.erl
shizzard: `CacheLimitBytes` would be better, probably?

apps/arweave/src/ar_mining_worker.erl
shizzard: There is still a way to get an error here, for example, if the session was removed (`{error, session_not_found}`).

apps/arweave/src/ar_mining_worker.erl
shizzard: This will conflict with the commitment to hold the cache size within the configured size.

apps/arweave/src/ar_mining_worker.erl
shizzard: Ah, it was `do_not_cache` before. It was used not to cache the `chunk2`s that we will not use, only those that correspond to received `h1`s. It is now broken I think, I will need to handle this case in `chunk2` event.

apps/arweave/src/ar_mining_worker.erl
shizzard: I mean, we need to formulate this commitment, because for now I don't understand what should we do with the cache size in different setups.

shizzard: Resolved in private conversation, the logic is to reserve the proper number of subchunks of space in the cache and delay the task just as we do with `compute_h0`.

rebar.config
```diff
@@ -3,7 +3,6 @@
 	{jiffy, {git, "https://github.com/ArweaveTeam/jiffy.git", {ref, "f25a120f02951b9fd618d1f8f6343fdb8dbe133f"}}},
 	{gun, "1.3.3"},
 	{cowboy, "2.10.0"},
-	{graphql, {git, "https://github.com/shopgun/graphql-erlang.git", {branch, "master"}}},
```
JamesPiechota, line 6: I thought we removed this in a previous commit - maybe we need to rebase this branch? (weird that github doesn't seem to think so)

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -85,24 +117,17 @@ computed_hash(Worker, computed_h2, H2, Preimage, Candidate) ->
...
-%% @doc Returns true if the mining candidate belongs to a valid mining session. Always assume
-%% that a coordinated mining candidate is valid (its cm_lead_peer is set)
-is_session_valid(_State, #mining_candidate{ cm_lead_peer = Peer })
-		when Peer /= not_set ->
-	true;
-is_session_valid(
-		#state{ active_sessions = Sessions },
-		#mining_candidate{ session_key = SessionKey }) ->
-	sets:is_element(SessionKey, Sessions).
+-spec set_cache_limits(Worker :: pid(), SubchunkCacheLimitBytes :: non_neg_integer(), VDFQueueLimit :: non_neg_integer()) -> ok.
```
JamesPiechota, line 124:
```suggestion
-spec set_cache_limits(Worker :: pid(), SubChunkCacheLimitBytes :: non_neg_integer(), VDFQueueLimit :: non_neg_integer()) -> ok.
```

For consistency with the rest of the code, I think we should do `SubChunk` rather than `Subchunk` (there are few places to update, but I'll just comment here)

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: CacheLimitBytes sounds good.

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: Or CacheLimitBytes as you suggest below: https://github.com/ArweaveTeam/arweave/pull/697/files#r1966570633

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: bump

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,362 +284,412 @@ terminate(_Reason, _State) ->
...
-	State2 = report_hashes(State),
-	State4 = case try_to_reserve_cache_space(SessionKey, State2) of
-		{true, State3} ->
+	#state{
+		latest_vdf_step_number = LatestVDFStepNumber,
+		vdf_queue_limit = VDFQueueLimit
+	} = State,
+	#mining_candidate{ step_number = StepNumber } = Candidate,
+	State1 = report_and_reset_hashes(State),
+	% Check if we need to compute h0 early
```
JamesPiechota, line 313:
```suggestion
	% Only mine recent VDF Steps
```

Previously the intent of this line wasn't to compute h0 early, but decide whther or not to compute this VDF step at all. If StepNumber is too old, then we just drop it so that we don't waste CPU and DiskIO hashing an old VDF step that's likely to generate solutions with a higher probability of being rejected by peers.

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: Oh, perhaps this comment from a while back got lost. This is the point I'm referencing [here](https://github.com/ArweaveTeam/arweave/pull/697/files#r1999936725)

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,362 +284,412 @@ terminate(_Reason, _State) ->
...
+			case try_to_reserve_cache_space(1, Candidate2#mining_candidate.session_key, State) of
+				{true, State1} ->
+					%% Read the recall range; the result of the read will be reported by the `chunk1` task.
+					ar_mining_io:read_recall_range(chunk1, self(), Candidate2, RecallRange1Start),
+					%% Mark chunk2 as missing, not to wait for it to arrive.
+					State2 = mark_chunk2_missing(Candidate2, State1),
+					State2;
+				false ->
+					%% We don't have enough cache space to read the recall range, so we'll try again later.
+					add_delayed_task(self(), computed_h0, Candidate),
```
JamesPiechota, line 360: There may be a potential race condition here - but I'm not totally sure it matters. since a miner can sync+pack while mining, it's possible that the first time they get here only recall_range_1 exists, but when the delayed task is processed, both recall_range_1 and recall_range_2 exist. I think this is probably fine? Just calling it out in case.

Maybe be moot if we revert back to delaying the `compute_h0` rather than `computed_h0` task

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: Bumping this. I think your point was a good one. But maybe we defer to another PR?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,362 +284,412 @@ terminate(_Reason, _State) ->
...
-					%% 2. chunk1 that was just read and used to compute H1
-					{noreply, update_sub_chunk_cache_size(-2, SessionKey, State3)}
-			end
+			ar_mining_stats:h1_solution(),
+			%% Update the cache to store h1.
+			case ar_mining_cache:with_cached_value(
+				?VDF_STEP_CACHE_KEY(Candidate#mining_candidate.cache_ref, Candidate#mining_candidate.nonce),
+				Candidate#mining_candidate.session_key,
+				State1#state.chunk_cache,
+				fun(CachedValue) -> {ok, CachedValue#ar_mining_cache_value{h1 = H1}} end
```
JamesPiechota, line 400: I think this logic differs from before - not a problem, I just can't quite wrap my head around it.

Previously if we found a solution on H1 we didn't even try to find a solution for H2 (e.g. this [comment](https://github.com/ArweaveTeam/arweave/pull/697/files#diff-7d9b88c10de9e64afe9594e33a3e801b5b6a9421c6aa51f4d7e529a26a05fa9eL467-L469))

But now it looks like we're caching H1, presumably so that we can use it to try for H2?

I honestly can't think of a reason why we wouldn't try for H2, just in case, but I haven't thought it all the way through (e.g. there's no way two solutions at the same VDF step will be accepted, so we know *some* work will be wasted - but then again it's not a lot of work). 

What's your thinking behind caching H1 and continuing rather than discarding it as we did before?

JamesPiechota: Oh actually, do we even need to cache H1 here? Since it looks like we do it down below here: https://github.com/ArweaveTeam/arweave/pull/697/files#diff-7d9b88c10de9e64afe9594e33a3e801b5b6a9421c6aa51f4d7e529a26a05fa9eR416

```
		fun
			(#ar_mining_cache_value{chunk2_missing = true}) ->
				%% This node does not store chunk2. If we're part of a coordinated
				%% mining set, we can try one of our peers, otherwise we're done.
				case Config#config.coordinated_mining of
					false -> ok;
					true ->
						DiffPair = case get_partial_difficulty(State2, Candidate) of
								not_set -> get_difficulty(State2, Candidate);
								PartialDiffPair -> PartialDiffPair
							end,
						ar_coordination:computed_h1(Candidate, DiffPair)
				end,
				%% Remove the cached value from the cache.
				{ok, drop};
			(#ar_mining_cache_value{chunk2 = undefined} = CachedValue) ->
				%% chunk2 hasn't been read yet, so we cache chunk1 and wait for it.
				{ok, CachedValue#ar_mining_cache_value{h1 = H1}};
			(#ar_mining_cache_value{chunk2 = Chunk2} = CachedValue) ->
				%% chunk2 has already been read, so we can compute H2 now.
				ar_mining_hash:compute_h2(self(), Candidate#mining_candidate{ chunk2 = Chunk2 }),
				{ok, CachedValue#ar_mining_cache_value{h1 = H1}}
		end
	```

File: apps/arweave/src/ar_mining_worker.erl [hunk 38]
```diff
@@ -245,362 +284,412 @@ terminate(_Reason, _State) ->
...
-			%% *even if* this puts us over the chunk cache limit. This may have to be
-			%% revisited later if we find that this causes unacceptable memory bloat.
-			State2 = update_sub_chunk_cache_size(
-				ar_block:get_nonces_per_recall_range(PackingDifficulty), SessionKey, State),
-			%% First flag all nonces in the range as do_not_cache, then cache the specific
-			%% nonces included in the H1 list. This will make sure we don't cache the chunk2s
-			%% that are read for the missing nonces.
-			State3 = do_not_cache(Candidate3, State2),
-			{noreply, cache_h1_list(Candidate3, H1List, State3)};
+			State1 = mark_chunk2_missing(Candidate3, State),
```
JamesPiechota, line 540: why do we mark chunk2 missing here? I'm still working to sort out the new cache state (which is much better than the confusing cycle_chunk_cache stuff) - I thought we only marked chunk2 missing if Range2 does not exist

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: Also in this case I do think "NoncesPerRecallRange" or something like that might be better. In spora_2_6 there's one nonce per chunk in replica.2.9 there's one nonce per subchunk - so "SubChunksPerRecallRange" isn't totally accurate for spora_2_6 mining, but NoncesPerRecallRange is

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,362 +284,412 @@ terminate(_Reason, _State) ->
...
+		chunk2 -> Candidate#mining_candidate.partition_number2
+	end, Count),
+	State;
+process_chunks(
+	WhichChunk, Candidate, RangeStart, Nonce, SubchunksPerChunk,
+	SubchunksPerRecallRange, [], SubchunkSize, Count, State
+) ->
+	%% No more ChunkOffsets means no more chunks have been read. Iterate through all the
+	%% remaining nonces and remove the full chunks from the cache.
+	State2 = remove_subchunks_from_cache(Candidate#mining_candidate{ nonce = Nonce }, SubchunksPerChunk, State),
```
JamesPiechota, line 581:
```suggestion
	State2 = remove_sub_chunks_from_cache(Candidate#mining_candidate{ nonce = Nonce }, SubchunksPerChunk, State),
```

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,362 +284,412 @@ terminate(_Reason, _State) ->
...
+		WhichChunk, Candidate, RangeStart, Nonce + SubchunksPerChunk,
+		SubchunksPerChunk, SubchunksPerRecallRange, [], SubchunkSize, Count, State2
+	);
+process_chunks(
+	WhichChunk, Candidate, RangeStart, Nonce, SubchunksPerChunk,
+	SubchunksPerRecallRange, [{ChunkEndOffset, Chunk} | ChunkOffsets], SubchunkSize, Count, State
+) ->
+	SubchunkStartOffset = RangeStart + Nonce * SubchunkSize,
+	ChunkStartOffset = ChunkEndOffset - ?DATA_CHUNK_SIZE,
+	case {SubchunkStartOffset < ChunkStartOffset, SubchunkStartOffset >= ChunkEndOffset} of
```
JamesPiechota, line 593:
```suggestion
	case {SubChunkStartOffset < ChunkStartOffset, SubChunkStartOffset >= ChunkEndOffset} of
```

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: I'm still wrestling with this name. Each VDF step contains (usually) 2 full recall ranges per partition. But since the VDF_STEP_CACHE_KEY includes the Nonce as an argument, it's actually representing only part of the data handled in a VDF_STEP I think?

I think what's confusing me is we have 3 concepts:
1. a VDF session which contains multiple vdf steps
2. a VDF step which includes multiple recall ranges
3. a recall range which contains multiple chunks / nonces

I believe VDF_STEP_CACHE_KEY is used to index into 3rd category, but whenever I read it I first think it's referencing the 2nd category so I get confused. I do think we should name it something *other than* VDF_STEP_CACHE_KEY. IMO it does feel like it's the "main" cache key - so CACHE_KEY or MINING_CACHE_KEY or something simliar. Since it stores an `#ar_mining_cache_value` I think MINING_CACHE_KEY might work well (i..e Key => Value paradigm)

apps/arweave/src/ar_mining_worker.erl
shizzard: Yea, the logic here was that previously we would cache both chunks in the same place, assuming that chunk2 would replace chunk1. So previously we marked the whole range as `do_not_cache` and then, on the next line, we stored the H1s with `cache_h1_list`, replacing these `do_not_cache` flags with actual H1s; all the range was marked as `do_not_cache`, except for the chunks we have H1 for.
During rework that the logic changed, and now we mark the whole range as `chunk2_missing` (separate record flag), and then traverse the list of H1s, storing those we just received from peer. The only thing missing here is that we need to also reset the `chunk2_missing` flag to `false` for those entries we have H1 for. I'll fix it. 
Good call.

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: @shizzard This one got lost cause I forgot to unresolve. After completing the review, I do think we should change the name of the cache key

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -245,362 +283,399 @@ terminate(_Reason, _State) ->
...
+	process_all_sub_chunks(WhichChunk, Rest, Candidate2, Nonce + 1, State2);
+process_all_sub_chunks(WhichChunk, Rest, _Candidate, Nonce, State) ->
+	%% The chunk is not a multiple of the subchunk size.
+	?LOG_ERROR([{event, failed_to_split_chunk_into_sub_chunks},
+			{remaining_size, byte_size(Rest)},
+			{nonce, Nonce},
+			{chunk, WhichChunk}]),
+	State.
+
+process_subchunk(chunk1, Candidate, SubChunk, State) ->
```
JamesPiechota, line 632:
```suggestion
process_sub_chunk(chunk1, Candidate, SubChunk, State) ->
```

JamesPiechota (approved):

JamesPiechota (approved):
LGTM!

PR #707, opened 2025-02-07
fix: set the worker timeout long enough for all workers to finish 

JamesPiechota:


apps/arweave/src/ar_entropy_gen.erl
```diff
@@ -228,24 +228,9 @@ do_prepare_entropy(State) ->
...
-                    {bucket_end_offset, BucketEndOffset},
-                    {previous_slice_index, PreviousSliceIndex},
-                    {slice_index, SliceIndex}]),
-            ar_entropy_storage:is_ready(StoreID);
-        _ ->
-            ok
-    end,
+
+    %% Make sure all prior entropy writes are complete.
+    ar_entropy_storage:is_ready(StoreID),
```
ldmberman, line 233: What does this line do?

apps/arweave/src/ar_entropy_storage.erl
```diff
@@ -356,6 +367,19 @@ record_entropy(ChunkEntropy, BucketEndOffset, StoreID, RewardAddr) ->
 	end,
 
 	release_semaphore(Filepath).
+
+%% @doc If we are not at the beginning of the entropy, shift the offset to
+%% the left. store_entropy will traverse the entire 2.9 partition shifting
+%% the offset by sector size. It may happen some sub-chunks will be written
```
ldmberman, line 373: > It may happen some sub-chunks will be written to the neighbouring storage module(s) on the left or on the right since the storage module may be configured to be smaller than the partition.

This is no longer relevant, right?

apps/arweave/src/ar_entropy_gen.erl
JamesPiechota: Right now we support generating the entropy in parallel with writing it to disk (i..e while writing entropy A to disk we can start generating entropy B)

the `is_ready` call bascialy waits on the previous write to finish.

Unfortunately there have been a few race conditions due to this paralllization, so now I always wait before generting entropy which ultimately removes the parallelization.

I'd like to try to reintroduce it later, but for now it's proving problematic (e.g. when we use to just advance the cursor by ?DATA_CHUNK_SIZE the parallelization was fine, but now that we query the next synced interval we *have* to wait for th write to finish otherwise that query may be wrong and cause the same chunk to be written twice - that's the reason I removed the parallelization)

apps/arweave/src/ar_entropy_storage.erl
JamesPiechota: yep! Sorry I have a edit locally that removes that comment, was just waiting on e2e tests to pass

apps/arweave/src/ar_entropy_gen.erl
ldmberman: If we take a noticeable performance hit because of that I would may be roll back "advance_offset" - it should not be needed anyway considering we store intermediate cursor values.

JamesPiechota: Got it. Once I get 2.9.4 I'm planning to look more at performance - as part of that push I'll see if there's a reliable way to parallelize entropy generation and storage. RIght now I'm more worried about reliability, we've been playing a bit of whackamole the past few releases when it comes to race conditions. I think this PR gets to a pretty reliable place (at least the e2e tests pass reliably) - I'm thinking to restart from there

PR #709, opened 2025-02-10
Improve and cleanup Arweave boot procedure

humaite:
`bin/arweave` and `bin/arweave.dev` are now generated using a template from relx (official one). This template has been modified to support the most used Arweave commands and offer an unique entry-point.

Legacy scripts in `bin/` (including `start` and `stop`) are now using the generated template. They are working like the previous scripts.

`bin/arweave.env` has been removed and all its content merged into arweave script template.

The VM parameters - previously set in `arweave.env` script - are now configured using `vm.args` file. It will be easier to understand and to see what arguments the vm is using.

Few scripts have been removed (not used anymore).

see: https://github.com/ArweaveTeam/infra/issues/87


JamesPiechota (approved):

JamesPiechota:
Looks great! can you add in `e2e` to the set of scripts being updated in this PR?

PR #711, opened 2025-02-12
Fix Arweave developer mode

humaite:
When using ARWEAVE_DEV environment variable, a new release must always be recreated. In the current implementation, it was looking for `releases` directory, but was removed by `ar-rebar3` (cleanup).


JamesPiechota (approved):
LGTM

PR #714, opened 2025-02-13
Add tx poller

ldmberman:


apps/arweave/src/ar_tx_poller.erl
```diff
@@ -0,0 +1,160 @@
...
+	{ok, #state{}}.
+
+handle_call(Request, From, State) ->
+	?LOG_WARNING("Unexpected call: ~p from ~p", [Request, From]),
+	{reply, ignored, State}.
+
+handle_cast(check_for_received_txs, State) ->
+	%% Check if there have been any transactions received in the last
+	%% ?CHECK_INTERVAL_MS milliseconds.
+	TimestampDiff = erlang:system_time(microsecond) - State#state.last_seen_tx_timestamp,
```
JamesPiechota, line 63: it looks like we're taking the timestamp in microseconds but comparing it against an interval in milliseconds. should this be `erlang:system_time(millisecond)`?

JamesPiechota (approved):
Other than the microsecond/millisecond question: LGTM!

apps/arweave/src/ar_tx_poller.erl
```diff
@@ -0,0 +1,160 @@
...
+	State3 =
+		case TimestampDiff > 0 andalso TimestampDiff > ?CHECK_INTERVAL_MS of
+			true ->
+				State2 = check_for_received_txs(State),
+				ar_util:cast_after(?POLL_INTERVAL_MS, self(), check_for_received_txs),
```
JamesPiechota, line 68: Actually one more question: It looks like the node will only query at most one tx every ?POLL_INTERVAL_MS, is that correct/by design? I.e. once the CHECK_INVTERVAL_MS has elapsed, the node will query the mempool of the a peer, and the request one tx at a time at most every ?POLL_INTERVAL_MS seconds (rather then just request them sequentially as fast as it can.)

ldmberman: Yes, I should probably make it requests txs without delays

apps/arweave/src/ar_tx_poller.erl
```diff
@@ -0,0 +1,160 @@
...
+check_for_received_txs(#state{ pending_txids = [TXID | PendingTXIDs] } = State) ->
+	case ar_mempool:is_known_tx(TXID) of
+		true ->
+			ok;
+		false ->
+			download_and_verify_tx(TXID)
+	end,
+	State#state{ pending_txids = PendingTXIDs };
+
+check_for_received_txs(#state{ pending_txids = [] } = State) ->
```
JamesPiechota, line 124: When this clause is hit, that restarts the process, right? So the flow is like
1. If no tx has been pushed in CHECK_INTERVAL_MS, start the `check_for_received_txs` loop
2. First get the the mempool of a peer
3. Wait POLL_INTERVAL_MS then query a tx from some per
4. Wait POLL_INVERVAL_MS then query the next tx from some peer
5. repeat 3,4 until the pending list is empty, then return to 2
6. repeat 2-5 until a tx is pushed to the node

apps/arweave/src/ar_tx_poller.erl
ldmberman: They are both in microseconds (system_time(microsecond) and last_seen_tx_timestamp)

JamesPiechota: I mean this bit: `TimestampDiff > ?CHECK_INTERVAL_MS`

CHECK_INTERVAL_MS is in milliseconds I think?

apps/arweave/src/ar_tx_poller.erl
ldmberman: correct

apps/arweave/src/ar_tx_poller.erl
ldmberman: Ah, my bad, good catch

apps/arweave/src/ar_tx_poller.erl
ldmberman: https://github.com/ArweaveTeam/arweave/pull/714/commits/e99067efb355b0ec939791ea7526be57d85f5370

PR #720, opened 2025-02-18
Improve Domain Name Resolution

humaite:
Current arweave behavior is to select one IP address when a record contains more than one value (round robin). The new behavior is to retrieve the complete list and then, give the possibility to create DNS pool. Each IP address is simply added in the peer list (with the default port if not specified).

## Implementation Notes and Review

Few DNS pools have been created to help users and miners to select the closest trusted peers for their region. It has been inspired by the [`pool.ntp.org`](https://www.ntppool.org/zone) convention. Below the list of the current domain name pool available:

 - `peers.arweave.xyz`: contain all available trusted peers
   - `asia.peers.arweave.xyz`:
   - `europe.peers.arweave.xyz`
   - `north-america.peers.arweave.xyz`
   - `india.peers.arweave.xyz`
   - `oceania.peers.arweave.xyz`

The idea is to offer a simple naming convention hiding the name of arweave hosts and their IPs. In case of change, users will always have more than 1 node available from the pool. It will help system administrators to deal with the outages.

Furthermore, it will also be easier to monitor and offers a better view for everybody. Most of our trusted peers are still publicly available and a [status page](https://status.arweave.xyz/) for each of them can be seen. This is a quick view to check if the peer is correctly to HTTP requests.

This new implementation is quite easy at first glance, it only replace `inet:getaddr/2` by `inet:getaddrs/2`. This simple change allows to get the full list of IP addresses set on a domain name instead of one randomly. This function is called when a peer is configured via `ar` module (from command line or configuration file).

These peers (from the DNS pool) are then available in the trusted peers list through `ar_peers:get_trusted_peers/0`. When setting `peer europe.peers.arweave.xyz`, the function returns 6 peers available:

```erlang
ar_peers:get_trusted_peers().
% returns:
% [{91,239,53,95,1984},
%  {57,128,189,200,1984},
%  {67,220,66,3,1984},
%  {57,129,64,100,1984},
%  {23,88,18,239,1984},
%  {23,166,88,225,1984}]
```

To be sure, `dig` command can be used to list the whole list IP addresses available from this record. In this case, dig also returns 6 peers, with the same IP addresses.

```sh
$ dig +short europe.peers.arweave.xyz
23.88.18.239
57.129.64.100
23.166.88.225
91.239.53.95
67.220.66.3
57.128.189.200
```

Let check now what happens when a node goes down. Instead of shutdown the node directly, a new iptables rule will be added:

```sh
# drop all packets coming from 23.88.18.239
sudo iptables -I INPUT -i ${interface} -s 23.88.18.239 -j DROP

# drop all packets going to 23.88.18.239
sudo iptables -I INPUT -i ${interface} -d 23.88.18.239 -j DROP

# check if the peer is still reachable
nc -zvw3  23.88.18.239 1984
# it should return:
# nc: connect to 23.88.18.239 port 1984 (tcp) timed out: Operation now in progress

# kill the remaining connections if present
sudo ss -K dst 23.88.18.239
sudo ss -K src 23.88.18.239
```

When starting the server:

```
Setting the data chunk cache size limit to 1100 chunks.

        Peer 23.88.18.239:1984 is not available.

Joining the Arweave network...
Downloaded the block index successfully.
```

All other nodes are still reachable, only the blocked one is not available.

EDIT: renamed DNS pool records with `peers` instead of `pool` to avoid confusion with mining pools.

### DIFF
--- a/apps/arweave/src/ar_http_iface_client.erl
+++ b/apps/arweave/src/ar_http_iface_client.erl
@@ -1255,7 +1255,8 @@ get_peers(Peer) ->
 					timeout => 2 * 1000
 				}),
 			PeerArray = ar_serialize:dejsonify(Body),
-			lists:map(fun ar_util:parse_peer/1, PeerArray)
+			PeersList = lists:map(fun ar_util:parse_peer/1, PeerArray),
+			lists:flatten(PeersList)
 		end
 	catch _:_ -> unavailable
 	end.
--- a/apps/arweave/src/ar_util.erl
+++ b/apps/arweave/src/ar_util.erl
@@ -101,21 +101,34 @@ invert_map(Map) ->
     ).
 
 
-%% @doc Parse a string representing a remote host into our internal format.
+%%--------------------------------------------------------------------
+%% @doc Parse a string representing a remote host into our internal
+%%      format.
+%% @end
+%%--------------------------------------------------------------------
+-spec parse_peer(Hostname) -> Return when
+	Hostname :: string() | binary(),
+	Return :: [IpWithPort] | no_return(),
+	IpWithPort :: {A, A, A, A, Port},
+	A :: pos_integer(),
+	Port :: pos_integer().
+
 parse_peer("") -> throw(empty_peer_string);
-parse_peer(BitStr) when is_bitstring(BitStr) ->
-	parse_peer(bitstring_to_list(BitStr));
+parse_peer(BitStr) when is_binary(BitStr) ->
+	parse_peer(binary_to_list(BitStr));
 parse_peer(Str) when is_list(Str) ->
-    [Addr, PortStr] = parse_port_split(Str),
-    case inet:getaddr(Addr, inet) of
-		{ok, {A, B, C, D}} ->
-			{A, B, C, D, parse_port(PortStr)};
+	[Addr, PortStr] = parse_port_split(Str),
+	case inet:getaddrs(Addr, inet) of
+		{ok, [{A, B, C, D}]} ->
+			[{A, B, C, D, parse_port(PortStr)}];
+		{ok, AddrsList} when is_list(AddrsList) ->
+			[{A, B, C, D, parse_port(PortStr)} || {A, B, C, D} <- AddrsList];
 		{error, Reason} ->
 			throw({invalid_peer_string, Str, Reason})
 	end;
 parse_peer({IP, Port}) ->
 	{A, B, C, D} = parse_peer(IP),
-	{A, B, C, D, parse_port(Port)}.
+	[{A, B, C, D, parse_port(Port)}].
 
 peer_to_str(Bin) when is_binary(Bin) ->
 	binary_to_list(Bin);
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -194,6 +194,37 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for more information.
 
 You can find documentation regarding our HTTP interface [here](http_iface_docs.md).
 
+# Available Trusted Peers
+
+| type |                               domain name | notes |
+| ---- | ----------------------------------------- | ----- |
+| pool |                        `pool.arweave.xyz` | global pool containing all trusted peers |
```
JamesPiechota, line 201: We should probably find a different name for this since "pool" is already associated with "mining pool".

Maybe cluster? Group? ip_set?

JamesPiechota: Ah and I'd move this documentation into the main docs so that we don't have to worry about keeping it up to date in two locations: https://github.com/ArweaveTeam/docs.arweave.org-info/blob/master/mining/mining-guide.md#trusted-peers

But feel free to create a new section in the docs too - it doesn't have to be buried inside the mining guide, we can link out a separate chapter if needed

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -1255,7 +1255,8 @@ get_peers(Peer) ->
 					timeout => 2 * 1000
 				}),
 			PeerArray = ar_serialize:dejsonify(Body),
-			lists:map(fun ar_util:parse_peer/1, PeerArray)
+			PeersList = lists:map(fun ar_util:parse_peer/1, PeerArray),
```
JamesPiechota, line 1258: Why parse_peer here rather than safe_parse_peer?

JamesPiechota: Oop never mind. I miss read the changes. I though you introduced the safe_parse_peer as a wrapper around parse_peer.

apps/arweave/src/ar_util.erl
```diff
@@ -101,15 +101,28 @@ invert_map(Map) ->
     ).
 
 
-%% @doc Parse a string representing a remote host into our internal format.
+%%--------------------------------------------------------------------
+%% @doc Parse a string representing a remote host into our internal
+%%      format.
+%% @end
+%%--------------------------------------------------------------------
+-spec parse_peer(Hostname) -> Return when
```
JamesPiechota, line 109: it looks like `parse_peer(String)` always returns a list (sometimes a 1 element list), but `parse_peer{IP, PORT})` returns a single element. How hard would it be to have all `parse_peer` matches return a list? A little worried about potential complexity in the future re: return type

JamesPiechota: e.g. here: https://github.com/ArweaveTeam/arweave/pull/720/files#diff-de37132d068e49cb11eb29b5c880b8ef2454d93f79686171b6cff9392824a2c5R349

README.md
JamesPiechota: I'd love to not use "pool" - when miners set up their configuratin they may specify a pool address (referring to a mining pool) and set of trusted peers. If those trusted peers also use the word "pool" it's just going to make everything very confusing. I can already see the endless Discord questions "Can I use the pool.xxx.xyz peers even if I'm not in a mining pool?" - "I don't want to join a mining pool, does anyone have a set of trusted peers that aren't attached to a pool?" etc etc

basically any word other than pool. Please please please.

apps/arweave/src/ar_util.erl
JamesPiechota: oh I thought `parse_peer/1` sometimes returns a tuple and not a list? Here: https://github.com/ArweaveTeam/arweave/pull/720/files#diff-1f26621f4ea9ef47e071afce0f4ce9fc3e6bacc76089ac9037c00e0c07a5ce9eR129

JamesPiechota: i.e.
```
parse_peer({IP, Port}) ->
	{A, B, C, D} = parse_peer(IP),
	{A, B, C, D, parse_port(Port)}.
```

It's this match that I think will cause a problem, since it means `parse_peer/1` sometimes returns a list and sometimes a returns a tuple.

PR #728, opened 2025-02-23
Ignore replica2.9 modules when building data sync records

shizzard:


apps/arweave/src/ar_global_sync_record.erl
```diff
@@ -74,15 +74,14 @@ init([]) ->
...
-					end,
-				R = ar_sync_record:get(ar_data_sync, StoreID),
-				ar_intervals:union(R, Acc)
+				case Module of
+					"default" -> ar_intervals:union(ar_sync_record:get(ar_data_sync, "default"), Acc);
+					%% Ignore replica.2.9 packing. This is a temporary solution until
+					%% we can support data syncing in batches corresponding to the
+					%% replica.2.9 entropy footprint
+					{_, _, {replica_2_9, _}} -> Acc;
+					_ -> ar_intervals:union(ar_sync_record:get(ar_data_sync, ar_storage_module:id(Module)), Acc)
```
JamesPiechota, line 83: It'd be good to remove the get/union duplication. Maybe keep the old StoreID case statement and then do the new replica_2_9 case?

alternatively maybe `ar_storage_module:id` can be updated to handle the `"default"` module

PR #737, opened 2025-02-26
Various quick-n-dirty fixes for HTTP requests client and handlers

shizzard:
NB: WIP, do not merge.


apps/arweave/src/ar_http.erl
```diff
@@ -66,44 +66,46 @@ req(Args) ->
...
-			ar_rate_limiter:throttle(Peer, Path),
-			case request(PID, Args) of
-				{error, Error} when Error == {shutdown, normal}; Error == noproc ->
-					case ReestablishedConnection of
-						true ->
-							{error, client_error};
-						false ->
-							req(Args, true)
+	case catch ar_rate_limiter:throttle(Peer, Path) of
+		{'EXIT', _} -> {error, client_error};
```
JamesPiechota, line 70: When will this clause get triggered? It looks like `throttle` only returns `ok` or ... exits the whole application? (is that right?)

apps/arweave/src/ar_coordination.erl
```diff
@@ -155,15 +155,15 @@ get_self_plus_external_partitions_list() ->
 %%   {pdiff, PackingDifficulty}
 %% ]}
 get_cluster_partitions_list() ->
-	gen_server:call(?MODULE, get_cluster_partitions_list, infinity).
```
JamesPiechota, line 158: If we go with this change (i.e. setting all the timeouts to 600000 instead of infinity), might be worth defining the timeout as a constant - probably in ar.hrl?

apps/arweave/src/ar_rate_limiter.erl
```diff
@@ -40,7 +40,7 @@ throttle2(Peer, Path) ->
 			%% Do not throttle transaction gossip.
 			ok;
 		_ ->
-			case catch gen_server:call(?MODULE, {throttle, Peer, P}, infinity) of
+			case catch gen_server:call(?MODULE, {throttle, Peer, P}, 15000) of
```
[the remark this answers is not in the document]
shizzard: This call may fail for many reasons, just like any `gen_server:call`, for example if the server is not alive (`noproc`). In this case the function will immediately return `ok`.

apps/arweave/src/ar_coordination.erl
shizzard: I thought it might be worth putting into normal configuration?..

PR #738, opened 2025-03-03
fix: fix join error when trusted peer times out

JamesPiechota:
If a trusted peer timed out while querying transactions it hit an unhandled case clause and could prevent the node from joining. This PR addresses the failed case clause by ensuring a list argument is always a list.


shizzard:
LGTM

PR #742, opened 2025-03-05
Improvement/block processing

JamesPiechota:


apps/arweave/src/ar_nonce_limiter.erl
```diff
@@ -742,11 +747,32 @@ session_key(NextSeed, StepNumber, NextVDFDifficulty) ->
 get_session(SessionKey, #state{ session_by_key = SessionByKey }) ->
 	maps:get(SessionKey, SessionByKey, not_found).
 
-update_session(Session, StepNumber, StepCheckpointsMap, Steps) ->
+update_session(Session, StepNumber, StepCheckpointsMap, Steps, Source) ->
 	#vdf_session{ step_checkpoints_map = Map } = Session,
+	case find_step_checkpoints_mismatch(StepCheckpointsMap, Map) of
```
JamesPiechota, line 752: This is intended to just be a LOG change, right? I see the `false` "return" but it's not getting stored anywhere - that's intentional?

ldmberman: Yes, I should have put 'ok' there

PR #745, opened 2025-03-07
Shutdown Procedure for remaining sockets

humaite:
When stopping a node, cowboy is waiting for all connections to be correctly closed. Usually, it waits for the buffer to be empty before closing the socket. In most of the cases, this behavior is acceptable, but in some cases, in particular when a connection is very slow, this can have a direct impact on the delay to shutdown a node.

This commit adds a shutdown procedure when stopping the client connections. It has 3 stages: (1) when a node is stopped, all sockets are set in read-only mode and all clients are noticed of this change, and they will try to fetch the data before closing the connection (2) if some nodes are still present, the procedure sends another closing requests and will wait for a specific delay (3) if the connection is still up, the connection is simply killed, linger option is set to {true, 0} and the socket is definitively closed.

This draining procedure was inspired by ranch documentation example, that can be found at: https://ninenines.eu/docs/en/ranch/2.2/guide/connection_draining/

see: https://github.com/ArweaveTeam/arweave-dev/issues/817

### DIFF
--- a/apps/arweave/include/ar_config.hrl
+++ b/apps/arweave/include/ar_config.hrl
@@ -130,6 +130,10 @@
 %% The number of packing workers.
 -define(DEFAULT_PACKING_WORKERS, erlang:system_info(dirty_cpu_schedulers_online)).
 
+%% The default connection tcp delay when arweave is shutting down
+-define(SHUTDOWN_TCP_CONNECTION_TIMEOUT, 60).
+-define(SHUTDOWN_TCP_MAX_CONNECTION_TIMEOUT, ?SHUTDOWN_TCP_CONNECTION_TIMEOUT*5).
+
 %% @doc Startup options with default values.
 -record(config, {
 	init = false,
@@ -232,7 +236,8 @@
 	%% Undocumented/unsupported options
 	chunk_storage_file_size = ?CHUNK_GROUP_SIZE,
 	rocksdb_flush_interval_s = ?DEFAULT_ROCKSDB_FLUSH_INTERVAL_S,
-	rocksdb_wal_sync_interval_s = ?DEFAULT_ROCKSDB_WAL_SYNC_INTERVAL_S
+	rocksdb_wal_sync_interval_s = ?DEFAULT_ROCKSDB_WAL_SYNC_INTERVAL_S,
+	shutdown_tcp_connection_timeout = ?SHUTDOWN_TCP_CONNECTION_TIMEOUT
 }).
 
 -endif.
--- a/apps/arweave/include/ar_sup.hrl
+++ b/apps/arweave/include/ar_sup.hrl
@@ -1,14 +1,27 @@
 %% The number of milliseconds the supervisor gives every process for shutdown.
 -ifdef(AR_TEST).
--define(SHUTDOWN_TIMEOUT, 30000).
+-define(SHUTDOWN_TIMEOUT, 30_000).
 -else.
--define(SHUTDOWN_TIMEOUT, 30000).
+-define(SHUTDOWN_TIMEOUT, 300_000).
 -endif.
 
--define(CHILD(I, Type), {I, {I, start_link, []}, permanent, ?SHUTDOWN_TIMEOUT, Type, [I]}).
+-define(CHILD(I, Type), #{
+	id => I,
+	start => {I, start_link, []},
+	restart => permanent,
+	shutdown => ?SHUTDOWN_TIMEOUT,
+	type => Type,
+	modules => [I]
+}).
 
--define(CHILD_WITH_ARGS(I, Type, Name, Args),
-		{Name, {I, start_link, Args}, permanent, ?SHUTDOWN_TIMEOUT, Type, [Name]}).
+-define(CHILD_WITH_ARGS(I, Type, Name, Args), #{
+	id => Name,
+	start => {I, start_link, Args},
+	restart => permanent,
+	shutdown => ?SHUTDOWN_TIMEOUT,
+	type => Type,
+	modules => [Name]
+}).
 
 %% From the Erlang docs:
 %%
@@ -18,4 +31,11 @@
 %% milliseconds, the child process is unconditionally terminated using exit(Child,kill).
 %% If the child process is another supervisor, the shutdown time must be set to infinity to
 %% give the subtree ample time to shut down.
--define(CHILD_SUP(I, Type), {I, {I, start_link, []}, permanent, infinity, Type, [I]}).
+-define(CHILD_SUP(I, Type), #{
+	id => I,
+	start => {I, start_link, []},
+	restart => permanent,
+	shutdown => infinity,
+	type => Type,
+	modules => [I]
+}).
--- a/apps/arweave/src/ar.erl
+++ b/apps/arweave/src/ar.erl
@@ -365,7 +365,9 @@ show_help() ->
 				"resync and/or repack any flagged chunks. When running in verify mode several "
 				"flags are disallowed. See the node output for details."},
 			{"verify_samples (num)", io_lib:format("Number of chunks to sample and unpack "
-				"during 'verify'. Default is ~B.", [?SAMPLE_CHUNK_COUNT])}
+				"during 'verify'. Default is ~B.", [?SAMPLE_CHUNK_COUNT])},
+			{"shutdown_tcp_connection_timeout", "shutdown tcp connection timeout in seconds.",
+				"Default is ~Bs.", [?SHUTDOWN_TCP_CONNECTION_TIMEOUT]}
 		]
 	),
 	erlang:halt().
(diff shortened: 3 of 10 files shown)


apps/arweave/src/ar.erl
```diff
@@ -664,6 +665,18 @@ parse_cli_args(["rocksdb_flush_interval", Seconds | Rest], C) ->
 parse_cli_args(["rocksdb_wal_sync_interval", Seconds | Rest], C) ->
 	parse_cli_args(Rest, C#config{ rocksdb_wal_sync_interval_s = list_to_integer(Seconds) });
 
+%% shutdown procedure
+parse_cli_args(["shutdown_tcp_connection_timeout", Delay|Rest], C) ->
```
JamesPiechota, line 669: For now I think probably best to keep to the format of the other parse lines, eg.

```
parse_cli_args(["rocksdb_flush_interval", Seconds | Rest], C) ->
	parse_cli_args(Rest, C#config{ rocksdb_flush_interval_s = list_to_integer(Seconds) });
```

This file is already monstrous enough, but keeping the styling/format the same helps a little.

Plus in general we do a hard/fast fail if any of the flags are set wrong rather than continue processing (e.g. if list_to_integer fails, we should allow the node to exit as we do for the other flags0

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -95,8 +94,181 @@ start_http_iface_listener(Config) ->
...
+%% @end
+%%--------------------------------------------------------------------
+terminate_listener(Pids, Delay) ->
+	AfterDelay = Delay*3,
+	receive
+		{'DOWN', Ref, process, Pid, _} ->
+			Filter = fun({Pid, Ref}) -> false; (_) -> true end,
+			NewPids = lists:filter(Filter, Pids),
+			terminate_listener(NewPids, Delay);
+		_ -> terminate_listener(Pids, Delay)
```
JamesPiechota, line 120: This could potentially loop, but the `after AfterDelay` limits the looping to `Delay*3`ms, right?

JamesPiechota: oh wait, no it doesn't. 

Any risk to infinite loop here?

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -95,8 +94,181 @@ start_http_iface_listener(Config) ->
...
+	Delay :: pos_integer(),
+	Return :: [{pid(), reference()}, ...].
+
+terminate_connections(Delay) ->
+	Processes = ranch:procs(ar_http_iface_listener, connections),
+	[ terminate_connection(P, Delay) || P <- Processes ].
+
+%%--------------------------------------------------------------------
+%% @doc terminate a ranch/cowboy connection. this follows the drain
+%%      procedure explaining in ranch and cowboy documentation.
```
JamesPiechota, line 141:
```suggestion
%%      procedure explained in ranch and cowboy documentation:
%%      https://ninenines.eu/docs/en/ranch/2.2/guide/connection_draining/
```

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -95,8 +94,181 @@ start_http_iface_listener(Config) ->
...
+				peer => Peer,
+				delay => Delay,
+				connection => Connection
+			},
+
+			% now everything is ready, the socket is set in read-only mode. the
+			% remote peer will receive this information and will try to fetch
+			% the last piece of data from the buffer. It does not guarantee the
+			% socket will be closed, in particular if the connection is slow.
+			logger:warning(#{ connection => Peer, socket => Connection, reason => "shutdown", msg => "shutdown (read-only)"}),
```
JamesPiechota, line 199: Any way to use `?LOG_WARNING` for now? to be consistent with the rest of the code

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -95,8 +94,181 @@ start_http_iface_listener(Config) ->
...
+			% to 0 and close the connection (again). At
+			% this stage, the connection should be closed.
+			% and we should receive a message from
+			% monitoring.
+			logger:warning(#{ connection => Peer, socket => Connection, reason => "shutdown", msg => "killed (timeout)"}),
+			ranch_tcp:setopts(Socket, [{linger, {true, 0}}]),
+			ranch_tcp:close(Socket),
+			terminate_connection_loop(State);
+
+		{'DOWN', _Ref, process, Connection, _} ->
```
JamesPiechota, line 264: If this message comes in before one or both of the `{timeout, xxx}` messages, that's okay, right? Nothing else needs to be cleaned up because the connection has already terminated, and the pending timeout messages will just be sent into the void

apps/arweave/include/ar_config.hrl
```diff
@@ -229,7 +229,8 @@
 	%% Undocumented/unsupported options
 	chunk_storage_file_size = ?CHUNK_GROUP_SIZE,
 	rocksdb_flush_interval_s = ?DEFAULT_ROCKSDB_FLUSH_INTERVAL_S,
-	rocksdb_wal_sync_interval_s = ?DEFAULT_ROCKSDB_WAL_SYNC_INTERVAL_S
+	rocksdb_wal_sync_interval_s = ?DEFAULT_ROCKSDB_WAL_SYNC_INTERVAL_S,
+	shutdown_tcp_connection_timeout = 60_000
```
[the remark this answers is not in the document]
shizzard: Maybe we can stick to seconds in our configuration?..

Azmat1177:
unsubscribe my email

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -33,6 +36,21 @@
...
+
+init(_) ->
+	?LOG_INFO(#{ start => ?MODULE, pid => self() }),
+	% this process needs to be stopped in a clean way,
+	% if something goes wrong, the connections must
+	% be cleaned before leaving.
+	erlang:process_flag(trap_exit, true),
+	{ok, Config} = application:get_env(arweave, config),
+	Semaphores = Config#config.semaphores,
+	case start_http_iface_listener(Config) of
```
shizzard, line 50: It is better to move long-lasting operations from `init/1` into `handle_continue/2`. Init callback blocks parent supervisor from starting other children.
Perhaps it is not critical in this case, but if theory it may take some time, and the whole application start will hang until this function returns.

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -33,6 +36,21 @@
...
+	gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+init(_) ->
+	?LOG_INFO(#{ start => ?MODULE, pid => self() }),
+	% this process needs to be stopped in a clean way,
+	% if something goes wrong, the connections must
+	% be cleaned before leaving.
+	erlang:process_flag(trap_exit, true),
+	{ok, Config} = application:get_env(arweave, config),
+	Semaphores = Config#config.semaphores,
```
shizzard, line 49: You forgot to start semaphores here?..

apps/arweave/include/ar_sup.hrl
```diff
@@ -2,13 +2,26 @@
 -ifdef(AR_TEST).
 -define(SHUTDOWN_TIMEOUT, 30000).
 -else.
--define(SHUTDOWN_TIMEOUT, 30000).
+-define(SHUTDOWN_TIMEOUT, 900_000).
```
shizzard, line 5: 15 minutes?.. Looks too much.

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -84,17 +99,231 @@ start_http_iface_listener(Config) ->
...
+%%      or kill the remaining connections.
+%% @end
+%%--------------------------------------------------------------------
+-spec terminate_connections(Delay) -> Return when
+	Delay :: pos_integer(),
+	Return :: [{pid(), reference()}, ...].
+
+terminate_connections(Delay) ->
+	Processes = ranch:procs(ar_http_iface_listener, connections),
+	[ terminate_connection(P, Delay) || P <- Processes ].
```
shizzard, line 168: This whole loop terminates connections sequentially, while we can do that in parallel.

shizzard: Also, the whole logic is not clear to me.
What I would expect to do here:
1. Hibernate all listeners, to stop accepting connections;
2. Wait for at least one HTTP Keep-Alive timeout to allow fast clients to drain connections;
3. Loop over leftover sockets and set linger-zero (setopt);
4. Terminate all connections / terminate the whole ranch/cowboy application.

shizzard: Yes, right.
So, let me just explain what is bothering me in this algorithm.

First, we're shutting down the socket straight away, after hibernating the acceptors, with `socket:shutdown(Sock, write)`. If I recall correctly, this will immediately send the `FIN` towards the peer. My suggestion was to first wait for Keep-Alive timeout to allow for fast clients to drain the connection, not to terminate it in the middle of the transfer.

Second, we're waiting too long for the hanging connections to terminate. When we reach the `terminate_connection_loop`, we already terminated all the well-behaving clients, so there is no need to wait for others. If we follow my suggestion of waiting for a Keep-Alive timeout, we can just brutally terminate all the left sockets (`setopt(Sock, {linger, {true, 0}})`, and then terminate) and exit, which will be much faster.

Third, why cannot everything here be done within the `prep_stop` callback?.. The `terminate` callback is here to be called when the process is about to be terminated, which is not the same as the application/node shutdown. If for some reason this process will be shut down nicely (with shutdown exit signal, for example, with `supervisor:restart_child`), it will execute the whole shutdown sequence, terminating all sockets, and then restart itself, leaving the node in malfunctioning state. Yes, for now there are no cases where this might happen, but it might happen in the future. This is just not the right place to put this code.

Lets discuss these points?

PR #749, opened 2025-03-12
Add metrics counting failed & successful mining solutions

ldmberman:


apps/arweave/src/ar_metrics.erl
```diff
@@ -288,6 +288,22 @@ register() ->
...
+		{labels, [reason]},
+		{help, "The number of times we failed to prepare a block from a mining solution."}
+	]),
+	prometheus_gauge:new([
+		{name, mining_solution_success},
+		{help, "The number of times a block was successfully prepared "
+				"from a mining solution. It does not necessarily mean the block "
+				"ended up in the blockchain."}
+	]),
+	prometheus_gauge:new([
```
JamesPiechota, line 302: I think we can probably leave this metric out? at least in grafana I would typically just sum the other two metrics to get a graph of the total

apps/arweave/src/ar_mining_server.erl
```diff
@@ -703,6 +705,9 @@ prepare_solution(steps, Candidate, Solution) ->
 							Solution#mining_solution{ steps = Steps })
 			end;
 		false ->
+			ar_events:send(solution, {stale,
```
JamesPiechota, line 708: What's the significance of stale vs. rejected? Or maybe it's historical? (I see the rejected and stale event types already existed)

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1369,6 +1367,14 @@ apply_validated_block2(State, B, PrevBlocks, Orphans, RecentBI, BlockTXPairs) ->
 	ForkRootB = lists:last(PrevBlocks), %% The root of any detected fork
 	prometheus_gauge:set(block_time, B#block.timestamp - PrevB#block.timestamp),
 	record_economic_metrics(B, PrevB),
+	lists:foldl(
+		fun(OrphanH, OrphanHeight) ->
+			ar_watchdog:block_orphaned(OrphanH, OrphanHeight),
```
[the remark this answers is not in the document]
JamesPiechota: Ah never mind! I see that in ar_watchdog you filter based on blocks that this miner mined!

apps/arweave/src/ar_metrics.erl
JamesPiechota: I can make this edit.

JamesPiechota: https://github.com/ArweaveTeam/arweave/pull/749/commits/317f549b3ae81b9914e687de349a7d9b2973336c

apps/arweave/src/ar_mining_server.erl
ldmberman: We introduced this separation for the pool (ar_pool is relying on these events)

PR #751, opened 2025-03-12
metrics fix

vird:
metrics add:
* kryder_plus_rate_multiplier
* endowment give, take

metrics remove:
* network burden (set of metrics, use expected_block_reward instead)


JamesPiechota:
@vird based on our call today, I think this is ready to merge whenever you're ready?

PR #752, opened 2025-03-13
Rotate VDF servers every 30 seconds (in addition to rotating on error)

ldmberman:


apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -53,125 +60,114 @@ init([]) ->
...
 			end
 	end;
 
 handle_cast({maybe_request_sessions, SessionKey}, State) ->
 	#state{ remote_servers = Q } = State,
-	{{value, {RawPeer, _Timestamp}}, Q2} = queue:out(Q),
-	Now = erlang:system_time(millisecond),
-	RotatedServers = queue:in({RawPeer, Now}, Q2),
+	{{value, RawPeer}, Q2} = queue:out(Q),
+	RotatedServers = queue:in(RawPeer, Q2),
```
JamesPiechota, line 170:
```suggestion
	RotatedServers = rotate_servers(Q),
```

apps/arweave/src/ar_nonce_limiter_client.erl
```diff
@@ -53,125 +60,114 @@ init([]) ->
...
-							State2 = update_latest_session_key(Peer, SessionKey, State),
-							UpdateResponse =
-								ar_nonce_limiter:apply_external_update(Update, Peer),
-
-							SessionFound = case UpdateResponse of
-								#nonce_limiter_update_response{ session_found = false } ->
-									false;
+	{RawPeer, State2} = get_raw_peer_and_update_remote_servers(State),
+	#state{ remote_servers = Q } = State2,
+	RotatedServers = rotate_servers(Q),
```
JamesPiechota, line 72: This might double rotate the peers sometimes, right? e.g. if the timestamp has elapsed `get_raw_peer_and_update_remote_servers` will rotate, and then if there's an error querying RawPeer, it will rotate again. 

If they only have 2 remote peers, is it possible they could keep double rotate from the "bad" peer back to the "bad" peer?

what if `get_raw_peer_and_update_remote_servers` returns the `RotateServers` as well - i.e. if the timestap has elapsed it returns the new queue, and if the timestamp hasn't elapsed it return `rotate_servers(Q)`

ldmberman: get_raw_peer_and_update_remote_servers rotates conditionally based on the elapsed time whereas on error we always rotate. That's why I use State2 returned by get_raw_peer_and_update_remote_servers in all success cases and RotatedServers - in the error cases. There is indeed a double-rotation bug here, however, because we need to rotate State#state.remote_servers, not State2#state.remote_servers.

apps/arweave/src/ar_nonce_limiter_client.erl
ldmberman: https://github.com/ArweaveTeam/arweave/pull/752/commits/6e297d9821ed81e2d147441ab0602ba82b98a762

apps/arweave/src/ar_nonce_limiter_client.erl
ldmberman: https://github.com/ArweaveTeam/arweave/pull/752/commits/6e297d9821ed81e2d147441ab0602ba82b98a762

JamesPiechota: Looks good!

JamesPiechota (approved):
LGTM!

PR #754, opened 2025-03-14
Fix user script to follow logs

humaite:
With the new entry-point and the modification of the format used for the logs, this script was not able to follow correctly the log files in ./logs/ directory.

see: https://github.com/ArweaveTeam/arweave-dev/issues/833

### DIFF
--- a/bin/logs
+++ b/bin/logs
@@ -1,12 +1,6 @@
 #!/usr/bin/env bash
-
 set -e
 
 SCRIPT_DIR="$(dirname "$0")"
 LOGS_DIR="$(cd $SCRIPT_DIR/../logs && pwd -P)"
-
-ls -1 -t $LOGS_DIR \
-    | grep -F -v '(?!.*(idx|siz))' \
-    | grep -v slave \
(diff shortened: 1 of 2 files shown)


bin/logs
```diff
@@ -4,9 +4,4 @@ set -e
 
 SCRIPT_DIR="$(dirname "$0")"
 LOGS_DIR="$(cd $SCRIPT_DIR/../logs && pwd -P)"
-
-ls -1 -t $LOGS_DIR \
-    | grep -F -v '(?!.*(idx|siz))' \
-    | grep -v slave \
-    | head -1 \
-    | xargs -I % tail -n 500 "$@" "$LOGS_DIR/%"
+tail -f ${LOGS_DIR}/*.log
```
JamesPiechota, line 7: I think this one should only follow the `info` logs? and `./bin/debug-logs` follows the `debug` logs?

I haven't used the tools myself, but I thin that was their original intent

JamesPiechota (approved):
LGTM!

PR #756, opened 2025-03-18
openssl to cryptopp

vird:
benchmark-vdf shows same time as before
Pros:
* We no longer rely on outdated openssl (works on Mac)
* crypto++ has multiple FIPS validations (https://en.wikipedia.org/wiki/Crypto%2B%2B)

Cons:
* +4 minutes of compilation (can easily speed up with `make -j$(nproc)`)


vird:
After last fixes:
Compile time cryptopp reduced 4 min -> 2 min
BUT. I've found and fixed that VDF code still used openssl's SHA implementation
After that fix
OLD 1.21 s
NEW 1.46 s, so new is definitely a degradation
And we still need openssl (but we can use 3.0) in ecdsa cryptography (for generation random private key)

PR #758, opened 2025-03-20
Allow Users to set Erlang VM parameters

humaite:
This commit fixes a regression with the new entry-point. Users in some case needed to modify VM parameters
directly from the command line (and not from vm.args).

see: https://github.com/ArweaveTeam/arweave-dev/issues/842

### DIFF
--- a/bin/arweave
+++ b/bin/arweave
@@ -1000,6 +1000,39 @@ process_internal_args() {
     done
 }
 
+# This function takes a list of terms (usually arguments)
+# and split them in two categories, the one before --
+# and the one after. The one before is used as Erlang
+# VM parameters and should overwrite default configuration,
+# The last part (LOCAL_PARAMS) contains arweave parameters.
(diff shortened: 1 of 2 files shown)


bin/arweave
```diff
@@ -1000,6 +1000,39 @@ process_internal_args() {
...
+# The last part (LOCAL_PARAMS) contains arweave parameters.
+# This function export LOCAL_PARAMS and VM_PARAMS variables.
+parse_args() {
+	local separator="--"
+	local vm_params=""
+	local params=""
+	while test "${*}"
+	do
+		local arg="${1}"
+		if test "${arg}" = ${separator}
```
JamesPiechota, line 1016: I suck at reading bash scripts ... and so can't really understand if this works. (i.e. naively I don't see how we differentiate the args from before the separator from those after the separation). So I'll just approve the PR and trust that we hit any issues intesting!

JamesPiechota (approved):

PR #759, opened 2025-03-20
Update Arweave Documentation

humaite:
This commit adds a FAQ section to explain how to use the new arweave entry-point.

see: https://github.com/ArweaveTeam/arweave-dev/issues/843


JamesPiechota:
I think we should move this to https://github.com/ArweaveTeam/docs.arweave.org-info

PR #760, opened 2025-03-20
Speed up collection of syncing intervals

ldmberman:
The core motivation was to reduce the memory footprint of AllPeerIntervals. Now, instead of remembering peer intervals we make sure to only ask every peer for the intervals of every weave range once. To simplify the task, we introduce GET
/data_sync_record/[start]/[end]/[count_limit] on top of GET /data_sync_record/[start]/[count_limit]. We expect redundant data_sync_record calls to peers until the network adopts the new release, however the patch also reduces the size of each individual request (querying at most 100 intervals for each GB of range.)

### DIFF
--- a/apps/arweave/src/ar_data_discovery.erl
+++ b/apps/arweave/src/ar_data_discovery.erl
@@ -6,8 +6,8 @@
 
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2]).
 
--include_lib("arweave/include/ar.hrl").
--include_lib("arweave/include/ar_data_discovery.hrl").
+-include("../include/ar.hrl").
+-include("../include/ar_data_discovery.hrl").
 
 -record(state, {
 	peer_queue,
--- a/apps/arweave/src/ar_data_discovery.erl
+++ b/apps/arweave/src/ar_data_discovery.erl
@@ -6,8 +6,8 @@
 
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2]).
 
--include("../include/ar.hrl").
--include("../include/ar_data_discovery.hrl").
+-include("ar.hrl").
+-include("ar_data_discovery.hrl").
 
 -record(state, {
 	peer_queue,
--- a/apps/arweave/src/ar_global_sync_record.erl
+++ b/apps/arweave/src/ar_global_sync_record.erl
@@ -2,9 +2,9 @@
 
 -behaviour(gen_server).
 
--include("../include/ar.hrl").
--include("../include/ar_config.hrl").
--include("../include/ar_data_discovery.hrl").
+-include("ar.hrl").
+-include("ar_config.hrl").
+-include("ar_data_discovery.hrl").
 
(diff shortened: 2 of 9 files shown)


apps/arweave/src/ar_data_discovery.erl
```diff
@@ -6,8 +6,8 @@
 
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2]).
 
--include_lib("arweave/include/ar.hrl").
--include_lib("arweave/include/ar_data_discovery.hrl").
+-include("../include/ar.hrl").
```
shizzard, line 9:
```suggestion
-include("ar.hrl").
```

apps/arweave/src/ar_data_discovery.erl
```diff
@@ -6,8 +6,8 @@
...
+-include("../include/ar_data_discovery.hrl").
```
shizzard, line 10:
```suggestion
-include("ar_data_discovery.hrl").
```

apps/arweave/src/ar_global_sync_record.erl
```diff
@@ -2,9 +2,9 @@
 
 -behaviour(gen_server).
 
--include_lib("arweave/include/ar.hrl").
--include_lib("arweave/include/ar_config.hrl").
--include_lib("arweave/include/ar_data_discovery.hrl").
+-include("../include/ar.hrl").
+-include("../include/ar_config.hrl").
+-include("../include/ar_data_discovery.hrl").
```
shizzard, line 7: Same as above, `../include` bit is not needed here.

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -6,35 +6,35 @@
...
--include_lib("arweave/include/ar_wallets.hrl").
--include_lib("arweave/include/ar_pool.hrl").
+-include("../include/ar.hrl").
+-include("../include/ar_config.hrl").
+-include("../include/ar_consensus.hrl").
+-include("../include/ar_data_sync.hrl").
+-include("../include/ar_data_discovery.hrl").
+-include("../include/ar_mining.hrl").
+-include("../include/ar_wallets.hrl").
+-include("../include/ar_pool.hrl").
```
shizzard, line 37: Same as above.

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -370,7 +370,7 @@ get_sync_record(Peer) ->
 		peer => Peer,
 		method => get,
 		path => "/data_sync_record",
-		timeout => 180 * 1000,
+		timeout => 30 * 1000,
```
shizzard, line 373: Since we are working with timeouts here, maybe we can move it into configuration as well?.. Such things might be tuned by technically advanced miners to improve performance.

ldmberman: I created an [issue](https://github.com/ArweaveTeam/arweave-dev/issues/845), I suggest we take care of it separately

JamesPiechota (approved):

PR #761, opened 2025-03-20
Fix/chunk range extra check

ldmberman:


apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -871,6 +873,29 @@ is_offset_valid(Byte, BucketStart, ChunkOffset) ->
...
+	Chunks;
+filter_by_sync_record([], _Byte, _Start, _ChunkFileStart, _StoreID, _ChunkCount) ->
+	[];
+filter_by_sync_record([{PaddedEndOffset, Chunk} | Rest], Byte, Start, ChunkFileStart, StoreID, ChunkCount) ->
+	case ar_sync_record:is_recorded(PaddedEndOffset, ar_chunk_storage, StoreID) of
+		false ->
+			%% Filter out repacking artifacts when reading a range of chunks.
+			%% This warning may also occur due to unlucky timing when the chunk is
+			%% being removed.
+			?LOG_WARNING([{event, found_chunk_not_in_sync_record},
```
JamesPiechota, line 886: what do you think about invalidating the chunk when this occurs? so it can be resynced?

apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -871,6 +873,29 @@ is_offset_valid(Byte, BucketStart, ChunkOffset) ->
 	Delta = Byte - (BucketStart + ChunkOffset rem ?DATA_CHUNK_SIZE),
 	Delta >= 0 andalso Delta < ?DATA_CHUNK_SIZE.
 
+filter_by_sync_record(Chunks, _Byte, _Start, _ChunkFileStart, _StoreID, 1) ->
+	Chunks;
```
JamesPiechota, line 877: Is this to handle the special case where we query a single chunk (e.g. from a code path other than the mining code path)? In that case we *won't* filter?

apps/arweave/src/ar_chunk_storage.erl
ldmberman: I do not think resyncing is blocked in this case - we are not looking into the ar_data_sync sync_record here. Another possibility may be the chunk is synced and stored outside of ar_chunk_storage.

apps/arweave/src/ar_chunk_storage.erl
ldmberman: In this case we have already implicitly checked the chunk belongs to the sync record - https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_chunk_storage.erl#L168

JamesPiechota, line 877:
```suggestion
	%% The code paths which query a single chunk have already implicitly checked that
	%% the chunk belongs to the sync_record. E.g. ar_chunk_storage:get/2
	Chunks;
```

Does this work for a comment to explain?

What about ar_chunk_storage:get/3? It looks like we call it directly from `ar_entropy_storage` now: 
- https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_entropy_storage.erl#L225
- https://github.com/ArweaveTeam/arweave/blob/master/apps/arweave/src/ar_entropy_storage.erl#L326

 I think both of those locations already check that it exists in the sync_record, but just want to double check with you.

PR #765, opened 2025-04-02
refactor: expand use of #chunk_metadata and #chunk_proof

JamesPiechota:
This is an incremental step towards replacing some of our long, opaque tuple arguments with records that can be used across modules

### DIFF
--- a/apps/arweave/include/ar.hrl
+++ b/apps/arweave/include/ar.hrl
@@ -759,6 +759,22 @@
 	signature_type = ?DEFAULT_KEY_TYPE
 }).
 
+-record(chunk_metadata, {
+	chunk_data_key = not_set :: not_set | not_found | binary(),
+	tx_root = not_set :: not_set | not_found | binary(),
+	tx_path = not_set :: not_set | not_found | binary(),
+	data_root = not_set :: not_set | not_found | binary(),
+	data_path = not_set :: not_set | not_found | binary(),
+	chunk_size = not_set :: not_set | not_found | non_neg_integer()
+}).
+
+-record(chunk_offsets, {
+	absolute_offset = not_set :: not_set | non_neg_integer(),
+	bucket_end_offset = not_set :: not_set | non_neg_integer(),
+	padded_end_offset = not_set :: not_set | non_neg_integer(),
+	relative_offset = not_set :: not_set | non_neg_integer()
+}).
+
 %% A macro to convert AR into Winstons.
 -define(AR(AR), (?WINSTON_PER_AR * AR)).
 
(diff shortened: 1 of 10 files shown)


apps/arweave/include/ar.hrl
```diff
@@ -759,6 +759,22 @@
 	signature_type = ?DEFAULT_KEY_TYPE
 }).
 
+-record(chunk_metadata, {
+	chunk_data_key = not_set :: not_set | not_found | binary(),
```
ldmberman, line 763: not_found?

apps/arweave/src/ar_poa.erl
```diff
@@ -79,63 +85,66 @@ validate(Args) ->
 			end
 	end.
 
-%% @doc Validate the TXPath and DataPath for a chunk. This will return the ChunkID but won't
-%% validate that the ChunkID is correct.
-%% 
-%% SPoA: the proof of access
-%% RecallOffset: the absoluteoffset of the recall byte - 
- validate_paths(#poa{} = SPoA, TXRoot, RecallOffset, BlockStartOffset, BlockSize) ->
+chunk_proof(#chunk_metadata{} = ChunkMetadata, AbsoluteOffset) ->
```
ldmberman, line 88: Should we rename AbsoluteOffset to e.g. SeekByte because AbsoluteOffset is the end offset in other places?

apps/arweave/include/ar.hrl
JamesPiechota: Yeah that's a holdover from ar_repack. The data_path at least can be not found even when the chunks index metadata exists. 

So I need to distinguish not set vs not found for datapath at least. But I think not necessary for the other fields. 

I can remove the not found for those - but ok to leave for data path?

ldmberman: Is it a case of a broken record when we miss an entry in chunk_data_db? What do we do with this chunk_metadata with data_path=not_found then?

JamesPiechota: Yeah. If the chunks index data exists but the datapath does not then we invalidate the data record.

JamesPiechota: I mostly needed a distinction between "not yet queried" and "queried but not found"

apps/arweave/src/ar_poa.erl
JamesPiechota: Like this?
https://github.com/ArweaveTeam/arweave/pull/765/commits/fa674567a90713969d8eedf163d9edb9b500ef2a

apps/arweave/include/ar.hrl
JamesPiechota: Is this better?
https://github.com/ArweaveTeam/arweave/pull/765/commits/14f6c15f3ec5b9db1e2a1408f697978f33eac97c

ldmberman (approved):

PR #766, opened 2025-04-02
Improvement/use local account state on join

ldmberman:


apps/arweave/src/ar_wallets.erl
```diff
@@ -170,28 +170,38 @@ handle_call({add_wallets, RootHash, Wallets, Height, Denomination}, _From, DAG)
...
-			erlang:halt();
+			B =
+				case length(Blocks) >= ?STORE_BLOCKS_BEHIND_CURRENT of
+					true ->
+						lists:nth(?STORE_BLOCKS_BEHIND_CURRENT, Blocks);
+					false ->
+						lists:last(Blocks)
+				end,
+			case maps:get(trusted_peers, Args, []) of
+				[] ->
```
JamesPiechota, line 188: With this new logic we'll:
1. always check local state first
2. only if it's not found locally *and* trusted peers are set we'll try to download from them

is that right?

JamesPiechota: This change alone might have a pretty huge impact on the network by cutting down the traffic to the trusted peers + speeding up node launch time

apps/arweave/src/ar_storage.erl
```diff
@@ -882,17 +883,24 @@ read_wallet_list(WalletListHash) when is_binary(WalletListHash) ->
 read_wallet_list({ok, << K:48/binary, _/binary >>, Bin}, Tree, Keys, RootHash, K) ->
 	case binary_to_term(Bin) of
 		{Key, Value} ->
-			Tree2 = ar_patricia_tree:insert(Key, Value, Tree),
+			Tree2 = ar_patricia_tree:insert(Key, Value, Tree, K),
```
JamesPiechota, line 886: What is `K` here? It looks like we're now caching `K` in the patricia tree whereas before we weren't?

ldmberman: I renamed it to NodeHash - the hash of a node in the tree. We have them on disk but we nevertheless recomputed them during node initialization instead of fetching and caching the already computed values. Note this only applies to the initial tree buildup - the hashes are cached and only the affected parts of the tree are rehashed after every update.

PR #780, opened 2025-04-10
Lb/improve get chunk range sync record check

ldmberman:


apps/arweave/src/ar_chunk_storage.erl
```diff
@@ -1036,12 +1068,69 @@ get_packing_label(Packing, State) ->
...
+				{position, Position},
+				{padded_end_offset, PaddedEndOffset}])
+	end.
+
+fix_broken_chunk_storage_record3(ChunkFileStart, Position, State) ->
+	StoreID = State#state.store_id,
+	Filepath = filepath(ChunkFileStart, StoreID),
+	case file:open(Filepath, [read, write, raw]) of
+		{ok, F} ->
+			case file:pwrite(F, Position, << 0:(?OFFSET_BIT_SIZE) >>) of
```
JamesPiechota, line 1106: This will invalidate the chunk on disk, but leave the sync records intact, right? So a miner will have to run verify or something similar in order to "heal" the chunk on disk?

ldmberman: This fixes the broken state when the chunk is found in the chunk file, but not in the ar_chunk_storage sync record so there is nothing left to heal. Hm, actually, it may be worth overwriting the entire chunk space with 0s here because the content may be potentially blocklisted.

JamesPiechota: > This fixes the broken state when the chunk is found in the chunk file, but not in the ar_chunk_storage sync record

Ah, got it.

JamesPiechota (approved):

PR #781, opened 2025-04-10
Fix race condition where we may not detect double-signing

ldmberman:
Also, add some redundant correctness checks when including a double-signing proof into a block.

### DIFF
--- a/apps/arweave/src/ar_node_worker.erl
+++ b/apps/arweave/src/ar_node_worker.erl
@@ -490,18 +490,6 @@ handle_info({tx_ready_for_mining, TX}, State) ->
 	ar_events:send(tx, {ready_for_mining, TX}),
 	{noreply, State};
 
-handle_info({event, block, {double_signing, Proof}}, State) ->
-	Map = maps:get(double_signing_proofs, State, #{}),
-	Key = element(1, Proof),
-	Addr = ar_wallet:hash_pub_key(Key),
-	case is_map_key(Addr, Map) of
-		true ->
-			{noreply, State};
-		false ->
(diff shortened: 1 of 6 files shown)


apps/arweave/src/ar_node_worker.erl
```diff
@@ -1023,33 +1012,61 @@ may_be_get_double_signing_proof2(Iterator, RootHash, LockedRewards, Height) ->
...
 							andalso byte_size(Sig1) == ?RSA_BLOCK_SIG_SIZE
 							andalso byte_size(Sig2) == ?RSA_BLOCK_SIG_SIZE
 				end,
-			HasLockedReward =
+			CheckDifferentSignatures =
 				case CheckKeyType of
+					false ->
+						false;
+					true ->
+						Sig1 /= Sig2
```
JamesPiechota, line 1039: This is just a sanity check right? It shouldn't ever happen? i.e. we'd expect the signature to always be different unless the same block is selected twice?

apps/arweave/test/ar_post_block_tests.erl
```diff
@@ -573,7 +573,6 @@ test_reject_block_invalid_double_signing_proof(KeyType) ->
 	B7_2 = sign_block(B7#block{ tags = [<<"new_tag">>] }, B6, Key),
 	post_block(B6, valid),
 	post_block(B7, valid),
-	timer:sleep(1000),
```
JamesPiechota, line 576: 🙌

JamesPiechota (approved):

apps/arweave/src/ar_node_worker.erl
ldmberman: yes

PR #783, opened 2025-04-11
Some repack-in-place fixes

JamesPiechota:


apps/arweave/src/ar.erl
```diff
@@ -411,6 +411,8 @@ parse_cli_args(["verify", _ | _], C) ->
 	io:format("Invalid verify mode. Valid modes are 'purge' or 'log'.~n"),
 	timer:sleep(1000),
 	erlang:halt();
+parse_cli_args(["verify_samples", <<"all">> | Rest], C) ->
```
ldmberman, line 414: In ar.erl (unlike ar_config.erl) we have strings instead of binaries here.

ldmberman (approved):

apps/arweave/src/ar.erl
JamesPiechota: thanks for the fix!

PR #788, opened 2025-04-15
Lb/flexible solution base

ldmberman:


apps/arweave/src/ar_node_worker.erl
```diff
@@ -1885,6 +1888,89 @@ dump_mempool(TXs, MempoolSize) ->
...
+	FrontBlocks = ar_block_cache:get_validated_front(block_cache),
+	PrevBlocks = [ar_block_cache:get(block_cache, B#block.previous_block)
+			|| B <- FrontBlocks],
+	pick_prev_block_for_solution2(Solution, Source, FrontBlocks, PrevBlocks).
+
+pick_prev_block_for_solution2(Solution, Source, FrontBlocks, PrevBlocks) ->
+	#mining_solution{ mining_address = MiningAddress } = Solution,
+	case [B || B <- FrontBlocks, B#block.reward_addr == MiningAddress] of
+		[PrevB | _] ->
+			pick_prev_block_for_solution(Solution, Source, PrevB);
```
JamesPiechota, line 1901: This clause is to ensure we always build on one of our previously mined blocks if possible, right?

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1885,6 +1888,89 @@ dump_mempool(TXs, MempoolSize) ->
...
+pick_prev_block_for_solution3(Solution, Source, [PrevB | FrontBlocks], PrevBlocks) ->
+	#mining_solution{ step_number = StepNumber } = Solution,
+	MaybeSolutionOffPrevB = may_be_place_solution_on_block(PrevB, StepNumber, Solution),
+	case MaybeSolutionOffPrevB of
+		{false, _Reason} ->
+			pick_prev_block_for_solution3(Solution, Source, FrontBlocks, PrevBlocks);
+		Solution2 ->
+			{PrevB, Solution2}
+	end;
+pick_prev_block_for_solution3(Solution, Source, [], PrevBlocks) ->
```
JamesPiechota, line 1915: I don't quite follow why we need both `pick_prev_block_for_solution2` and `pick_prev_block_for_solution3`?

apps/arweave/src/ar_node_worker.erl
```diff
@@ -2055,12 +2140,29 @@ handle_found_solution(Args, PrevB, State, IsRebase) ->
 						{false, rebase_threshold}
 				end
 		end,
-	%% Check steps and step checkpoints.
-	HaveSteps =
+
+	PrevCDiff = PrevB#block.cumulative_diff,
+	CDiff = ar_difficulty:next_cumulative_diff(PrevCDiff, Diff, Height),
+	NoDoubleSigning =
```
JamesPiechota, line 2146: We're adding this check here so that a miner can detect accidental double signing before they publish?

apps/arweave/src/ar_node_worker.erl
ldmberman: Yes, we have to build on our block at this height, if any, otherwise we will double-sign.

apps/arweave/src/ar_node_worker.erl
ldmberman: pick_prev_block_for_solution2 looks for our blocks among the FrontBlocks

pick_prev_block_for_solution3 runs when there are none of our blocks in FrontBlocks, it goes block by block checking if we can put the new block on top; if there are no suitable blocks, it looks for our block among their (FrontBlock's) previous blocks

pick_prev_block_for_solution4 runs when there are none of our blocks among FrontBlock's previous blocks, it iterates over previous blocks

I will try to make it more readable

apps/arweave/src/ar_node_worker.erl
ldmberman: Yes

apps/arweave/src/ar_node_worker.erl
ldmberman: Rearranged and renamed the functions

JamesPiechota (approved):

PR #796, opened 2025-05-01
Support repack_in_place from replica_2_9

JamesPiechota:


apps/arweave/src/ar_repack_fsm.erl
```diff
@@ -109,24 +113,84 @@ next_state(#repack_chunk{state = needs_chunk} = RepackChunk) ->
...
+	} = RepackChunk,
+
+	NextState = case {SourcePacking, TargetPacking} of
+		_ when SourcePacking == TargetPacking ->
+			write_chunk;
+		{{replica_2_9, _}, _} ->
+			%% Source is replica_2_9, so we need its entropy first before we can unpacked it.
+			needs_source_entropy;
+		{unpacked_padded, {replica_2_9, _}} ->
+			needs_target_entropy;
```
ldmberman, line 188: Why `needs_target_entropy` only when SourcePacking is unpacked_padded (and not unpacked or spora_2_6)?

ldmberman: Ah, every source packing goes through unpacked_padded when the target packing is replica_2_9. The comments are very helpful, thank you.

ldmberman (approved):

PR #806, opened 2025-05-19
Improving Arweave Shutdown Procedure

humaite:
This commit try to find a way to close HTTP/1.1
connections with high latency peer. Indeed, HTTP
protocol is half-duplex (at least for the version
used right now) and some peers can request huge
amount of data with a poor connection.

In this situation, the server is stuck with TCP
waiting for an ACK, even when sending a FIN for
the shutdown procedure. A similar behavior seems
to be applied using close and linger. Still unclear
if it's a bug on Erlang side or some behavior
from arweave client/server configuration.

The new shutdown procedure is more brutal, in both
soft and hard method. The list of sockets are
directly listed from erlang:ports/0 functions and
filtered to extract only the sockets to close. Then,
the code loops until all sockets are closed. This
procedure is now executed as prep_stop callback
in ar application.

A new module/process called ar_shutdown_manager has
been created and added on the arweave application tree.
This process should be started first and stopped last,
it is in charge of cleaning up the remaining connections
from gun and cowboy when the application is going down.
Every shutdown procedures and related functions have been
migrated to this module.

The default inet backend is set to inet, but based on the
documentation, it could be better to use socket backend
instead in particular during shutdown time due to
non-blocking socket behaviors.

Another module called ar_timer has been created to
manage timers created by other modules and processes.
Those intervals are mainly used to synchronize data
on the network, but they are not managed. When arweave
is stopped, those intervals are still trying to do
their jobs by synchronizing data, and then, creates
new http connections to random peers.

Updated configuration to permit anyone to configure
http client and server sockets instead of recompiling
arweave everytime. Here the new options added for
the http client (gun):

  - http_client.http.closing_timeout:
    default: 15_000 milliseconds
    values: positive integer
    see: https://ninenines.eu/docs/en/gun/2.2/manual/gun/#_http_opts

  - http_client.http.keepalive:
    default: 60_000 milliseconds
    values: positive integer
    see: https://ninenines.eu/docs/en/gun/2.2/manual/gun/#_http_opts

  - http_client.tcp.delay_send:
    default: false
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_client.tcp.keepalive:
    default: true
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_client.tcp.linger:
    default: false
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_client.tcp.linger_timeout:
    default: 0
    values: positive integer
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_client.tcp.nodelay:
    default: true
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_client.tcp.send_timeout_close:
    default: true
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_client.tcp.send_timeout:
    default 15_000 milliseconds
    values: positive integer
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

Here the list of the new options for the http
server (cowboy):

  - http_api.http.active_n:
    default: 100
    values: positive integer
    see: https://ninenines.eu/docs/en/cowboy/2.10/manual/cowboy_http/#_options

  - http_api.idle_timeout:
    default: 60_000 milliseconds
    values: positive integer
    see: https://ninenines.eu/docs/en/cowboy/2.10/manual/cowboy_http/#_options

  - http_api.http.inactivity_timeout:
    default: 300_000 milliseconds
    values: positive integer
    see: https://ninenines.eu/docs/en/cowboy/2.10/manual/cowboy_http/#_options

  - http_api.http.linger_timeout:
    default: 1000 milliseconds
    values: positive integer
    see: https://ninenines.eu/docs/en/cowboy/2.10/manual/cowboy_http/#_options

  - http_api.http.request_timeout:
    default: 5000 milliseconds
    values: positive integer
    see: https://ninenines.eu/docs/en/cowboy/2.10/manual/cowboy_http/#_options

  - http_api.tcp.delay_send:
    default: false
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_api.tcp.keepalive:
    default: true
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_api.tcp.linger
    default: false
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_api.tcp.linger_timeout:
    default: 0
    values: positive integer
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_api.tcp.listener_shutdown:
    default: 5000
    values: 'brutal_kill' or 'infinity' or positive integer
    see: https://ninenines.eu/docs/en/cowboy/2.10/manual/cowboy_http/#_options

  - http_api.tcp.nodelay
    default: true
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_api.tcp.num_acceptors:
    default: 10
    values: positive integer
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_api.tcp.send_timeout_close
    default: true
    values: boolean
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

  - http_api.tcp.send_timeout
    default: 15_000 milliseconds
    values: positive integer
    see: https://www.erlang.org/docs/24/man/inet#setopts-2

Some parameters have not been added because of potential
incompatibilities with different version of Erlang and/or
issues with the socket backend. An example with correct
values can be found in test fixtures.

inet_backend socket seems to generate error during
tests. Instead of setting it directly in arweave,
it is now possible to configure it from arweave
configuration. It is still possible to configure
it from config/sys.config.

  - socket.backend:
    default: inet
    values: inet or socket
    see: https://www.erlang.org/doc/apps/kernel/inet.html#t:inet_backend/0

socket.backend configure the Erlang inet_backend
kernel parameter before starting arweave. This
value can also be fixed in config/sys.config, but
to avoid odd behaviors, the default backend
(called inet) should be used instead. The socket
backend is still in testing, but users can
switch to it if needed.

max_connections parameters has been updated
only in record to reflect other parameters.

Many socket parameters have been modified to improve
the closing time of the socket. Users can also tune
their system if needed, in particular these values on
Linux, by using sysctl:

  - net.core.default_qdisc:
    see: https://sysctl-explorer.net/net/core/default_qdisc/

  - net.ipv4.tcp_congestion_control:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_dsack:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_fack:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_fin_timeout:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_keepalive_intvl:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_keepalive_probes:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_keepalive_time:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_min_rtt_wlen:
    see https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_orphan_retries:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_retries1:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_retries2:
    see: https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_sack:
    see https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_synack_retries:
    see https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_syn_retries:
    see https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

  - net.ipv4.tcp_tw_reuse:
    see https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html

More documentation is required on this part, to
explain how to configure correctly these values
depending of user's needs.

In some specific cases, the application can
still hang a bit during shutdown. The active connections
can be listed using lsof or ss commands.

    ss -p sport = 1984
    ss -p dport = 1984 # set remove peer port here

Where 1984 is the port used by arweave server. This
command will list all connections with their state on
the process using them. To kill them manually - only on
Linux - the flag -K can be added.

    ss -pK sport = 1984
    ss -pK dport = 1984 # set remove peer port here

This is not a clean way of closing sockets, but if the
service got stuck because of a remaining connections,
it can help.  Another solution to list available peers
is to filter the program name, usually beam:

    ss -p | grep beam

This commit fixes also the test suite. Cowboy
required a proplist for its transport parameters,
if not configured correctly, some tests are
failing.


apps/arweave/src/ar_config.erl
```diff
@@ -721,6 +721,17 @@ parse_options([{<<"shutdown_tcp_connection_timeout">>, Delay} | Rest], Config)
 		parse_options(Rest, NewConfig);
 parse_options([{<<"shutdown_tcp_connection_timeout">>, InvalidValue} | Rest], Config) ->
 	{error, {bad_type, shutdown_tcp_connection_timeout, integer}, InvalidValue};
+parse_options([{<<"shutdown_tcp_mode">>, Mode}|Rest], Config) ->
+	case Mode of
+		"soft" ->
```
JamesPiechota, line 726: I think Mode may be a binary string when specificied in the config.json. e.g. `<<"soft">>`

apps/arweave/src/ar_shutdown_manager.erl
```diff
@@ -0,0 +1,598 @@
...
+
+%%--------------------------------------------------------------------
+%% @hidden
+%% @doc terminate all active http connections from ranch/cowboy. This
+%% @end
+%%--------------------------------------------------------------------
+terminate_connections() ->
+	% this process should not exit, it will be linked to another
+	% process in charge to kill all connections.
+	erlang:process_flag(trap_exit, true),
```
JamesPiechota, line 104: why do we need to call this twice (both here and during init/0)? (asking to understand, not to question)

apps/arweave/src/ar_shutdown_manager.erl
```diff
@@ -0,0 +1,598 @@
...
+		reverse => boolean()
+	},
+	Return :: [map()].
+
+data_filters(_, [], _) -> [];
+data_filters([], Datas, _) -> Datas;
+data_filters(Filters, Datas, Opts) ->
+	data_filters(Filters, Datas, [], Opts).
+
+data_filters_test() ->
```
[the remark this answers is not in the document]
humaite: I don't think it's critical. This part was added only for debugging purpose for now. It helped me to filter sockets. It's not used in the main code.

apps/arweave/src/ar.erl
```diff
@@ -372,7 +372,61 @@ show_help() ->
...
+			{"http_api.tcp.linger_timeout",
+				"Configure HTTP Server TCP linger timeout parameter."
+			},
+			{"http_api.tcp.nodelay",
+				"Configure HTTP Server TCP nodelay parameter."
+			},
+			{"http_api.tcp.nopush",
+				"Configure HTTP Server TCP nopush parameter."
+			},
+			{"http_api.tcp.send_timeout_close",
```
JamesPiechota: I think `http_api.tcp.read_ahead` is missing from the help

apps/arweave/src/ar_config.erl
```diff
@@ -721,6 +721,144 @@ parse_options([{<<"shutdown_tcp_connection_timeout">>, Delay} | Rest], Config)
...
+			parse_options(Rest, NewConfig);
+		Mode ->
+			{error, {bad_value, shutdown_tcp_mode}, Mode}
+	end;
+
+%% Gun client parameters
+parse_options([{<<"http_client.http.keepalive">>, Timeout}|Rest], Config) ->
+	case Timeout of
+		<<"infinity">> ->
+			parse_options(Rest, Config#config{ 'http_client.http.keepalive' = infinity });
```
JamesPiechota, line 758: FYI: The cli values for this option are true/false vs. here they are infinity/integer.

apps/arweave/src/ar_config.erl
```diff
@@ -721,6 +721,144 @@ parse_options([{<<"shutdown_tcp_connection_timeout">>, Delay} | Rest], Config)
...
+parse_options([{<<"http_api.tcp.delay_send">>, Delay}|Rest], Config) ->
+	case Delay of
+		_ when is_boolean(Delay) ->
+			parse_options(Rest, Config#config{ 'http_api.tcp.delay_send' = Delay });
+		_ ->
+			{error, {bad_value, 'http_api.tcp.delay_send'}, Delay}
+	end;
+parse_options([{<<"http_api.tcp.keepalive">>, Keepalive}|Rest], Config) ->
+	case Keepalive of
+		_ when is_boolean(Keepalive) ->
```
JamesPiechota, line 860: Ditto above: config values are true/false but cli options are infinity/integer

Maybe the tcp.keepalive and http.keepalive were swapped?

apps/arweave/src/ar_shutdown_manager.erl
JamesPiechota: the `erlang:process_flag(trap_exit, true),` is still called in `terminate_connections` and `init` - is this intentional?

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -1434,3 +1435,34 @@ add_header(Name, Value, Headers) ->
 	?LOG_ERROR([{event, invalid_header}, {name, Name}, {value, Value}]),
 	Headers.
 
+%%--------------------------------------------------------------------
+%% @doc stop the application and execute the shutdown procedure.
+%% @end
+%%--------------------------------------------------------------------
+-spec stop(State) -> Return when
```
JamesPiechota: Where does `stop` get called from? Is it part of the normal gen_server termination process? Or are we calling it explicitly somwhere? (having trouble finding where)

apps/arweave/src/ar_node_sup.erl
```diff
@@ -69,7 +69,7 @@ ar_http_iface_server_spec() ->
 	#{ id => ar_http_iface_server
 	 , start => {ar_http_iface_server, start_link, []}
 	 , type => worker
-	 , shutdown => ?SHUTDOWN_TCP_MAX_CONNECTION_TIMEOUT
+	 , shutdown => ?SHUTDOWN_TCP_CONNECTION_TIMEOUT*2*1000
```
JamesPiechota, line 72: Should we be grabbing `SHUTDOWN_TCP_CONNECTION_TIMEOUT` from config now?

apps/arweave/src/ar_config.erl
humaite: HTTP keepalive is different than TCP keepalive. `http_client.http.keepalive` configures directly gun and the value can be `infinity` (disabled) or a positive integer:

> keepalive (infinity)
> Time between pings in milliseconds. Since the HTTP protocol has no standardized way to ping the server, Gun will simply send an empty line when the connection is idle. Gun only makes a best effort here as servers usually have configurable limits to drop idle connections. Disabled by default due to potential incompatibilities.
> -- https://ninenines.eu/docs/en/gun/2.2/manual/gun/#_http_opts

apps/arweave/src/ar_config.erl
humaite: `http_client.tcp.keepalive` and `http_api.tcp.keepalive` are configuring the TCP socket directly. This is boolean.

> {keepalive, Boolean} (TCP/IP sockets) - Enables/disables periodic transmission on a connected socket when no other data is exchanged. If the other end does not respond, the connection is considered broken and an error message is sent to the controlling process. Defaults to false.
> https://www.erlang.org/doc/apps/kernel/inet.html

apps/arweave/src/ar_config.erl
humaite: Indeed, `http_api.tcp.keepalive` is wrong from `ar.erl`. I will fix that.

apps/arweave/src/ar_node_sup.erl
humaite: I think this part needs to be improved, but this value stay like that for now.

apps/arweave/src/ar_http_iface_client.erl
humaite: Removed. It was some garbage from the previous implementation.

apps/arweave/src/ar_shutdown_manager.erl
humaite: It's intentional. The `terminate_connections/0` function can be called manually by an operator, outside of the `ar_shutdown_manager` process. Because this function is blocking, waiting for spawned processes to be killed, this is required.

JamesPiechota (approved):

PR #812, opened 2025-05-22
Lb/sync replica 2 9

ldmberman:


apps/arweave/src/ar_data_sync.erl
```diff
@@ -3422,6 +3453,37 @@ get_merkle_rebase_threshold() ->
...
+%% 1020-2048. This means the node has the first 10 chunks of the first entropy footprint,
+%% the last 24 chunks of the first entropy footprint and chunks 20-48 of the second
+%% entropy footprint. These chunks are from the first partition. The offset of the chunks
+%% from the second partition is shifted by the number of chunks in the replica 2.9
+%% entropy generated per partition (which is slightly bigger than the number of chunks
+%% that can fit in the 3.6 TB partition).
+%%
+%% Note that Packing does not have to be replica_2_9. We maintain this record
+%% for any packing so that it is convenient to serve the data to any client.
+add_footprint(PaddedOffset, Packing, StoreID) ->
```
JamesPiechota, line 3469: `add_footprint` is adding a single chunk to the footprint record? Or its adding a full footprint?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -3422,6 +3453,37 @@ get_merkle_rebase_threshold() ->
...
+	Offset = get_footprint_offset(PaddedOffset),
+	ar_sync_record:add(Offset, Offset - 1, Packing, ar_data_sync_footprints, StoreID).
+
+get_footprint_offset(PaddedOffset) ->
+	FootprintsPerPartition = ?REPLICA_2_9_ENTROPY_COUNT div ?COMPOSITE_PACKING_SUB_CHUNK_COUNT,
+	FootprintSize = ?REPLICA_2_9_ENTROPY_SIZE div ?COMPOSITE_PACKING_SUB_CHUNK_SIZE,
+	Partition = ar_replica_2_9:get_entropy_partition(PaddedOffset),
+	PartitionStartOffset = Partition * FootprintsPerPartition * FootprintSize,
```
JamesPiechota, line 3477: I think this calculation will cause the partition start offset to drift further and further from the recall partition start offset as we get to higher partitions? If I've read it correctly?

My understanding is that we now have the entropy start and partition start mostly aligned for all partitions (entropy start might start a partial chunk later so we don't split buckets, but no more than one chunk later i think)

apps/arweave/src/ar_data_sync.erl
```diff
@@ -3493,7 +3555,7 @@ process_disk_pool_immature_chunk_offset(Iterator, TXRoot, TXPath, AbsoluteOffset
 		false ->
 			{Offset, _, ChunkSize, DataRoot, DataPathHash, ChunkDataKey, Key, _, _, _} = Args,
 			case update_chunks_index({AbsoluteOffset, Offset, ChunkDataKey,
-					TXRoot, DataRoot, TXPath, ChunkSize, unpacked}, State) of
+					TXRoot, DataRoot, TXPath, ChunkSize, unpacked}, false, State) of
```
JamesPiechota, line 3558: Why don't we add the footprint here? Is it because we're in the diskpool and only want to track footprints for persisted storage modules?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -2026,6 +2094,66 @@ handle_get_data_sync_record(Start, End, Limit, Req) ->
...
+						{400, #{}, jiffy:encode(#{ error => footprint_number_too_large }), Req};
+					false ->
+						{ok, {StoreID2, Packing2, FootprintSize, FootprintsPerPartition}}
+				end;
+			Reply2 ->
+				Reply2
+		end,
+	case CheckFootprintNumber of
+		{ok, {StoreID3, Packing3, FootprintSize2, FootprintsPerPartition2}} ->
+			PartitionStartOffset = Partition * FootprintsPerPartition2 * FootprintSize2,
```
JamesPiechota, line 2131: Tagging in case similar comment applies: https://github.com/ArweaveTeam/arweave/pull/812/files#r2103123546

apps/arweave/src/ar_data_sync.erl
ldmberman: A single chunk

apps/arweave/src/ar_data_sync.erl
ldmberman: Yes, we need to support removing everything above the given offset for the disk pool data (ar_sync_record:cut), which is expensive for footprints because we need to potentially iterate over many footprints removing the data from each.

apps/arweave/src/ar_http_iface_server.erl
```diff
@@ -30,7 +30,8 @@
 
 -define(ENDPOINTS, ["info", "block", "block_announcement", "block2", "tx", "tx2",
 		"queue", "recent_hash_list", "recent_hash_list_diff", "tx_anchor", "arql", "time",
-		"chunk", "chunk2", "data_sync_record", "sync_buckets", "wallet", "unsigned_tx",
+		"chunk", "chunk2", "data_sync_record", "sync_buckets", "footprint_buckets",
+		"wallet", "unsigned_tx",
```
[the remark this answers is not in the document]
ldmberman: I added `name_route([<<"footprints">>, _Partition, _Number])` below, we do not add a short GET /footprints endpoint

apps/arweave/src/ar_data_discovery.erl
```diff
@@ -67,6 +70,32 @@ get_bucket_peers(Bucket, Cursor, Peers) ->
...
+	end.
+
+get_footprint_bucket_peers(Bucket, Cursor, Peers) ->
+	case ets:next(ar_data_discovery_footprint_buckets, Cursor) of
+		'$end_of_table' ->
+			UniquePeers = sets:to_list(sets:from_list(Peers)),
+			PickedPeers = pick_peers(UniquePeers, ?QUERY_BEST_PEERS_COUNT),
+			PickedPeers;
+		{Bucket, _Share, Peer} = Key ->
+			get_footprint_bucket_peers(Bucket, Key, [Peer | Peers]);
```
JamesPiechota, line 90: This logic will cause it to keep advancing through the ETS table accumulating Peers until it hits a entry from a new bucket, right? At which point the `_` branch is entered and the peers are picked

apps/arweave/src/ar_data_discovery.erl
```diff
@@ -105,16 +135,13 @@ handle_cast(update_network_data_map, #state{ peers_pending = N } = State)
...
-						{ok, SyncBuckets} ->
-							gen_server:cast(?MODULE, {add_peer_sync_buckets, Peer,
-									SyncBuckets});
-						{error, request_type_not_found} ->
-							get_sync_buckets(Peer);
-						Error ->
-							?LOG_DEBUG([{event, failed_to_fetch_sync_buckets},
-								{peer, ar_util:format_peer(Peer)},
-								{reason, io_lib:format("~p", [Error])}])
+					fetch_sync_buckets(Peer),
```
JamesPiechota, line 146: Do we want to disable the legacy syncing if `footprint_based_syncing` is enabled? My gut sense from the repack_in_place work is that having individual chunks repacked along side footprints can impact performance negatively - but perhaps the implications is different for syncing/packing.

apps/arweave/src/ar_data_sync.erl
```diff
@@ -964,68 +965,86 @@ handle_cast(sync_data2, State) ->
...
+			false ->
+				noop;
+			true ->
+				End > DiskPoolThreshold
+		end,
+	case IntersectsDiskPool of
+		noop ->
+			noop;
+		true ->
+			gen_server:cast(self(), {collect_peer_intervals, Start, End, normal});
```
JamesPiechota: under `normal` syncing will the node ignore replica.2.9 packed data? i.e. for the tip partition will it only sync from `unpacked` data?

apps/arweave/src/ar_footprint_record.erl
```diff
@@ -0,0 +1,200 @@
...
+-spec get_intervals(
+	Partition :: non_neg_integer(),
+	Footprint :: non_neg_integer(),
+	Packing :: term(),
+	StoreID :: string()
+) -> term().
+get_intervals(Partition, Footprint, Packing, StoreID) ->
+	FootprintSize = get_footprint_size(),
+	FootprintsPerPartition = ?REPLICA_2_9_ENTROPY_COUNT div ?COMPOSITE_PACKING_SUB_CHUNK_COUNT,
+	PartitionStartOffset = Partition * FootprintsPerPartition * FootprintSize,
```
JamesPiechota: I think this calculation will cause the partition start offset to drift further and further from the recall partition start offset as we get to higher partitions? If I've read it correctly?

My understanding is that we now have the entropy start and partition start mostly aligned for all partitions (entropy start might start a partial chunk later so we don't split buckets, but no more than one chunk later i think)

apps/arweave/include/ar_data_discovery.hrl
```diff
@@ -8,6 +8,14 @@
...
 -endif.
 
+%% Similar to ?NETWORK_DATA_BUCKET_SIZE, except for a footprint bucket
+%% contains several "footprints" - sets of chunks spread out across the partition.
+-ifdef(AR_TEST).
+-define(NETWORK_FOOTPRINT_BUCKET_SIZE, 36). % 12 (footprints) * 3 (chunks); ~10 MB
+-else.
+-define(NETWORK_FOOTPRINT_BUCKET_SIZE, 37888). % 37 (footprints) * 1024 (chunks); ~10 GB
+-endif.
+
```
JamesPiechota, line 18: Since footprints now represent 1024 chunks - what do you think about getting rid of the bucket concept? I don't have any good numbers on the performance/network data impacts of having the bucket size be 1 vs. 37 ... but if the performance impact is minimal, eliminating buckets might greatly simplify the code I think?

apps/arweave/src/ar_data_discovery.erl
ldmberman: Yep

apps/arweave/src/ar_data_sync.erl
ldmberman: yes

apps/arweave/src/ar_data_discovery.erl
ldmberman: By legacy syncing, do you mean disk pool syncing?

apps/arweave/src/ar_footprint_record.erl
ldmberman: You mean, we could reduct the actual FootprintSize a bit, and the same logic will keep working?

apps/arweave/include/ar_data_discovery.hrl
ldmberman: Without buckets, we have 3_600_000_000_000 / (1024 * 262144) ~ 13.5k footprint requests per peer per partition to learn about the chunk presence.  The maximum payload size is 250 KB (in practice should be more like 25 KB) so working with 100 peers sums up to tens of gigabytes of extra traffic per partition per discovery cycle. It may work though, I see the appeal.

apps/arweave/src/ar_data_discovery.erl
JamesPiechota: By "legacy" I mean the chunk-by-chunk syncing rather than the new footprint-by-footprint syncing. My sense from repack in place is that if we miss the two approaches within the same storage module it can lead to the chunk-by-chunk packing adding "random" chunks scattered throughout the partition that can mess with the efficiency of the footprint-by-footprint packing. So for repack-in-place I noticed better performance if I didn't mix the "chunk-by-chunk" approach with the "footprint-by-footprint" approach

This behavior may be unique to repack-in-place though (i.e. perhaps related to the logic it uses to determine if a given footprint needs to be repacked) so may be moot for sync/pack.

apps/arweave/src/ar_footprint_record.erl
JamesPiechota: I was thinking that PartitionStartOffset should always be set to `Partition * ?PARTITION_SIZE` (i.e. 3.6TB * partition number)

But I may be misunderstanding the intent of `PartitionStartOffset`

apps/arweave/include/ar_data_discovery.hrl
JamesPiechota: Ah okay, and with buckets the network traffic is reduced by 37x?

ldmberman: Yep

apps/arweave/src/ar_footprint_record.erl
ldmberman: @JamesPiechota I've updated the logic such that the footprint record with all the data will not have holes between partitions (assuming there are no holes in the data, off course) by grounding it on the fact every partition has 3600000221184 div 262144 buckets.

apps/arweave_config/include/arweave_config.hrl
```diff
@@ -133,6 +133,10 @@
 -define(DEFAULT_REPLICA_2_9_WORKERS, 8).
 -endif.
 
+%% The default maximum number of replica 2.9 entropies to cache at a time
+%% while syncing data. Each entropy is 256 MB.
+-define(DEFAULT_REPLICA_2_9_ENTROPY_CACHE_MAX_ENTROPIES, 2).
```
JamesPiechota: I forget - is an "entropy" the same as a "footprint"? E.g. 1 footprint contains 1 entropy

apps/arweave/src/ar_replica_2_9.erl
```diff
@@ -172,7 +171,7 @@ get_entropy_key(RewardAddr, AbsoluteEndOffset, SubChunkStartOffset) ->
 %% area where the 2.9 entropy of every chunk is unique.
 -spec get_sector_size() -> pos_integer().
 get_sector_size() ->
```
JamesPiechota: Do we want to remove all these `ar_replica_2_9` functions now that they've been moved to `ar_block`? (mostly concerned about potential future code diversion or confusion over which function to use)

apps/arweave/src/ar_data_discovery.erl
```diff
@@ -35,6 +38,10 @@
 %% into the peer map every DATA_DISCOVERY_COLLECT_PEERS_FREQUENCY_MS milliseconds.
 -define(DATA_DISCOVERY_COLLECT_PEERS_COUNT, 1000).
 
+%% The number of the release adding support for the
+%% GET /footprint_buckets endpoint.
+-define(GET_FOOTPRINT_BUCKETS_SUPPORT_RELEASE, 87).
```
JamesPiechota: FYI: alpha6 is now `87` and master is now `88` - just in case that impacts your testing

apps/arweave/src/ar_data_sync.erl
```diff
@@ -33,10 +34,17 @@
 -include("ar_data_sync.hrl").
 -include("ar_sync_buckets.hrl").
 
+-ifdef(TEST).
```
JamesPiechota: Is this `TEST` rather than `AR_TEST` on purpose?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -33,10 +34,17 @@
...
+
+%% The key for storing migration cursor in the migrations_index database.
+-define(FOOTPRINT_MIGRATION_CURSOR_KEY, <<"footprint_migration_cursor">>).
+
 -ifdef(AR_TEST).
--define(COLLECT_SYNC_INTERVALS_FREQUENCY_MS, 5_000).
+-define(COLLECT_SYNC_INTERVALS_FREQUENCY_MS, 1_000).
 -else.
--define(COLLECT_SYNC_INTERVALS_FREQUENCY_MS, 300_000).
+-define(COLLECT_SYNC_INTERVALS_FREQUENCY_MS, 10_000).
```
JamesPiechota, line 49: Oh I like reducing this - I always wondered if the 5 minute frequency might be contributing to some of the slow sync times. Do we expect the faster loop time to have some other trade off? Like more network traffic?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -45,6 +53,13 @@
 -define(DEVICE_LOCK_WAIT, 5_000).
 -endif.
 
+%% The number of chunks to migrate per batch during footprint migration.
```
JamesPiechota, line 58: What is the footprint migration? is this a dev-only concept to allow you to migrate data on your test server(s), but we'd remove it before launch? (since no one else has presumably done any footprint syncing yet)

apps/arweave/src/ar_data_sync.erl
```diff
@@ -715,6 +730,20 @@ debug_get_disk_pool_chunks(Cursor) ->
...
+	IsMatchingPacking =
+		case Packing of
+			{replica_2_9, _Addr} ->
+				true;
+			unpacked_padded ->
+				true;
+			_ ->
+				false
+	end,
+	IsMatchingPacking andalso (AbsoluteOffset > ar_block:strict_data_split_threshold() orelse ChunkSize == ?DATA_CHUNK_SIZE).
```
JamesPiechota: What happens to replica_2_9, partial chunks, from before the strict data split threshold? I see they don't get added to the `ar_footprint_record` - does that just mean that they won't be advertised via `GET /footprints`? Or is there another impact as well? (and if they aren't advertised... how will they get synced?)

apps/arweave/src/ar_data_sync.erl
```diff
@@ -981,68 +1017,115 @@ handle_cast(sync_data2, State) ->
...
+			false ->
+				noop;
+			true ->
+				End > DiskPoolThreshold
+		end,
+	%% Alternate between "normal" and footprint-based syncing.
+	%% Footprint-based syncing downloads replica 2.9 chunks footprint by footprint
+	%% to avoid redundant entropy generations for unpacking. "Normal" syncing ignores
+	%% replica 2.9 data and mostly downloads unpacked data from peers storing it.
+	SyncPhase2 =
```
JamesPiechota, line 1081: is this the basic stucture:
1. Collect peer intervals for the range `range_start` to `range_end` for all peers which advertise data in that range
2. first time through we sync "normally"
3. Everytime we complete syncing the `range_start` to `range_end` range, we start over and flip the move from `normal` to `footprint` or back

so we do a full pass through the range in one mode, then a full pass in the other mode, etc...

apps/arweave/src/ar_peer_intervals.erl
```diff
@@ -26,79 +31,187 @@
...
-			],
-			Peers = ar_data_discovery:pick_peers(HotPeers, ?QUERY_BEST_PEERS_COUNT),
-
-			case length(Peers) < ?QUERY_BEST_PEERS_COUNT of
+		{End2, EnqueueIntervals} = do_fetch(Start, End, StoreID, normal),
+		gen_server:cast(Parent, {enqueue_intervals, EnqueueIntervals}),
+		gen_server:cast(Parent, {collect_peer_intervals, End2, End, normal})
+	end);
+
+fetch(Start, End, StoreID, footprint) when Start >= End ->
```
JamesPiechota: can this be merged with the `normal` code? Looks like they are the same, only difference is the `Type` value?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -3807,3 +3956,73 @@ record_chunk_cache_size_metric() ->
 		_ ->
 			ok
 	end.
+
+may_be_run_footprint_record_initialization(State) ->
```
JamesPiechota:
```suggestion
maybe_run_footprint_record_initialization(State) ->
```

apps/arweave/src/ar_data_sync.erl
JamesPiechota: Ah I think I see. if a node already has r29 data from before this patch, then we need to migrate all that packed r29 data into the footprint records (so that it will be stored in both the sync_record and the footprint_record)

apps/arweave/src/ar_footprint_record.erl
```diff
@@ -0,0 +1,462 @@
...
+-include_lib("eunit/include/eunit.hrl").
+
+-moduledoc """
+    This module exports functions for maintaining
+    a replica 2.9 entropy-aligned record of the synced chunks.
+    It differs from the normal record (ar_data_sync) in that it only registers
+    the bucket numbers of the synced chunks and records chunks with the same footprint
+    next to each other. For example, a record may contain intervals 0-10, 1000-1024,
+    1020-2048. This means the node has the first 10 chunks of the first entropy footprint,
+    the last 24 chunks of the first entropy footprint and chunks 20-48 of the second
```
JamesPiechota:
```suggestion
    the last 24 chunks of the first entropy footprint and chunks 24-48 of the second
```

Should this be chunks 24-48? Or is the overlap between footprint 1 and 2 intentional?

apps/arweave/src/ar_footprint_record.erl
```diff
@@ -0,0 +1,462 @@
...
+
+%%%===================================================================
+%%% Public interface.
+%%%===================================================================
+
+%% @doc Add a chunk to the footprint record.
+-spec add(Offset :: non_neg_integer(), Packing :: term(), StoreID :: string()) -> ok.
+add(Offset, Packing, StoreID) ->
+	PaddedOffset = ar_block:get_chunk_padded_offset(Offset),
+	FootprintOffset = get_offset(PaddedOffset),
```
JamesPiechota:
```suggestion
	FootprintOffset = get_offset(Offset),
```

apps/arweave/src/ar_util.erl
```diff
@@ -247,6 +248,11 @@ unique(Res, [X|Xs]) ->
 		true -> unique(Res, Xs)
 	end.
 
+%% @doc Pad the given value to the closest multiple equal or above the given value.
+-spec pad_to_closest_multiple_equal_or_above(Value :: non_neg_integer(), Multiple :: non_neg_integer()) -> non_neg_integer().
+pad_to_closest_multiple_equal_or_above(Value, Multiple) ->
```
JamesPiechota, line 253: Is this the same as `ar_util:ceil_int`?

Hm, or is the difference that `ceil_int` always rounds up, but pad_to_xxxx may not round up if it's already at a multiple?

apps/arweave/src/ar_footprint_record.erl
```diff
@@ -0,0 +1,462 @@
...
+
+	ChunksPerPartition = ar_util:pad_to_closest_multiple_equal_or_above(?PARTITION_SIZE, ?DATA_CHUNK_SIZE) div ?DATA_CHUNK_SIZE,
+	Partition = ar_replica_2_9:get_entropy_partition(PaddedOffset),
+	PartitionOffset = (PaddedOffset - Partition * ?PARTITION_SIZE) div ?DATA_CHUNK_SIZE - 1,
+
+	%% Which footprint within the partition
+	Footprint = PartitionOffset rem FootprintsPerPartition,
+	%% Position within the footprint
+	FootprintOffset = PartitionOffset div FootprintsPerPartition,
+	Partition * ChunksPerPartition + Footprint * FootprintSize + FootprintOffset + 1.
```
JamesPiechota, line 64: What kind of chunk offset does this yield? e.g. EndOffset? PaddedEndOffset? StartOffset?

apps/arweave/src/ar_global_sync_record.erl
```diff
@@ -107,14 +110,22 @@ handle_call(Request, _From, State) ->
 	?LOG_WARNING([{event, unhandled_call}, {module, ?MODULE}, {request, Request}]),
 	{reply, ok, State}.
 
-handle_cast(update_serialized_sync_buckets, State) ->
+handle_cast({update_serialized_sync_buckets, serialized_sync_buckets = Key}, State) ->
```
JamesPiechota, line 109: Out of curiosity: why assign `serialized_sync_buckets` to `Key` rather than just use the `serialized_sync_buckets` atom directly?

apps/arweave/src/ar_packing_server.erl
```diff
@@ -243,13 +245,30 @@ encipher_replica_2_9_chunk(Chunk, Entropy) ->
...
 	Key = ar_replica_2_9:get_entropy_key(RewardAddr, BucketEndOffset, SubChunkStartOffset),
-	PackingState = get_packing_state(),
-	RandomXState = get_randomx_state_by_packing({replica_2_9, RewardAddr}, PackingState),
 
-	Entropy = ar_mine_randomx:randomx_generate_replica_2_9_entropy(RandomXState, Key),
-	%% Primarily needed for testing where the entropy generated exceeds the entropy
-	%% needed for tests.
-	binary_part(Entropy, 0, ?REPLICA_2_9_ENTROPY_SIZE).
+	entropy_generation_lock(Key, RewardAddr, BucketEndOffset, SubChunkStartOffset),
+	case ar_replica_2_9_entropy_cache:get(Key) of
```
JamesPiechota: Is the entropy cache needed now to handle 2 chunks that need to be packed to the same entropy but one chunk comes from `normal` syncing and one from `footprint` syncing? Or asked another way: why do we need the entropy cache? when do we expect there to be cache hit?

apps/arweave/src/ar_packing_server.erl
```diff
@@ -868,6 +891,60 @@ exor_replica_2_9_sub_chunks(
...
+	ets:delete(?MODULE, {entropy_generation_lock, Key}).
+
+update_entropy_generation_stats(Key, RewardAddr, BucketEndOffset, SubChunkStartOffset) ->
+	Tab = entropy_generation_stats,
+	Time = erlang:monotonic_time(millisecond),
+	ets:update_counter(Tab, Key, {2, 1}, {Key, 0, Time}),
+	may_be_report_redundant_entropy_generation(Key, RewardAddr, BucketEndOffset, SubChunkStartOffset),
+	remove_outdated_entropy_generation_stats().
+
+may_be_report_redundant_entropy_generation(Key, RewardAddr, BucketEndOffset, SubChunkStartOffset) ->
```
JamesPiechota: I remember you mentioning redundant entropy generation a little while back and were going to add some debugging functions to help. Are they still needed?

apps/arweave/src/ar_peer_intervals.erl
```diff
@@ -231,3 +343,292 @@ get_peer_intervals(Peer, Left, SoughtIntervals) ->
...
+	end.
+
+print_peers(Peers) ->
+	string:join([io_lib:format("~p", [ar_util:format_peer(Peer)]) || Peer <- Peers], ", ").
+
+%%%===================================================================
+%%% Tests
+%%%===================================================================
+
+-ifdef(TEST).
```
JamesPiechota: `TEST` vs. `AR_TEST` - is this intentional?

apps/arweave/src/ar_storage_module.erl
```diff
@@ -124,8 +124,9 @@ packing_label(Packing) ->
 
 %% @doc Return the storage module with the given identifier or not_found.
 %% Search across both attached modules and repacked in-place modules.
-get_by_id(?DEFAULT_MODULE) ->
-	?DEFAULT_MODULE;
+get_by_id(Atom) when is_atom(Atom) ->
```
JamesPiechota, line 130: `?DEFAULT_MODULE` is `"default"` rather than `default` - will this still handle that case correctly?

apps/arweave/src/ar_sync_record.erl
```diff
@@ -52,7 +52,9 @@
 	%% The index of the storage module; undefined for the "default" storage.
 	storage_module_index,
 	%% The number of entries in the write-ahead log.
-	wal
+	wal,
+	%% Whether the sync record is in memory only.
```
JamesPiechota, line 56: Is this only true for tests?

apps/arweave_config/include/arweave_config.hrl
ldmberman: yes

apps/arweave/src/ar_replica_2_9.erl
ldmberman: Makes sense, done [here](https://github.com/ArweaveTeam/arweave/pull/812/commits/4cbd4ff0fc995b4c3e39a87642c0ed764e8a5121)

apps/arweave/src/ar_data_sync.erl
ldmberman: no, good catch, changed to AR_TEST

apps/arweave/src/ar_data_sync.erl
ldmberman: Yes, nodes will ask peers for the intervals (they likely do not have at this point) more frequently.

apps/arweave/src/ar_data_sync.erl
ldmberman: They won't be advertised via GET /footprints, yes. They may be still synced as unpacked chunks.

apps/arweave/src/ar_data_sync.erl
ldmberman: yes, correct!

apps/arweave/src/ar_data_sync.erl
JamesPiechota: Ah got it. The set of "partial chunks before the strict data split threshold" and the set of "chunks that we store unpacked in rocksdb" are the same?

apps/arweave/src/ar_util.erl
ldmberman: yep

apps/arweave/src/ar_sync_record.erl
ldmberman: yes, the new tests in ar_peer_intervals.erl

apps/arweave/src/ar_global_sync_record.erl
ldmberman: for brevity and to make it slightly less error-prone (like when there is a typo in one of the atoms)

apps/arweave/src/ar_footprint_record.erl
ldmberman: There is no such distinction, a chunk is mapped to a bucket and the entire bucket is registered in the footprint record.

apps/arweave/src/ar_storage_module.erl
ldmberman: Hm, good catch, I think I removed it by mistake

apps/arweave/src/ar_packing_server.erl
ldmberman: I see some possibilities when the cache may be not hit:

- we cannot prepare the entropy fast enough for the chunks fetched during the "normal" sync (they are fetched unpacked);
- we process several footprints at a time (perhaps, collecting small leftovers; perhaps, syncing too fast)

apps/arweave/src/ar_footprint_record.erl
ldmberman: good catch, I need to make it clearer

apps/arweave/src/ar_packing_server.erl
ldmberman: Yes, I think this log may be noisy occasionally, but I would like to watch for it

apps/arweave/src/ar_peer_worker.erl
```diff
@@ -0,0 +1,962 @@
...
+%%% Public interface.
+%%%===================================================================
+
+start_link(Peer) ->
+	Name = peer_name(Peer),
+	gen_server:start_link({local, Name}, ?MODULE, [Peer], []).
+
+%% @doc Convert peer tuple to a process name atom.
+peer_name(Peer) ->
+	list_to_atom("ar_peer_" ++ ar_util:format_peer(Peer)).
```
ldmberman: This may potentially exhaust the atom table

JamesPiechota: Good catch!

Updated: https://github.com/ArweaveTeam/arweave/pull/812/commits/3aac97a97672bf822ebd4f2ebb50eb8b6a09e514

PR #826, opened 2025-07-10
Fix Timeout Issue when peers clock are not sync

humaite:
erlang:halt/0 or erlang:halt/1 should not be used, or only if something really bad happens. usually,
when an erlang application must be stopped, even
more when it's a release, init:halt/0 or init:halt/1 should be used instead. init:halt functions will
take care to close everything before the shutdown. Using erlang:halt functions are doing the opposite, and generate a core dump, without cleaning anything, it could also explain why many time the peers are in bad state after a shutdown.

If only a few part of the trusted peers have a bad clock, it should not be a problem. The previous
method was to stop arweave when one peers had
some trouble, if more than 10 nodes are configured (for example peers.arweave.xyz), then it should not be a problem.

see: https://github.com/ArweaveTeam/arweave-dev/issues/907

### DIFF
--- a/apps/arweave/src/ar.erl
+++ b/apps/arweave/src/ar.erl
@@ -518,7 +518,7 @@ show_help() ->
 			)}
 		]
 	),
-	erlang:halt().
+	init:stop(1).
 
 parse_config_file(["config_file", Path | Rest], Skipped, _) ->
 	case read_config_from_file(Path) of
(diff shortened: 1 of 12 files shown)


JamesPiechota (approved):
LGTM. My only concern is simply not knowing whether `init:stop(1)` vs. `erlang:halt()` will have any unintended consequences. Since we're hoping to push out a release next week (perhaps optmistically...), do you think we'll have time to merge and deploy this change to our internal nodes to get some validation that there aren't any unintended consequences?

PR #829, opened 2025-07-10
new VDF opt

vird:
adapted from optimizations made by discord user @hihui

### DIFF
--- a/apps/arweave/c_src/Makefile
+++ b/apps/arweave/c_src/Makefile
@@ -14,9 +14,12 @@ else
 	CXXFLAGS ?= -O3
 endif
 
-# _mm_crc32_u32 support
-CFLAGS += -msse4.2
-CXXFLAGS += -msse4.2
+UNAME_SYS := $(shell uname -s)
+ifeq ($(UNAME_SYS), Linux)
+	# _mm_crc32_u32 support
+	CFLAGS += -msse4.2
+	CXXFLAGS += -msse4.2
+endif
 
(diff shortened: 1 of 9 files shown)


apps/arweave/c_src/Makefile
```diff
@@ -84,12 +86,20 @@ RX4096_OBJECTS = $(addsuffix .o, $(basename $(RX4096_SOURCES)))
...
-UNAME_SYS := $(shell uname -s)
+# NOTE tabs here will cause build fail
 ifeq ($(UNAME_SYS), Linux)
   $(C_SRC_DIR)/vdf/vdf_exp_x86.o: CXXFLAGS += -msha
 endif
 ifeq ($(UNAME_SYS), Darwin)
   $(C_SRC_DIR)/vdf/vdf_exp_arm.o:  CXXFLAGS += -march=armv8-a+crypto
+  $(C_SRC_DIR)/vdf/vdf_exp_arm2.o: CXXFLAGS += -march=armv8-a+crypto
+endif
+ifeq ($(UNAME_SYS), Darwin)
```
JamesPiechota, line 97:
```suggestion
```

vird: I'm not sure that it's good idea. Make works really strange with OBJ: CMD commands
So this delimiter ensures that if we accidentally make (C_SRC_DIR)/vdf/vdf_exp_arm2.o: CXXFLAGS += -march=armv8-a+crypto multiline in future this will not capture next line as part of command

vird: Also please note there are spaces above and tabs below
Code above works only with zero indentation or spaces (as far last time I've checked)

JamesPiechota: ah got it. Okay yeah if the space/tabs are different in each if block makes sense to keep them separate

PR #833, opened 2025-07-15
Feat/modular vdf impl

JamesPiechota:


apps/arweave/c_src/vdf/ar_vdf_nif.c
```diff
@@ -137,7 +146,47 @@ static ERL_NIF_TERM vdf_sha2_exp_nif(ErlNifEnv* envPtr, int argc, const ERL_NIF_
 	size_t outCheckpointSize = VDF_SHA_HASH_SIZE*checkpointCount;
 	ERL_NIF_TERM outputTermCheckpoint;
 	unsigned char* outCheckpoint = enif_make_new_binary(envPtr, outCheckpointSize, &outputTermCheckpoint);
-	vdf_sha2_exp_ptr(Salt.data, Seed.data, temp_result, outCheckpoint, checkpointCount, skipCheckpointCount, hashingIterations);
+	vdf_sha2_fused_ptr(Salt.data, Seed.data, temp_result, outCheckpoint, checkpointCount, skipCheckpointCount, hashingIterations);
```
JamesPiechota, line 149: What do you think about refactoring the body of `vdf_sha2_fused_ptr` and `vdf_sha2_fused_nif` to eliminate the duplicate code? It looks like its onlyt he `xxx_ptr` that differs between the two? So maybe that could be an argument to the refactored helper function.

apps/arweave/src/ar_config.erl
```diff
@@ -190,6 +190,17 @@ parse_options([{<<"verify_samples">>, <<"all">>} | Rest], Config) ->
 parse_options([{<<"verify_samples">>, Opt} | _], _) ->
 	{error, {bad_type, verify_samples, number}, Opt};
 
+parse_options([{<<"vdf">>, Mode} | Rest], Config) ->
+	ParsedMode = case Mode of
+		"openssl" ->openssl;
```
JamesPiechota, line 195: should we also include `opensslite` for consistency with the CLI args?

apps/arweave/c_src/vdf/ar_vdf_nif.c
vird: About inlining. I prefer to "do not allow compiler can do stupid thing". Any inlining as notation can be ignored by compiler. Alternative using define macro, but in this case it will become a bit messy

apps/arweave/src/ar_config.erl
vird: As discussed, we probably want just drop openssl, and just have one option "openssl" which just will use openssllite

PR #835, opened 2025-07-16
Add Erlang R26 support

humaite:
Arweave is working correctly on R26 and the next release should offer a version of it running on this Erlang version.


JamesPiechota:
Oh! Thanks for doing this @humaite ! I was just about to look into it - looks great. Once you merge I can give it a try for the next release.

PR #836, opened 2025-07-16
Lb/tx ignore registry fix

ldmberman:


apps/arweave/src/ar_mining_worker.erl
```diff
@@ -1008,10 +1015,29 @@ mark_second_recall_range_missing(Nonce, NoncesLeft, Candidate, State) ->
 		{error, Reason} ->
 			%% NB: this clause may cause a memory leak, because mining worker will wait for
 			%% chunk2 to arrive.
-			?LOG_ERROR([{event, mining_worker_failed_to_add_chunk_to_cache}, {reason, Reason}]),
+			log_error(mining_worker_failed_to_add_chunk_to_cache, Reason, []),
 			mark_second_recall_range_missing(Nonce + 1, NoncesLeft - 1, Candidate, State)
 	end.
 
+-ifdef(AR_TEST).
```
JamesPiechota: I'd remove this bit. The cache_limit_exceeded log indicates a real problem (it's being looked into). I pushed a commit [here](https://github.com/ArweaveTeam/arweave/commit/316aa0284d0e93ef5faa6c053ae234e3c806b1a0) that should temporarily silence it by increasing the default cache limit in tests - once we address the root issue we'll revert that commit

PR #837, opened 2025-07-16
Mining cache fixes

shizzard:
This PR fixes the following issues:
- Mining cache leak, causing some cached subchunks to stay in cache until session rotates;
- OOM node crashes, caused by cached values holding for large binaries, preventing those to be garbage-collected;
- A bug, in some cases causing the mining server to crash on solution validation.

This PR also introduces the mining cache anomaly scanner, that will report cached records in invalid state.

### DIFF
--- a/apps/arweave/src/ar_mining_cache.erl
+++ b/apps/arweave/src/ar_mining_cache.erl
@@ -147,15 +147,18 @@ release_for_session(SessionId, Size, Cache0) ->
 	Cache1 :: #ar_mining_cache{}.
 drop_session(SessionId, Cache0) ->
 	?LOG_DEBUG([{event, drop_session}, {session_id, SessionId}]),
-	{Session, Sessions} = maps:take(SessionId, Cache0#ar_mining_cache.mining_cache_sessions),
-	maybe_search_for_anomalies(SessionId, Session),
-	Cache0#ar_mining_cache{
-		mining_cache_sessions = Sessions,
-		mining_cache_sessions_queue = queue:filter(
-			fun(SessionId0) -> SessionId0 =/= SessionId end,
-			Cache0#ar_mining_cache.mining_cache_sessions_queue
-		)
-	}.
+	case maps:take(SessionId, Cache0#ar_mining_cache.mining_cache_sessions) of
+		{Session, Sessions} ->
+			maybe_search_for_anomalies(SessionId, Session),
+			Cache0#ar_mining_cache{
+				mining_cache_sessions = Sessions,
+				mining_cache_sessions_queue = queue:filter(
+					fun(SessionId0) -> SessionId0 =/= SessionId end,
+					Cache0#ar_mining_cache.mining_cache_sessions_queue
+				)
+			};
+		_ -> Cache0
+	end.
 
 %% @doc Checks if a session exists in the cache.
 -spec session_exists(SessionId :: term(), Cache0 :: #ar_mining_cache{}) ->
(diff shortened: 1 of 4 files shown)


apps/arweave/src/ar_mining_worker.erl
```diff
@@ -993,15 +993,15 @@ mark_single_chunk2_missing_or_drop(Nonce, NoncesLeft, Candidate, State) ->
 %% @doc Mark the chunk2 as missing for the whole recall range.
 mark_second_recall_range_missing(Candidate, State) ->
 	#mining_candidate{ packing_difficulty = PackingDifficulty } = Candidate,
-	mark_second_recall_range_missing(0, ar_block:get_max_nonce(PackingDifficulty), Candidate, State).
+	mark_second_recall_range_missing(0, ar_block:get_nonces_per_recall_range(PackingDifficulty), Candidate, State).
```
JamesPiechota, line 996: interesting - was this the change that was causing an occasional sub-chunk to leak?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -733,11 +736,11 @@ process_sub_chunk(chunk2, Candidate, SubChunk, State) ->
...
 			(#ar_mining_cache_value{h1 = undefined} = CachedValue) ->
 				%% H1 is not yet calculated, cache the chunk2 for this nonce.
-				{ok, CachedValue#ar_mining_cache_value{chunk2 = SubChunk}};
-			(#ar_mining_cache_value{h1 = H1} = CachedValue) ->
+				{ok, CachedValue#ar_mining_cache_value{ chunk2 = SubChunk }};
+			(#ar_mining_cache_value{h1 = H1, chunk1 = Chunk1} = CachedValue) ->
 				%% H1 is already calculated, compute H2 and cache the chunk2 for this nonce.
-				ar_mining_hash:compute_h2(self(), Candidate2#mining_candidate{ h1 = H1 }),
-				{ok, CachedValue#ar_mining_cache_value{chunk2 = SubChunk}}
+				ar_mining_hash:compute_h2(self(), Candidate2#mining_candidate{ h1 = H1, chunk1 = Chunk1 }),
```
JamesPiechota, line 742: is the the `byte_size(not_set)` fix?

apps/arweave/src/ar_mining_cache.erl
```diff
@@ -146,13 +146,19 @@ release_for_session(SessionId, Size, Cache0) ->
...
 drop_session(SessionId, Cache0) ->
-	Cache0#ar_mining_cache{
-		mining_cache_sessions = maps:remove(SessionId, Cache0#ar_mining_cache.mining_cache_sessions),
-		mining_cache_sessions_queue = queue:filter(
-			fun(SessionId0) -> SessionId0 =/= SessionId end,
-			Cache0#ar_mining_cache.mining_cache_sessions_queue
-		)
-	}.
+	?LOG_DEBUG([{event, drop_session}, {session_id, SessionId}]),
+	case maps:take(SessionId, Cache0#ar_mining_cache.mining_cache_sessions) of
```
JamesPiechota, line 150: Was this `case` added so that we only search for anomalies when actually dropping a session? or does it have another purpose too?

apps/arweave/src/ar_mining_worker.erl
shizzard: Yes, this is the one I was sure I fixed previously. If you remember, there was off-by-one error, and I'm pretty sure I did apply the fix for it. After that I didn't even think of the bug being the same thing once again.

apps/arweave/src/ar_mining_worker.erl
shizzard: Yes, this is the one. 
When we handle the chunks, the mining candidate for a nonce is virtually "split" into two, one sent to calculate the H1, and another for H2.

apps/arweave/src/ar_mining_cache.erl
shizzard: No, just that. The maximum number of sessions is very small though, you can count it as a constant operation.

shizzard:
> PR looks good! Can you try reverting the Cache limit hack and see if we still get the `cache_limit_exceeded` error in tests? https://github.com/ArweaveTeam/arweave/commit/316aa0284d0e93ef5faa6c053ae234e3c806b1a0
> 
> 
> 
> the hope is that with your fix for the cache leak, the default cache limit should work for tests.

I did try to run it as is, and got errors, so I decided to rebase on master. 
I'll take a look at configuration though.

JamesPiechota:
> I did try to run it as is, and got errors, so I decided to rebase on master. I'll take a look at configuration though.

Got it, the latest errors *might* just be the old-style flakiness we've had for a while. If you're able to get it to green after a 2 or 3 runs, that's a positive indicator. At that point I'd see if reverting the cache limit hack makes things worse.

PR #842, opened 2025-08-03
ar_pricing: use `calendar:system_time_to_universal_time/2`

ariel-anieli:
Hello maintainers,

This is my first ever contribution to the project; your feedback is much appreciated.

1c5733e3 introduced `system_time_to_universal_time/2` for backward compatibility.

The minimum OTP version is now 24; [`calendar:system_time_to_universal_time/2`](https://www.erlang.org/doc/apps/stdlib/calendar.html#system_time_to_universal_time/2) was introduced in OTP 21.

Thank you,

### DIFF
--- a/apps/arweave/src/ar_pricing.erl
+++ b/apps/arweave/src/ar_pricing.erl
@@ -573,13 +573,13 @@ get_perpetual_gb_cost(Init, Height) ->
 %% @doc Return the cost in USD of storing 1 GB per year at the given time.
 -spec get_gb_cost_per_year_at_timestamp(Timestamp::integer(), Height::nonegint()) -> usd().
 get_gb_cost_per_year_at_timestamp(Timestamp, Height) ->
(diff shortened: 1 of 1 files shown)


ariel-anieli:
> Thanks @ariel-anieli ! Since that change is in a pretty sensitive area of the code, it would be worth having a small unit test in place to confirm no change in behavior. I suspect that codepath is already covered by some of the higher order tests, but having one specifically for the time function would give me more peace of mind.
> 
> Some notes about the arweave tests:
> 
>     * We use `eunit` but the majority of our tests are not actually unit tests - most are actually integration tests, we just use the `eunit` framework and utilities for test registration and assertions.
> 
>     * you can run all the tests with `./bin/test` or limit the test run to specific module with `./bin/test module_name`

Thanks for the hints, @JamesPiechota; I'll look into them.

PR #845, opened 2025-08-13
fix: when data_sync_record gets throttled, recover gracefully

JamesPiechota:
This PR does a few things:
- adds some extra headers when we have a 429. Headers aren't used yet, but I figure we can make use of them later if we want
- fixes a sort issue in `ar_data_discover:get_bucket_peers` - previously we did a conversion to set that would lose the sorting of the peer list. And since the list was sorted based on peers with the most advertised data, it's possible the jumbled list would have had nodes focusing too much effort on low-data peers.
- exposes `ar_rate_limiter:is_throttled` so a node can check if it is currently throttling itself when querying a given peer
introduces `ar_rate_limiter:set_cooldown` this is used when we get a 429 from a peer (i.e. could be a situation where the peer's rate limit is lower than our own, or perhaps we're just hitting it with too many concurrent requests). The cooldown period isn't enforced, but nodes can query it if they want
- filters the list of peers that we query data_sync_records from to skip those that are throttled or in cooldown
removes the exception from batch_pmap and instead flags the given job as having timed out

The sum total of these changes allows the node to be more selective about which peers it queries sync records from, and to recover more gracefully when one peer times out.

With these changes the node seems to query sync records far more regularly and with greater throughput than before - and the average request time is a couple seconds instead of ~1 minute (this is because we back off of peers that are being throttled)

### DIFF
--- a/apps/arweave/src/ar_data_discovery.erl
+++ b/apps/arweave/src/ar_data_discovery.erl
@@ -55,16 +55,10 @@ get_bucket_peers(Bucket) ->
 
 get_bucket_peers(Bucket, Cursor, Peers) ->
 	case ets:next(?MODULE, Cursor) of
-		'$end_of_table' ->
-			UniquePeers = sets:to_list(sets:from_list(Peers)),
-			PickedPeers = pick_peers(UniquePeers, ?QUERY_BEST_PEERS_COUNT),
-			PickedPeers;
 		{Bucket, _Share, Peer} = Key ->
 			get_bucket_peers(Bucket, Key, [Peer | Peers]);
-		_ ->
-			UniquePeers = sets:to_list(sets:from_list(Peers)),
-			PickedPeers = pick_peers(UniquePeers, ?QUERY_BEST_PEERS_COUNT),
-			PickedPeers
+		_ -> % matches `end_of_table` or an unexpected value
+			ar_util:unique(Peers)
 	end.
 
 %% @doc Return a list of peers where 80% of the peers are randomly chosen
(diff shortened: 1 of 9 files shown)


apps/arweave/src/ar_data_discovery.erl
```diff
@@ -55,16 +55,10 @@ get_bucket_peers(Bucket) ->
...
-			PickedPeers = pick_peers(UniquePeers, ?QUERY_BEST_PEERS_COUNT),
-			PickedPeers;
 		{Bucket, _Share, Peer} = Key ->
 			get_bucket_peers(Bucket, Key, [Peer | Peers]);
-		_ ->
-			UniquePeers = sets:to_list(sets:from_list(Peers)),
-			PickedPeers = pick_peers(UniquePeers, ?QUERY_BEST_PEERS_COUNT),
-			PickedPeers
+		_ -> % matches `end_of_table` or an unexpected value
+			ar_util:unique(Peers)
```
ldmberman, line 61: In the new version we return all peers corresponding to the given buckets and not QUERY_BEST_PEERS_COUNT peers?

ldmberman: Ah, we pick them on the caller side now, I see.

ldmberman (approved):

apps/arweave/src/ar_data_discovery.erl
JamesPiechota: Yeah, by moving the pick outside the call, it lets us do additional filtering before picking.

PR #852, opened 2025-08-19
fix: fix the hasrate loss when running cm on 2.9.5 alpha

JamesPiechota:
Also fixes the cache_limit_exceeded error spam

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -342,7 +342,6 @@ jobs:
             ar_mining_io_tests,
             ar_mining_worker_tests,
             ar_poller_tests,
-            ar_reject_chunks_tests,
             ar_replica_2_9_nif_tests,
             ar_semaphore_tests,
             ar_start_from_block_tests,
--- a/apps/arweave/src/ar_mining_worker.erl
+++ b/apps/arweave/src/ar_mining_worker.erl
@@ -356,7 +356,7 @@ handle_task({computed_h0, Candidate, _ExtraArgs}, State) ->
 			%% recall range.
 			State1 = release_cache_range_space(1, Candidate2#mining_candidate.session_key, State),
 			%% Mark second recall range as missing, not to wait for it to arrive.
-			State2 = mark_second_recall_range_missing(Candidate2, State1),
+			State2 = mark_recall_range_missing(chunk2, Candidate2, State1),
 			%% Read the recall range; the result of the read will be reported by the `chunk1` task.
 			ar_mining_io:read_recall_range(chunk1, self(), Candidate2, RecallRange1Start),
 			State2;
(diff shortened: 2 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -342,7 +342,6 @@ jobs:
             ar_mining_io_tests,
             ar_mining_worker_tests,
             ar_poller_tests,
-            ar_reject_chunks_tests,
```
humaite, line 345: Is it normal to remove this test from the suite?

apps/arweave/test/ar_reject_chunks_tests.erl
```diff
@@ -300,6 +299,7 @@ test_rejects_chunks_exceeding_disk_pool_limit() ->
 		ar_test_node:post_chunk(main, ar_serialize:jsonify(FirstProof3))
 	),
 	ar_test_node:mine(peer1),
+	assert_wait_until_height(main, 1),
```
ldmberman, line 302: Why do we need to wait here?

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -525,8 +525,7 @@ handle_task({compute_h2_for_peer, Candidate, _ExtraArgs}, State) ->
 			ar_mining_stats:h1_received_from_peer(Peer, length(H1List)),
 			%% Mark second recall range as missing, just like if we didn't have the recall range
```
ldmberman: These comments are outdated, right?

.github/workflows/test.yml
JamesPiechota: It wsa a duplicate registration. Also exists in the long running tests section

apps/arweave/test/ar_reject_chunks_tests.erl
JamesPiechota: it turns out we ma not need it. I added it because the loop afterwards is checking the status of the tx on th `main` node but the block had been mined on the `peer1` node so without the wait it's possible the block doesn't get gossipped to `main` before the time out occurs.

That said, adding the wait did not ddress the core issue. I can remove it if needed.

apps/arweave/src/ar_mining_worker.erl
JamesPiechota: Updated!

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -523,15 +523,12 @@ handle_task({compute_h2_for_peer, Candidate, _ExtraArgs}, State) ->
...
 		true ->
 			ar_mining_stats:h1_received_from_peer(Peer, length(H1List)),
-			%% Mark second recall range as missing, just like if we didn't have the recall range
-			%% in a regular mining setup.
-			%% This is done because we only have a part of H1 list that passes diff checks.
-			State1 = mark_second_recall_range_missing(Candidate3, State),
-			%% After we marked the whole second recall range as missing, we can cache the H1 list.
-			%% During this process, we also reset the chunk2_missing flag to false for the entries
-			%% we have H1 for.
+			%% First we mark the whole second recall range as missing
```
ldmberman: I am confused, do we mark the second or the first recall range as missing?

JamesPiechota: Sigh. *first* recall range. I'd done the comment update before having coffee and incorrectly said "second"

Ive updated it again!

ldmberman (approved):

PR #865, opened 2025-08-28
chore: fix some minor issues in comments

zhedazijingang:
fix some minor issues in comments


JamesPiechota:
We generally don't merge commits that only address comment typos/spelling. We get too much spam from people looking to build credibility by having low-value commits merged into repos.

PR #870, opened 2025-09-29
Upgrade Jiffy to its latest version

humaite:
While compiling previous version, it fails on MacOS 26+. Instead of patching our fork of jiffy, this PR is upgrading it. Indeed, the latest version of jiffy seems to works correctly on MacOS 26+.


JamesPiechota (approved):
LGTM!

PR #878, opened 2025-10-08
fix: update v2_index_data_size from the mining_stats

JamesPiechota:
Previously we did it via global_sync_record, but sicne that is use to advertise chunks for syncing it could sometimes misrepresent the total data stored

### DIFF
--- a/apps/arweave/src/ar_mining_stats.erl
+++ b/apps/arweave/src/ar_mining_stats.erl
@@ -2,7 +2,7 @@
 -behaviour(gen_server).
 
 -export([start_link/0, start_performance_reports/0, pause_performance_reports/1, mining_paused/0,
-		set_total_data_size/1, set_storage_module_data_size/6,
+		set_storage_module_data_size/6,
 		vdf_computed/0, raw_read_rate/2, chunks_read/2, h1_computed/2, h2_computed/2,
 		h1_solution/0, h2_solution/0, block_found/0, block_mined_but_orphaned/0,
 		h1_sent_to_peer/2, h1_received_from_peer/2, h2_sent_to_peer/1, h2_received_from_peer/1,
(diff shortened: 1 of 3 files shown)


apps/arweave/src/ar_mining_stats.erl
```diff
@@ -184,17 +183,22 @@ block_mined_but_orphaned() ->
 block_mined_but_orphaned(Now) ->
 	increment_count(block_mined_but_orphaned, 1, Now).
 
-set_total_data_size(DataSize) ->
+update_total_data_size() ->
+	Pattern = {
+		{partition, '_', storage_module, '_', packing, '_'}, '$1'
+	},
+	Sizes = [Size || [Size] <- ets:match(?MODULE, Pattern)],
```
ldmberman, line 190: This has a side-effect that if a node has multiple storage modules covering the same data, everything will be added up. I think, it is fine, I suggest we just mention it in the metric description in ar_metrics.erl.

JamesPiechota: Does this work? https://github.com/ArweaveTeam/arweave/pull/878/commits/966612b0e5fe371e7d41857097cf32547619e83a

I guess the only edge case that can't be handled by `v2_index_data_size` or `v2_index_data_size_by_packing` is if they define 2 overlapping storage modules with the same packing format. In that case they won't be able to get a count of the unique bytes stored and indexed.

But I think that's a rare case? It seems the more common configurations are:
- non-overlapping modules of the same packing type
- non-overlapping modules with different packing types
- overlapping modules where each module is a different packing type

All 3 of those can be handled by `v2_index_data_size` or `v2_index_data_size_by_packing`

Maybe okay to leave the 4th case (overlapping modules where each modules has the same packing type) as an edge case to be addressed later if there's a need?

ldmberman: Makes sense, thank you!

ldmberman (approved):

PR #879, opened 2025-10-09
Create Arweave Config Application

humaite:
Create a new Erlang application called arweave_config. This one will contain all the functions/modules/processes related to arweave configuration.

ar_config.hrl header file has been moved into this new application, and renamed arweave_config.hrl. Instead of removing completely the old configuration, we integrate it in arweave_config.

arweave_config MUST BE started before arweave application.

see: https://github.com/ArweaveTeam/arweave-dev/issues/967


apps/arweave_config/src/arweave_config.erl
```diff
@@ -0,0 +1,102 @@
...
+%%--------------------------------------------------------------------
+-spec start() -> ok | {error, term()}.
+
+start() ->
+	case application:ensure_all_started(?MODULE, permanent) of
+		{ok, Dependencies} ->
+			?LOG_DEBUG("arweave_config started dependencies: ~p", Dependencies),
+			ok;
+		Elsewise ->
+			Elsewise
```
JamesPiechota, line 47: I haven't seen this before. Is it a community standard way of handling the else?

humaite: This is a "shortcut" when you don't want to recreate the wheel. We already know the matching pattern (e.g. `{ok, Variable}`), and we just want to forward other data structure to the caller. In this case, `{error, ...}` or any other datastructure, will be then managed by the caller. I don't think this is a standard, perhaps more a convention.

PR #882, opened 2025-10-15
Arweave Legacy Configuration Process Management

humaite:
This PR adds a dynamic interface to legacy configuration process. The configuration is now stored in two places:

  1. in `arweave_config` environment
  2. in `arweave_config_legacy` process state

This module introduce arweave_config_legacy, a temporary module/process
to store the legacy configuration inside a state process instead of
storing them directly inside application environment.

see: https://github.com/ArweaveTeam/arweave-dev/issues/966

Integrate Common Test Suite in Github Worflow

Integrate common test with rebar3 in the workflow. arweave_config will
use common test suite for integration testing. Due to non-reproductible
build with CMake, a workaround has been added in the suite. It will
clean up all produced objects before compiling them.

see: https://github.com/ArweaveTeam/arweave-dev/issues/979

Potential Race Condition in ar_http_iface_tests

Added parallel map support to increase the number of requests
in parallel, but it seems one of the problem is coming from
blacklist middleware implementation.

see: https://github.com/ArweaveTeam/arweave-dev/issues/982


apps/arweave_config/include/arweave_config.hrl
```diff
@@ -184,7 +184,7 @@
 	block_gossip_peers = [],
 	local_peers = [],
 	sync_from_local_peers_only = false,
-	data_dir = ".",
+	data_dir = "./data",
```
humaite, line 187: I've modified the default `data_dir` path for safety. Indeed, when executing test, a cleanup function is executed, and removes the data directory. unfortunately, when configure the wrong temporary `data_dir` path, it removes the whole git tree, leaving an empty directory. By configuring this value to something different than `./`, we are limiting the risk - but we are not removing it.

JamesPiechota (approved):
LGTM!

humaite:
> How can I run the common-test locally?
> 
> EDIT: ah maybe: `./rebar3 ct -v --logdir _ct_logs`?
> 
> EDIT2: yeah I see you describe it here: [ArweaveTeam/arweave-dev#979](https://github.com/ArweaveTeam/arweave-dev/issues/979)

Yeah, but the coverage does not work correctly, probably due to the current arweave test suite doing some odd stuff. I was trying to fix that last week, without success for now. But I don't think it's quite critical for the moment, it will be the subject of another issue.

PR #883, opened 2025-10-16
Simplify Arweave External Dependencies Compilation

humaite:
`build_randomx.sh` and `build_secp256k1.sh` have been removed and replaced with a new Makefile in `./apps/arweave/lib` directory. This will improve portability and decrease compilation time. Instead of having a long list of rebar `pre_hook`, we only have 4 for the all the dependencies.

see: https://github.com/ArweaveTeam/arweave-dev/issues/981


apps/arweave/lib/Makefile
```diff
@@ -0,0 +1,148 @@
...
+RANDOMX_CMAKE_OPTIONS_4096 ?=
+	-DUSE_HIDDEN_VISIBILITY=ON \
+	-DRANDOMX_ARGON_MEMORY=524288 \
+	-DRANDOMX_DATASET_BASE_SIZE=4294967296
+
+RANDOMX_CMAKE_OPTIONS_squared ?=
+	-DUSE_HIDDEN_VISIBILITY=ON \
+	-DRANDOMX_ARGON_MEMORY=524288 \
+	-DRANDOMX_DATASET_BASE_SIZE=2147483648 \
+	-DRANDOMX_DATASET_BASE_SIZE=2147483648 \
```
JamesPiechota:
```suggestion
```

apps/arweave/lib/Makefile
```diff
@@ -0,0 +1,148 @@
...
+	cd RandomX/build$(1) \
+		&& cmake $$(RANDOMX_CMAKE_OPTIONS_$(1)) ..
+
+RandomX/build$(1)/librandomx.a: RandomX/build$(1)/Makefile
+	cd RandomX/build$(1) \
+		&& cmake --build .
+
+RandomX/build$(1)/librandomx$(1).a: RandomX/build$(1)/librandomx.a
+	cd RandomX/build$(1) \
+		&& cp librandomx.a librandomx$(1).a
```
JamesPiechota, line 112: previous pre_hook did an `mv` rather than `cp` - not sure if relevant, just flagging

PR #884, opened 2025-10-16
Lb/data root syncing

ldmberman:


apps/arweave/src/ar_data_root_sync.erl
```diff
@@ -0,0 +1,143 @@
...
+-include("ar_sup.hrl").
+
+-record(state, {
+	store_id,
+	range_start,
+	range_end,
+	scan_cursor
+}).
+
+-define(DATA_ROOTS_SYNC_RELEASE_NUMBER, 88).
```
JamesPiechota:
```suggestion
-define(DATA_ROOTS_SYNC_RELEASE_NUMBER, 89).
```

apps/arweave/src/ar_data_root_sync.erl
```diff
@@ -0,0 +1,143 @@
...
+-define(DATA_ROOTS_SYNC_SCAN_INTERVAL_MS, 600_000). % 10 minutes.
+-endif.
+
+%%%===================================================================
+%%% Public interface.
+%%%===================================================================
+
+start_link(StoreID) ->
+	Name = list_to_atom("ar_data_root_sync_" ++ ar_storage_module:label(StoreID)),
+	gen_server:start_link({local, Name}, ?MODULE, [StoreID], []).
```
JamesPiechota:
```suggestion
	Name = name(StoreID),
	gen_server:start_link({local, Name}, ?MODULE, [StoreID], []).
	
name(StoreID) ->
	list_to_atom("ar_data_root_sync_" ++ ar_storage_module:label(StoreID)).
```

apps/arweave/src/ar_data_root_sync.erl
```diff
@@ -0,0 +1,143 @@
+-module(ar_data_root_sync).
+
+-behaviour(gen_server).
+
+-export([start_link/1]).
```
JamesPiechota:
```suggestion
-export([start_link/1, name/1]).
```

apps/arweave/src/ar_data_root_sync_sup.erl
```diff
@@ -0,0 +1,38 @@
...
+init([]) ->
+	Workers = register_workers(),
+	{ok, {{one_for_one, 5, 10}, Workers}}.
+
+register_workers() ->
+	{ok, Config} = application:get_env(arweave, config),
+	lists:map(
+		fun(StorageModule) ->
+			StoreID = ar_storage_module:id(StorageModule),
+			Name = list_to_atom("ar_data_root_sync_" ++ ar_storage_module:label(StoreID)),
```
JamesPiechota:
```suggestion
			Name = ar_data_root_sync:name(StoreID),
```

apps/arweave/src/ar_data_root_sync.erl
```diff
@@ -0,0 +1,143 @@
...
+sync_block_data_roots(_StoreID, Cursor, RangeEnd) when Cursor >= RangeEnd ->
+	{ok, Cursor};
+sync_block_data_roots(StoreID, Cursor, RangeEnd) ->
+	{BlockStart, BlockEnd, _} = ar_block_index:get_block_bounds(Cursor),
+	Cursor2 =
+		case BlockStart >= RangeEnd of
+			true ->
+				RangeEnd;
+			false ->
+				case ar_sync_record:is_recorded(Cursor + 1, data_roots, ?DEFAULT_MODULE) of
```
JamesPiechota, line 105: Are we using a `data_roots` sync record in the `default` module so that we can track which chunk ranges we have data_roots for?  i.e. We don't store the data_roots directly in the `data_roots_index` rocksdb because it makes it hard to know which chunk ranges we have data_roots for and which we don't?

apps/arweave/src/ar_data_root_sync_sup.erl
```diff
@@ -0,0 +1,38 @@
...
+			?CHILD_WITH_ARGS(ar_data_root_sync, worker, Name, [StoreID])
+		end,
+		Config#config.storage_modules
```
JamesPiechota, line 38: Why multiple data_root_sync workers rather than just 1? (since they'll all be storing the data to the `default` module anyways)

Is it primarily for performance reason (e.g. to allow us to sync data_roots for each storage module in paralllel)

apps/arweave/src/ar_data_sync.erl
```diff
@@ -877,6 +890,25 @@ handle_cast({move_data_root_index, Cursor, N}, State) ->
 	move_data_root_index(Cursor, N, State),
 	{noreply, State};
 
+handle_cast({store_data_roots, BlockStart, BlockEnd, TXRoot, Entries}, State) ->
+	#sync_data_state{ store_id = ?DEFAULT_MODULE } = State,
```
JamesPiechota, line 895: Should this be part of the `handle_cast` match to avoid a `batmatch`? Or maybe we want a catastrophic `badmatch` because this situation should never occur?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -940,7 +972,8 @@ handle_cast({cut, Start}, #sync_data_state{ store_id = StoreID,
 				true ->
 					ok = delete_chunk_metadata_range(Start, End, State),
 					ok = ar_chunk_storage:cut(Start, StoreID),
-					ok = ar_sync_record:cut(Start, ar_data_sync, StoreID)
+					ok = ar_sync_record:cut(Start, ar_data_sync, StoreID),
+					ok = ar_sync_record:cut(Start, data_roots, ?DEFAULT_MODULE)
```
JamesPiechota: Do we need to also remove from the `data_root_offset_index` `ar_kv`?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -3958,6 +3994,49 @@ record_chunk_cache_size_metric() ->
...
+%% Return only entries corresponding to non-empty transactions.
+%% Return the complete list of entries in the order they appear in the data root index,
+%% which corresponds to sorted #tx records in the block.
+%% Return {ok, {TXRoot, BlockSize, [{DataRoot, TXSize, TXStartOffset, TXPath}, ...]}}
+%% or {error, Reason}.
+get_data_roots_for_offset(Offset) ->
+	{BlockStart, BlockEnd, TXRoot} = ar_block_index:get_block_bounds(Offset),
+	true = Offset >= BlockStart andalso Offset < BlockEnd,
+	StoreID = ?DEFAULT_MODULE,
+	DB = {data_root_offset_index, StoreID},
```
JamesPiechota, line 4002: This db maps from a byte offset to a block containing that byte to a set of keys that can be used with the `data_root_index` to query the actual data_roots - is that right?

apps/arweave/src/ar_header_sync.erl
```diff
@@ -98,39 +98,45 @@ handle_cast({join, Height, RecentBI, Blocks}, State) ->
...
-		lists:foldl(
-			fun(B, S) ->
-				element(2, add_block(B, S))
-			end,
-			State3,
-			Blocks
-		),
-	ok = store_sync_state(State4),
-	{noreply, State4};
+	case Status of
```
JamesPiechota, line 125: Oh interesting - what happened during an error condition before?

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -778,6 +778,139 @@ post_cm_partition_table_to_pool(Peer, Payload) ->
...
+			timeout => 10 * 1000,
+			connect_timeout => 2000,
+			headers => p2p_headers()
+		}),
+	handle_get_data_roots_response(Response, Offset).
+
+handle_get_data_roots_response({ok, {{<<"200">>, _}, _, Body, _, _}}, Offset) ->
+	case ar_serialize:binary_to_data_roots(Body) of
+		{ok, {TXRoot, BlockSize, Entries}} ->
+			handle_get_data_root_response(TXRoot, BlockSize, Entries, Offset);
```
JamesPiechota: Why are is one function plural `data_roots` and one singular `data_root`? It looks like in all cases we are returning multiple data roots - or is it we're always returning one? haha I'm having a little trouble parsing through... In either case it seems like both functions should be singular or both plural?

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -778,6 +778,139 @@ post_cm_partition_table_to_pool(Peer, Payload) ->
...
+
+prepare_data_root_pairs(Entries, BlockStart, BlockSize) ->
+	Result = lists:foldr(
+		fun
+			(_, {error, _} = Error) ->
+				Error;
+			({_DataRoot, 0, _TXStartOffset, _TXPath}, _Acc) ->
+				{error, invalid_zero_tx_size};
+			({DataRoot, TXSize, TXStartOffset, TXPath}, {ok, {Total, Acc}}) ->
+				Label = TXStartOffset + TXSize - BlockStart,
```
JamesPiechota: What does `Label` mean in this case? Is it the block-relative offset of the end of the tx?

apps/arweave/src/ar_http_iface_client.erl
```diff
@@ -778,6 +778,139 @@ post_cm_partition_table_to_pool(Peer, Payload) ->
...
+					{error, invalid_total_tx_size}
+			end;
+		Error6 ->
+			Error6
+	end.
+
+get_padded_size(TXSize, BlockStart) ->
+	case BlockStart >= ar_block:strict_data_split_threshold() of
+		true ->
+			ar_poa:get_padded_offset(TXSize, 0);
```
JamesPiechota, line 894: Intersting this implementation differs a little from the one in `ar_block` and `ar_chunk_storage` - why `0` here instead of `ar_block:strict_data_split_threshold()`?

Oh and similarly why `BlockStart >=` vs. `BlockStart >`?

apps/arweave/src/ar_data_sync.erl
ldmberman: Yes, exactly, AbsoluteBlockStartOffset => {TXRoot, BlockSize, DataRootIndexKeySet}

apps/arweave/src/ar_header_sync.erl
ldmberman: The node crashed. It is a not very realistic condition though - when we start from a deep fork. I noticed it thanks to the initially wrong test setup.

apps/arweave/src/ar_data_sync.erl
ldmberman: We already do it in remove_orphaned_data/3

apps/arweave/src/ar_data_sync.erl
ldmberman: Yes, it should never occur.

apps/arweave/src/ar_data_root_sync_sup.erl
ldmberman: I think, the implementation is cleaner this way. It may also speed syncing up indeed.

apps/arweave/src/ar_data_root_sync.erl
ldmberman: Hm, I am reconsidering it now. Probably makes sense to just use data_root_offset_index instead. It will cost 1 db lookup/block compared with just memory lookup (at the expense of some RAM overhead) which I think is fine.

apps/arweave/src/ar_http_iface_client.erl
ldmberman: We are padding the size here (every tx is padded individually), not the absolute offset.

The first block following the strict rules is the one starting at the threshold. Why do you consider `BlockStart >`?

apps/arweave/src/ar_http_iface_client.erl
ldmberman: Fixed here - https://github.com/ArweaveTeam/arweave/pull/884/commits/e9bbc4f51720a4fda8f6bc4ffb2c58e2519f3b2e

apps/arweave/src/ar_http_iface_client.erl
ldmberman: Originally this is how we call offsets of merkle tree nodes. Renamed into MerkleLabel here - https://github.com/ArweaveTeam/arweave/pull/884/commits/e9bbc4f51720a4fda8f6bc4ffb2c58e2519f3b2e

apps/arweave/src/ar_http_iface_client.erl
JamesPiechota: > Why do you consider BlockStart >?

Mostly just pattern matching 😂 I noticed a few places use `>`
- https://github.com/ArweaveTeam/arweave/blob/lb/sync-replica-2-9/apps/arweave/src/ar_block.erl#L687
- https://github.com/ArweaveTeam/arweave/blob/lb/sync-replica-2-9/apps/arweave/src/ar_chunk_storage.erl#L100
- https://github.com/ArweaveTeam/arweave/blob/lb/sync-replica-2-9/apps/arweave/src/ar_chunk_storage.erl#L321
- https://github.com/ArweaveTeam/arweave/blob/lb/sync-replica-2-9/apps/arweave/src/ar_data_sync.erl#L746
- https://github.com/ArweaveTeam/arweave/blob/lb/sync-replica-2-9/apps/arweave/src/ar_data_sync.erl#L3442

But the use cases are slightly different, so may be irrelevant.

ldmberman: Ah, right, those are chunk's offsets >start offset up to and including the end offset.

PR #885, opened 2025-10-20
Integrate New Parameter Format

humaite:
This PR is adding support for the new parameter format, introducing as well the new methods to dynamically configure Arweave during runtime. This PR does not break the legacy configure (yet), and offer a compatible interface between both configuration method. Instead of redefining a parameter in many different places, all of them are defined at the same place. A small example of a parameter definition can be seen below.

```erlang
#{
	parameter_key => [data,directory],
	default => "./data",
	runtime => false,
	type => path,
	deprecated => false,
	legacy => data_dir,
	required => true,
	short_description => "",
	long_description => "",
	environment => <<"AR_DATA_DIRECTORY">>,
	short_argument => $D,
	long_argument => <<"--data-directory">>,
	handle_get => fun legacy_get/2,
	handle_set => fun legacy_set/3
}.
```

Only 2 parameters have been implemented for the moment, `[debug]` (legacy: `debug`) and `[data,directory]` (legacy: `data_dir`). The first one can be configured during runtime, the last one cannot. 

The only way to configure these values using the new method is to user `arweave_config:set/2`.

```erlang
arweave_config:set([data,directory], "./data").
% or
arweave_config:set(<<"data.directory">>, "./data").
```

When all parameters have been configured, and just before starting arweave, `arweave_config:runtime/0` function must be called to alert `arweave_config` the node is now ready to be started, and to protect non-runtime parameters.

Support for arguments, environments and Web API will be created in another PR.

Note: `arweave_config` is currently checking parameters during runtime, and not during compile time. A `parse_transform` function will be added after the validation of the new configuration format. For now, it's easier to have this kind of configuration.

see: https://github.com/ArweaveTeam/arweave-dev/issues/977

### DIFF
--- a/apps/arweave_config/src/arweave_config.erl
+++ b/apps/arweave_config/src/arweave_config.erl
@@ -1,4 +1,10 @@
 %%%===================================================================
+%%% GNU General Public License, version 2 (GPL-2.0)
+%%% The GNU General Public License (GPL-2.0)
+%%% Version 2, June 1991
+%%%
+%%% ------------------------------------------------------------------
+%%%
 %%% @copyright 2025 (c) Arweave
 %%% @author Arweave Team
 %%% @author Mathieu Kerjouan
@@ -9,26 +15,29 @@
 %%% stored and specified.
 %%%
 %%% WARNING: this module/application is in active development.
-%%%
 %%% @end
-%%%
-%%% ------------------------------------------------------------------
-%%%
-%%% GNU General Public License, version 2 (GPL-2.0)
-%%% The GNU General Public License (GPL-2.0)
-%%% Version 2, June 1991
 %%%===================================================================
 -module(arweave_config).
 -vsn(1).
 -behavior(application).
+-behavior(gen_server).
 -export([
-	start/0,
-	stop/0,
+	get/1,
+	get/2,
 	get_env/0,
-	set_env/1
+	is_runtime/0,
+	runtime/0,
+	set/2,
+	set_env/1,
+	start/0,
+	start_link/0,
+	stop/0
 ]).
 % application behavior callbacks.
 -export([start/2, stop/1]).
+% gen_server behavior callbacks
+-export([init/1, terminate/2, handle_call/3, handle_cast/2, handle_info/2]).
+-compile({no_auto_import,[get/1]}).
 -include("arweave_config.hrl").
 -include_lib("kernel/include/logger.hrl").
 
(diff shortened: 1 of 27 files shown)


apps/arweave_config/src/arweave_config.erl
```diff
@@ -80,6 +89,214 @@ get_env() ->
...
+
+%%--------------------------------------------------------------------
+%% @doc Start arweave_config process.
+%% @end
+%%--------------------------------------------------------------------
+start_link() ->
+	gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+%%--------------------------------------------------------------------
+%% @doc Switch to runtime mode. No rollback is possible there, this is
```
JamesPiechota, line 229: Prior to calling `runtime` the assumption is we're still loading the static config and then after calling `runtime` **only** dynamic config values are allowed to change? Have I understood correctly?

apps/arweave_config/src/arweave_config.erl
```diff
@@ -80,6 +89,214 @@ get_env() ->
...
+get(Key) ->
+	case arweave_config_parser:key(Key) of
+		{ok, Parameter} ->
+			case arweave_config_store:get(Parameter) of
+				{ok, Value} ->
+					{ok, Value};
+				_Elsewise ->
+					arweave_config_spec:get_default(Parameter)
+			end;
+		Elsewise ->
```
JamesPiechota, line 140: haha, I'm guessing `Elsewise` is an Erlang community standard? Never heard the word before!

apps/arweave_config/src/arweave_config_parameters.erl
```diff
@@ -0,0 +1,84 @@
...
+					legacy_set(K, V, S)
+			end
+		}
+	].
+
+legacy_get(_K, #{ spec := #{ legacy := L }}) ->
+	V = arweave_config_legacy:get(L),
+	{ok, V}.
+
+legacy_set(_K, V, #{ spec := #{ legacy := L }}) ->
```
JamesPiechota, line 82: do we need to protect this with a check for `runtime == true`? (I'm still going through the logic, forgive a naive question)

apps/arweave_config/src/arweave_config_parser.erl
```diff
@@ -0,0 +1,327 @@
...
+%%
+%%--------------------------------------------------------------------
+is_parameter_list([]) -> true;
+is_parameter_list([<<>>|_]) -> false;
+is_parameter_list([Item|Rest]) when is_atom(Item) ->
+	is_parameter_list(Rest);
+is_parameter_list([Item|Rest]) when is_integer(Item) ->
+	is_parameter_list(Rest);
+is_parameter_list([Item|Rest]) when is_binary(Item) ->
+	is_parameter_list(Rest);
```
JamesPiechota, line 99: What's your thinking on this pattern vs. `is_atom(Item); is_integer(Item); is_binary(Item)`

I don't have a strong opinion, just trying to understand the context as we gradually align on a coding style

apps/arweave_config/src/arweave_config_parser.erl
```diff
@@ -0,0 +1,327 @@
...
+	{error, #{
+			rest => Rest,
+			buffer => Buffer,
+			reason => bad_key
+		 }
+	}.
+
+key_convert_to_atom(Item, Rest, Buffer) ->
+	try
+		Atom = binary_to_existing_atom(Item),
```
JamesPiechota, line 210: neat, and we're guaranteed that an atom already exists for all valid configurations because we defined them in `arweave_config_parameters`?

apps/arweave_config/src/arweave_config_spec.erl
```diff
@@ -0,0 +1,1064 @@
...
+% REQUIRED:  defines the  configuration key  used to  identify arweave
+% parameter, usually stored in a data store like ETS.
+%---------------------------------------------------------------------
+-callback parameter_key() -> Return when
+	Return :: parameter().
+
+%---------------------------------------------------------------------
+% OPTIONAL: defines how to retrieve the value using the parameter key.
+%---------------------------------------------------------------------
+-callback handle_get(ParameterKey, State) -> Return when
```
JamesPiechota, line 144: callbacks are a new concept to me. It looks like this module defines `arweave_config_spec` and lists all the expected functions within the module. Those funcions are then defined within the `arweave_config_spec_xxx.erl` modules? 

What i don't quite see if we have a `handle_get` callback defined here, but then within `arweave_config_spec_handle_get.erl` all we have is an `init` - where is `handle_get` actually defined? (and ditto for the other callbacks)

JamesPiechota: I'm guessing somewhere we have a module with a `-behavior(arweave_config_spec)` line? and *it* implements the `handle_get`, `handle_set` etc...?

JamesPiechota: Ah wait, I guess it's all in the maps defined in `arweave_config_parameters`? Somewhere there's some magic to convert those maps to behaviors that implement `arweave_config_spec`?

JamesPiechota: Ah okay that "magic" down lower in this file

apps/arweave_config/src/arweave_config_parameters.erl
JamesPiechota: Ah I see, it gets checked here: https://github.com/ArweaveTeam/arweave/pull/885/files#diff-4752b02a7ebf9e1972017fcd8df4cfca009c009356e0d43f28853ed6555e1ed1R891

apps/arweave_config/src/arweave_config_spec.erl
```diff
@@ -0,0 +1,1064 @@
...
+		Elsewise ->
+			Elsewise
+	end.
+
+%%--------------------------------------------------------------------
+%% @hidden
+%% 3. let retrieve the value (if set) and use the handle_set/3
+%% callback to set the value.
+%%--------------------------------------------------------------------
+apply_set_value(Parameter, Value, Spec = #{ set := Set }) ->
```
JamesPiechota, line 927:
```suggestion
apply_set_value(Parameter, Value, Spec = #{ handle_set := Set }) ->
```

Should this be `handle_set` instead of `set`? Or have I misunderstood something

JamesPiechota: Ah I see it gets renamed to `set` here: https://github.com/ArweaveTeam/arweave/pull/885/files#diff-c0dd7572e35573eb43aeed0c2a9024756c90f985dc3625910403175f8ff3d753R21

JamesPiechota (approved):
LGTM! I added a bunch of comments, but they are all mostly trying to confirm my understanding. This PR is big enough that our best bet is probably to merge it and continue the testing/reviewing live over time (I don't trust myself to fully understand all the ins and outs - but from what I read it looks really good)

apps/arweave_config/src/arweave_config.erl
humaite: correct. The goal is to create a clear separation between the startup phase and the runtime phase. During the startup phase, the parameters are being set using environment (static values), arguments (static values) and configuration file (static values). When everything has been correctly configured, and we think it's good to start arweave, we call `arweave_config:runtime/0` and then, only dynamic parameters will be allowed to be configured. By default, if not explicitly configured on a parameter, all parameters are not allowed to be configured during runtime, but only during startup.

apps/arweave_config/src/arweave_config.erl
humaite: Yeah... You can find it on all Erlang books :')

apps/arweave_config/src/arweave_config_parameters.erl
humaite: when a parameter is set, a pipeline of functions are executed, one of them is to check if the node is in runtime mode or node and if the parameter is allowed to be configured in this mode. So, it is transparent. The `legacy_set` function has been created for tests, I will remove it later to make things easier to configure.

apps/arweave_config/src/arweave_config_parser.erl
humaite: true, less code, but I have the bad habit to create many functions using pattern matching.

apps/arweave_config/src/arweave_config_parser.erl
humaite: I think so. Most of the atoms are generally created during compilation time, so, it should do the job. I can't see a way to create a new atom on the fly, except if someone is connected to the node and creates them manually.

apps/arweave_config/src/arweave_config_spec.erl
humaite: The idea behind behaviors and callback is to have a way to deal with returned values instead of pipeline of functions. In this case, it's easier to create tests because we can easily control the input and we already practically know the output. A callback can be any exported functions, but with a known specification (not mandatory, because not enforced in Erlang, it only acts as a developer contract, not an application contract, except if we enable dialyzer, but even with it, there are no guarantees).

`arweave_config_spec` is defining its own behavior, it gives the opportunity to create a complete module with specific callbacks. It will be useful when we will have complex parameter to set, instead to create a map, a totally isolated module will contain all the logic. On other hands, to make things easier, a list of `map()` can also be passed. Each map should have the same keys than the behavior.

The `-behavior(X).` preprocessor flag is only there to alert the compiler. When the compiler sees this flag, it will look in the module called `X` to see if there is some callback definition there. If it's the case, and if the callback is not optional, it will check on the compiled module if the function is present and respect the contract.

A callback does not need to be always defined, this is usually a way to alert the developers we will use the module for some specific behavior. For example, when creating cowboy REST api or when we when to create a module to deal with an end-point, cowboy does not enforce that, and only requires to define a function called `init/2`. No needs to enforce a behavior there.

> Ah okay that "magic" down lower in this file

In `apps/arweave_config/src/arweave_config_specs` directory you will find all the features enabled in arweave specification. Each files contain a function called `init/2` in charge of checking if the module passed is (1) a map (2) an atom. If it's a map, it will look for a specific key. If it's an atom, it will assume this is a module, and then will look in the module if the required functions are defined (or not).

apps/arweave_config/src/arweave_config_spec.erl
humaite: Yes, `handle_set` and `handle_get` have been called like that to follow OTP convention (e.g. `gen_server:handle_cast`, `gen_server:handle_call`...) but when these functions are valid, they are renamed `set` and `get`. See below the ETS table containing the final structure used to check the parameters.

```erlang
ets:tab2list(arweave_config_spec).
[{[debug],
  #{default => false,runtime => true,
    set => #Fun<arweave_config_parameters.2.129821681>,
    type => boolean,
    get => #Fun<arweave_config_parameters.0.129821681>,
    deprecated => false,
    parameter_key => [debug],
    short_argument => 100,long_argument => <<"--debug">>,
    environment => <<"AR_DEBUG">>,legacy => debug,
    short_description => [],long_description => []}},
 {[data,directory],
  #{default => "./data",runtime => false,
    set => #Fun<arweave_config_parameters.1.129821681>,
    type => path,
    get => #Fun<arweave_config_parameters.0.129821681>,
    deprecated => false,
    parameter_key => [data,directory],
    short_argument => 68,
    long_argument => <<"--data-directory">>,
    environment => <<"AR_DATA_DIRECTORY">>,legacy => data_dir,
    short_description => [],long_description => []}}]
```

as you can see, `set` and `get` are pointing to a lambda function.

humaite:
> LGTM! I added a bunch of comments, but they are all mostly trying to confirm my understanding. This PR is big enough that our best bet is probably to merge it and continue the testing/reviewing live over time (I don't trust myself to fully understand all the ins and outs - but from what I read it looks really good)

Yes... This is a huge addition. I already tried to create lot of tests, but it is not done yet. The PoC did not have any tests, and it was a bit messy. Here, I'm trying to fix that. Furthermore, the code is not "enabled" yet, this is mostly an alpha feature, the legacy part is still being used, and the dynamic configuration will be fully enabled ONLY when we are sure most of the features are working correctly.

PR #887, opened 2025-10-30
Arweave Config: Add environment support

humaite:
This commit adds environment support to arweave config. It means all parameters including a named environment variable can now be set directly from OS environment.

see: https://github.com/ArweaveTeam/arweave-dev/issues/970


JamesPiechota (approved):
LGTM!

PR #888, opened 2025-10-31
Integrate HTTP API Interface for Arweave Config

humaite:
This commit integrate cowboy as part of arweave config to give access to
an HTTP API interface and dynamically configure parameters while arweave
is running.

The process in chage of this interface is not started automatically for
the moment, but can be started using two methods: (1) in "standalone"
mode using arweave_config_http_server:start_link() or (2) as child using
arweave_config_http_server:start_as_child(). The second method will start the
process under arweave_config_sup supervisor.

The API is using v0 prefix (this is not a stable interface for the
moment). Here the list of supported end-points:

  - /v0/config: export the whole configuration with parameter set

  - /v0/config/${parameter}: display a parameter value

  - /v0/environment: display the environment used by arweave

This implementation is using JSEND JSON format, instead of recreating
the wheel, JSEND format is easy and flexible enough for the private API
end-points.

see: https://github.com/ArweaveTeam/arweave-dev/issues/978

### DIFF
--- a/apps/arweave_config/src/arweave_config_http_server.erl
+++ b/apps/arweave_config/src/arweave_config_http_server.erl
@@ -1,0 +1,406 @@
+%%%===================================================================
+%%% GNU General Public License, version 2 (GPL-2.0)
+%%% The GNU General Public License (GPL-2.0)
+%%% Version 2, June 1991
+%%%
+%%% ------------------------------------------------------------------
+%%%
+%%% @copyright 2025 (c) Arweave
+%%% @author Arweave Team
+%%% @author Mathieu Kerjouan
+%%% @doc Configuration HTTP Server Interface.
+%%%
+%%% This module is using cowboy to handle configuration requests. The
+%%% goal is to configure dynamically some values using a web
+%%% interface. The default TCP port used is `4891'.
+%%%
+%%% The interface is using a RESTful like module, where a path is
+%%% representing an object.
+%%%
+%%% A configuration path is converted to a parameter:
+%%%
+%%% ```
+%%% v1/config/debug
+%%%
+%%% % becomes
+%%%
+%%% [debug]
+%%% '''
+%%%
+%%% This API is also versionned, the `v0' version is mostly a draft to
+%%% see how the different methods are behaving.
+%%%
+%%% JSON data returned try to follow jsend format.
+%%%
+%%% see: https://github.com/omniti-labs/jsend
+%%%
+%%% == Examples ==
+%%%
+%%% By default, the values being used and returned are raw:
+%%%
+%%% ```
+%%% # get the value of global.debug parameter
+%%% $ curl localhost:4891/v1/config/debug
+%%% {"status":"success","data":true}
+%%%
+%%% # set the value of global.debug parameter
+%%% $ curl localhost:4891/v1/config/global/debug -d false
+%%% {"status":"success","data":false}
+%%% '''
+%%%
+%%% === Unix Socket Support ===
+%%%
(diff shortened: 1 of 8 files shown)


apps/arweave_config/src/arweave_config_http_server.erl
```diff
@@ -0,0 +1,347 @@
...
+%%--------------------------------------------------------------------
+%% @hidden
+%%--------------------------------------------------------------------
+% @todo returns API specifications
+% init(Req = #{ path := <<"/v0">> }, State) ->
+% 	Reply = cowboy_req:reply(200, #{}, <<>>, Req),
+% 	{ok, Reply, State};
+% @todo returns arweave and beam arguments
+% init(Req = #{ path := <<"/v0/arguments">> }, State) -> ok;
+init(Req = #{ path := <<"/v0/environment">> }, State) ->
```
JamesPiechota: For now we're just treating everything as a GET, right? (i.e. if someone POSTs to these endpoints we'll just treat it as a GET and return the current value(s))

apps/arweave_config/src/arweave_config_http_server.erl
```diff
@@ -0,0 +1,347 @@
...
+		status => success,
+		data => Config
+	}),
+	Reply = cowboy_req:reply(200, Headers, Body, Req),
+	{ok, Reply, State};
+init(Req = #{ path := <<"/v0/config/">> }, State) ->
+	init(Req#{ path => <<"/v0/config">> }, State);
+init(Req = #{ path := <<"/v0/config/", Key/binary>> }, State)
+	when Key =/= <<>> ->
+		apply_config(Key, Req, State);
```
JamesPiechota, line 210: Ah I see, this endpoint does distinguish GET and POST, just the other ones that don't

apps/arweave_config/src/arweave_config_http_server.erl
```diff
@@ -0,0 +1,347 @@
...
+	Key3 = case Key2 of
+		_ when is_list(Key2) -> list_to_binary(Key2);
+		_ when is_binary(Key2) -> Key2
+	end,
+	case arweave_config_parser:key(Key3) of
+		{ok, Parameter} ->
+			config1(Parameter, Req, State);
+		_ ->
+			NewState = #{
+				status => 400,
```
JamesPiechota, line 271: Can this be a 404?

JamesPiechota: ah wait I see, this is just parsing the Key, the lookup happens in config1

apps/arweave_config/src/arweave_config_http_server.erl
humaite: The environment end-point returns a snapshot of the environment variables available during the startup. This is a read-only value, I don't think arweave will ever modify its environment, so, no need to support POST. So, yeah, even with a POST, I will fix that.

apps/arweave_config/src/arweave_config_http_server.erl
humaite: correct

apps/arweave_config/src/arweave_config_http_server.erl
humaite: yes, if the parameter does not exist or is incorrect, we return a 400, this is mostly an issue on the client side, not on the server side.

apps/arweave_config/src/arweave_config_http_server.erl
humaite: fixed

PR #920, opened 2025-11-24
Integrate HTTP Server Logging

humaite:
This commit adds support for http server logging. This logging
method is not enabled by default. ar_logger module has been rewritten
to be more compatible with the new arweave dynamic configuration
module.

A full refacto has been done on ar_logger module to be compatible with
arweave_config. The idea is to generate logger configuration directly
from ar_logger module, with a long list of parameters coming from
arweave_config. To start/stop an handler:

    ar_logger:start(arweave_info).
    ar_logger:start(arweave_debug).
    ar_logger:start(arweave_http_api).

    ar_logger:stop(arweave_http_api).
    ar_logger:stop(arweave_debug).
    ar_logger:stop(arweave_info).

arweave_http_api has a custom template logging, using mostly metadata
from cowboy response and producing a logfmt like output. Here an example:

    2025-12-02T14:19:25.637013Z ip=1.2.3.4 port=48540 version=HTTP/1.1
        method=GET code=200 path=/info body_length=0 duration=1 msg=

Where "ip" and "port" are from the remote peer, "version" is the http
protocol version used, "method" is the HTTP method used, "code" is the
returned code to the peer, "path" is the path requested by the remote
peer, "body_length" is the size of the body if defined and "duration"
is the time to produce the answer in millisecond.

ar_logger:handlers/0 defines all the available handlers supported by
arweave with all their options. Filters have been created to decrease
the number of lines in debug (only storing debug level without other
levels).

see: https://github.com/ArweaveTeam/arweave-dev/issues/1006

### DIFF
--- a/apps/arweave/src/ar_http_iface_middleware.erl
+++ b/apps/arweave/src/ar_http_iface_middleware.erl
@@ -33,7 +33,12 @@
 execute(Req, #{ handler := ar_http_iface_handler }) ->
 	Pid = self(),
 	HandlerPid = spawn_link(fun() ->
-		Pid ! {handled, handle(Req, Pid)}
+		{Duration, Response = {Code, _, _, Resp}}
+			= timer:tc(fun() ->
+					handle(Req, Pid)
+				end),
+		log(Code, Resp, #{duration => Duration}),
+		Pid ! {handled, Response}
 	end),
 	{ok, TimeoutRef} = ar_timer:send_after(
 		?HANDLER_TIMEOUT,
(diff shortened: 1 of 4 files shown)


apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -33,7 +33,12 @@
 execute(Req, #{ handler := ar_http_iface_handler }) ->
 	Pid = self(),
 	HandlerPid = spawn_link(fun() ->
-		Pid ! {handled, handle(Req, Pid)}
+		T0 = erlang:system_time(),
+		Response = {Code, _, _, Resp} = handle(Req, Pid),
+		T1 = erlang:system_time(),
+		Duration = (T1-T0)/1000/1000,
```
khetzl: What's the reason behind calculating this difference manually, rather than using timer:tc?

humaite: true, I think its cleaner to use `timer:tc/1` instead of using `erlang:system_time/0`. Old habit, I will fix that.

khetzl (approved):
LGTM; one non-blocker, at your discretion

apps/arweave/src/ar_logger.erl
```diff
@@ -1,95 +1,333 @@
...
+
+	% this handler will log only log message containing the domain
+	% [arweave,http,api] in info level.
+	arweave_http_api => #{
+		level => info,
+		config => #{
+			type => file,
+			file => logfile_path(#{
+				prefix => "arweave-http-api",
+				level => debug
```
JamesPiechota, line 202: should this be `level => info`?

apps/arweave/src/ar_http_iface_middleware.erl
```diff
@@ -45,6 +50,76 @@ execute(Req, #{ handler := ar_http_iface_handler }) ->
 execute(Req, Env) ->
 	{ok, Req, Env}.
 
+%%--------------------------------------------------------------------
+%% @doc Logs client requests. HTTP logging is done only if
+%% arweave_http_api handler is started.
+%% @end
+%%--------------------------------------------------------------------
+log(Code, Resp, Init) ->
+	case ar_logger:is_started(arweave_http_api) of
```
JamesPiechota, line 59: Is the idea to make this configurable sometime in the future? (e.g. like `debug`)

apps/arweave/src/ar_logger.erl
humaite: This is a specific handler, only the logs coming from `[arweave,http,api]` domains are logged. So, the level here is not really important. I set it to `debug` to open a door for more logging feature in the future.

apps/arweave/src/ar_http_iface_middleware.erl
humaite: correct! If the `arweave_http_api` handler is not started (not found in the logger application), then, nothing is logged. By default, this handler is not configured, and at this time, the only way to do that is to connect to arweave via the remote console or using RPC.

```erlang
ar_logger:start_handler(arweave_http_api).
```

Another PR (https://github.com/ArweaveTeam/arweave/pull/923) is in preparation to let users starting/stopping both `debug` and `arweave_http_api` logger during runtime.

JamesPiechota (approved):
LGTM!

PR #923, opened 2025-12-02
Integrate HTTP Logging in Arweave Config

humaite:
Lot of new parameters have been added in arweave_config to configure
logger values. Those are not enabled yet, but have been created to test
logging integration in arweave_config..

Add a new arweave_config_spec callback called enabled to enable or not a
specific parameter.

Add a new arweave_config spec callback called inherit to import
specification keys from other parameters. The goal is to avoid
duplicated code in the parameter and limit the amount of lines.

Improve handle_set callback. this function is now taking 4 parameters,
the last one can be ignored most of the time, but it adds a way to set
custom arguments during call.

Full integration of info logger (arweave_info).

Full integration of debug logger (arweave_debug).

Full integration of http api logger (arweave_http_api).

Fixed many small bugs.

see: https://github.com/ArweaveTeam/arweave-dev/issues/1006

### DIFF
--- a/apps/arweave/src/ar_logger.erl
+++ b/apps/arweave/src/ar_logger.erl
@@ -110,35 +110,87 @@ handlers() -> #{
 	% This handler can be configured on demand, all message
 	% greater or equal than info are being logged.
 	arweave_info => #{
-		level => debug,
+		level => info,
 		config => #{
 			type => file,
 			file => logfile_path(#{
 				prefix => "arweave",
 				level => info
 			}),
-			max_no_files => 10,
-			max_no_bytes => 51_418_800,
+			compress_on_rotate => arweave_config:get(
+				[logging,compress_on_rotate],
+				false
+			),
+			max_no_files => arweave_config:get(
+				[logging,sync_mode_qlen],
+				10
+			),
+			max_no_bytes => arweave_config:get(
+				[logging,max_no_bytes],
+				51_418_800
+			),
 			modes => [raw, append],
-			sync_mode_qlen => 10,
-			drop_mode_qlen => 200,
-			flush_qlen => 1000,
-			burst_limit_enable => true,
-			burst_limit_max_count => 500,
-			burst_limit_window_time => 1000,
-			overload_kill_enable => true,
-			overload_kill_qlen => 20_000,
-			overload_kill_mem_size => 3_000_000,
-			overload_kill_restart_after => 5000
+			sync_mode_qlen => arweave_config:get(
+				[logging,sync_mode_qlen],
+				10
+			),
+			drop_mode_qlen => arweave_config:get(
+				[logging,drop_mode_qlen],
+				200
+			),
+			flush_qlen => arweave_config:get(
+				[logging,flush_qlen],
+				1000
+			),
+			burst_limit_enable => arweave_config:get(
+				[logging,burst_limit_enable],
+				true
+			),
+			burst_limit_max_count => arweave_config:get(
+				[logging,burst_limit_max_count],
+				500
+			),
+			burst_limit_window_time => arweave_config:get(
+				[logging,burst_limit_window_time],
+				1000
+			),
+			overload_kill_enable => arweave_config:get(
+				[logging,overload_kill_enable],
+				true
+			),
+			overload_kill_qlen => arweave_config:get(
+				[logging,overload_kill_qlen],
+				20_000
+			),
+			overload_kill_mem_size => arweave_config:get(
+				[logging,overload_kill_mem_size],
+				3_000_000
+			),
+			overload_kill_restart_after => arweave_config:get(
+				[logging,overload_kill_restart_after],
+				5000
+			)
 		},
 		formatter => {
 			logger_formatter, #{
-				chars_limit => 16256,
-				depth => 256,
+				chars_limit => arweave_config:get(
+					[logging,formatter,chars_limit],
+					16256
+				),
+				depth => arweave_config:get(
+					[logging,formatter,depth],
+					256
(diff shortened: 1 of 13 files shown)


apps/arweave/src/ar_logger.erl
```diff
@@ -159,28 +211,80 @@ handlers() -> #{
...
 			}),
-			max_no_files => 10,
-			max_no_bytes => 51_418_800,
+			compress_on_rotate => arweave_config:get(
+				[logging,handlers,debug,compress_on_rotate],
+				false
+			),
+			max_no_files => arweave_config:get(
+				[logging,handlers,debug,max_no_files],
+				10
```
JamesPiechota, line 220: I think 2.9.5 defaults to 20 debug logfiles

apps/arweave/src/ar_logger.erl
```diff
@@ -110,35 +110,87 @@ handlers() -> #{
...
 		config => #{
 			type => file,
 			file => logfile_path(#{
 				prefix => "arweave",
 				level => info
 			}),
-			max_no_files => 10,
-			max_no_bytes => 51_418_800,
+			compress_on_rotate => arweave_config:get(
+				[logging,compress_on_rotate],
```
JamesPiechota, line 121: Is the intent to have `[logging,xxx]` here rather than `[logging,handlers,info,xxx]`? Just checking since debug uses the latter

apps/arweave_config/src/arweave_config_parameters.erl
```diff
@@ -22,6 +23,7 @@
 init() ->
 	[
 		#{
+			enabled => true,
```
JamesPiechota, line 26: How is `enabled` used? I see it queried in arweave_config_spec_enabled, but I don't see where the value is used

apps/arweave_config/src/arweave_config_parameters.erl
```diff
@@ -52,28 +55,659 @@ init() ->
...
-							legacy_set(K, V, S);
+							logger:set_application_level(arweave_config, info),
+							logger:set_application_level(arweave, info),
+							ar_logger:start_handler(arweave_debug),
+							legacy_set(K, V, S, []);
 						{true, false} ->
 							logger:set_application_level(arweave_config, debug),
 							logger:set_application_level(arweave, debug),
-							legacy_set(K, V, S)
+							ar_logger:stop_handler(arweave_debug),
```
JamesPiechota: If `Old = false` and `V = true` - shouldn't we start the arweave_debug handler? Or maybe I misunderstand?

apps/arweave_config/src/arweave_config_parameters.erl
```diff
@@ -52,28 +55,659 @@ init() ->
...
+					{store, true};
+				(_,false,_,_) ->
+					ar_logger:stop_handler(arweave_debug),
+					{store, false}
+			end
+		},
+		#{
+			enabled => true,
+			parameter_key => [logging,handlers,debug,max_no_files],
+			inherit => {[logging,max_no_files], [type, default]},
```
JamesPiechota: Why does this one need the `[type, default]`?

apps/arweave/src/ar_logger.erl
humaite: My idea was to using `[logging,handlers,*]` for extra handlers. The default one (info) is defined in `[logging,...]`. I think every users have info logs enabled, like the console logs (I moved into `sys.config`).

apps/arweave_config/src/arweave_config_parameters.erl
humaite: All parameters specifications are using module callback called `arweave_config_spec_*`. Each modules have the same structure, a bit like a behavior, with `init/2` exported. When the args are configured, `arweave_config_spec` process will check every single parameters specifications and load them into an ETS table. When a parameter specification is checked, a list of valid keys is defined in `arweave_config_spec:callbacks_check/0`:

```erlang
callbacks_check() -> [
        % mandatory callbacks
        {parameter_key, arweave_config_spec_parameter_key},

        % optional callbacks
        {enabled, arweave_config_spec_enabled},
        {handle_get, arweave_config_spec_handle_get},
        {handle_set, arweave_config_spec_handle_set},
        {default, arweave_config_spec_default},
        {type, arweave_config_spec_type},
        {runtime, arweave_config_spec_runtime},
        {deprecated, arweave_config_spec_deprecated},
        {environment, arweave_config_spec_environment},
        {short_argument, arweave_config_spec_short_argument},
        {long_argument, arweave_config_spec_long_argument},
        {legacy, arweave_config_spec_legacy},
        {short_description, arweave_config_spec_short_description},
        {long_description, arweave_config_spec_long_description},
        {nargs, arweave_config_spec_nargs},
        {inherit, arweave_config_spec_inherit}
].
```

As you can see, the `enabled` specification can be found in this function as tuple, where the first element of the tuple defines the name of the valid specification key, and the second element defines where the module is defined.

For the case of `enabled` if defined to `false` or `{false, Msg}` where `Msg` is a `string()` or a `binary()` , the specification is not loaded and if defined, a DEBUG message is printed in logs.

```erlang
% from arweave_config_spec_enabled
fetch(#{ enabled := false }, State) ->
        PK = maps:get(parameter_key, State),
        ?LOG_DEBUG("parameter_key: ~p disabled", [PK]),
        skip;
fetch(#{ enabled := {false, Reason} }, State) ->
        PK = maps:get(parameter_key, State),
        ?LOG_DEBUG("parameter_key: ~p disabled (~p)", [PK, Reason]),
        skip;
```

PR #924, opened 2025-12-03
Fix Cowboy/Ranch Warning Message

humaite:
When starting cowboy/ranch, a warning message was displayed:

    Setting Ranch options together with socket options is deprecated.
    Please use the new map syntax that allows specifying socket options
    separately from other options.

see: https://github.com/ArweaveTeam/arweave-dev/issues/1019


JamesPiechota (approved):
LGTM!

PR #925, opened 2025-12-03
Disabling fail-fast on workflow

humaite:
To improve speed, fail-fast should be disabled. In case of timeout, the workflow will be restarted manually.


JamesPiechota (approved):
LGTM!

PR #927, opened 2025-12-10
Add Unix Signal Support

humaite:
Arweave uses default Erlang signal handler, most of the time,
this will do the job. Unfortunately, in some case, it can be
very helpful to have custom signal, for example when we are
losing control of the arweave node due to epmd session leak
or crash.

This commit adds custom signal handling for the following
signals:

  - sighup: reload the configuration

  - sigquit: exit arweave and generate a core dump

  - sigstop: do nothing for now

  - sigterm: stop arweave

  - sigusr1: print arweave state in logs

  - sigusr2: recover arweave

Both of these signals can be used with the kill or pkill commands,
for example, with sighup:

    kill -SIGHUP ${arweave_pid}

Instead of polluting arweave_config with diagnostic function and other
procedures to extract information on the node, a new application/lib
called arweave_diagnostic was created. This module will contain all
modules to extract diagnostic and recover arweave in case of issue.
Here a list of diagnostic created:

  - cpu: arweave_diagnostic:cpu/0
  - memory: arweave_diagnostic:memory/0
  - memory_worst: arweave_diagnostic:memory_worst/0
  - processes: arweave_diagnostic:processes/0
  - network: arweave_diagnostic:network/0
  - arweave: arweave_diagnostic:arweave/0
  - ets: arweave_diagnostic:ets/0
  - dets: arweave_diagnostic:dets/0
  - rocksdb: arweave_diagnostic:rocksdb/0

One can call also arweave_diagnostic:all/0 or
arweave_diagnostic:select/1.

see: https://github.com/ArweaveTeam/arweave-dev/issues/1000

### DIFF
--- a/apps/arweave_config/src/arweave_config.app.src
+++ b/apps/arweave_config/src/arweave_config.app.src
@@ -12,10 +12,40 @@
 	]},
 	{modules, [
 		arweave_config,
-		arweave_config_sup
+		arweave_config_environment,
+		arweave_config_http_server,
+		arweave_config_legacy,
+		arweave_config_parameters,
+		arweave_config_parser,
+		arweave_config_signal_handler,
+		arweave_config_spec,
+		arweave_config_spec_default,
+		arweave_config_spec_deprecated,
+		arweave_config_spec_enabled,
+		arweave_config_spec_environment,
+		arweave_config_spec_handle_get,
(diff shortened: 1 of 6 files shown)


JamesPiechota (approved):
LGTM!

PR #928, opened 2025-12-10
Add new Ranch metrics

khetzl:
Metrics of number of Ranch active connections, all connection, and acceptor processes.


apps/arweave/src/ar_metrics_collector.erl
```diff
@@ -36,6 +36,7 @@ add_metric_family({Name, Type, Help, Metrics}, Callback) ->
 	Callback(create_mf(?METRIC_NAME(Name), Help, Type, Metrics)).
 
 metrics() ->
+	RanchInfo = ranch:info(),
```
humaite, line 39: Why not moving this function somewhere else, like directly in `get_ranch_active_connections/2`?

khetzl: I'm trying to save that one ranch_info/0 call.

apps/arweave/src/ar_metrics_collector.erl
```diff
@@ -56,7 +57,13 @@ metrics() ->
 	 {ar_header_sync_bytes_total, gauge, "ar_header_sync process memory",
 		get_process_memory(ar_header_sync)},
 	 {ar_wallets_bytes_total, gauge, "ar_wallets process memory",
-		get_process_memory(ar_wallets)}
+		get_process_memory(ar_wallets)},
```
JamesPiechota, line 54: I know this long predates your PR (and even predates my time on the project), But I think we can remove all of the `message_queue_len` and process memory metrics. Unles @humaite you use them for something?

We currently track process memory and message queue length for all processes (not just the ones mentioned here) via the ar_process_sampler module (when `debug` is on).

humaite: No problem on my side. I'm currently also working on another tool to extract more information regarding all processes. I think we can remove it.

khetzl: I'll remove it. If we are updating this module, might as well.

PR #929, opened 2025-12-10
fix: remove histogram buckets from arweave metrics

JamesPiechota:
I don't believe any of our charts use the histogram data - always dooing "sum / count" to get an average over time. But the buckets multiply the number of series we pay for in grafanacloud.

### DIFF
--- a/apps/arweave/src/ar_metrics.erl
+++ b/apps/arweave/src/ar_metrics.erl
@@ -58,7 +58,7 @@ register() ->
 	%%       See: https://github.com/deadtrickster/prometheus.erl/blob/6dd56bf321e99688108bb976283a80e4d82b3d30/src/prometheus_time.erl#L2-L84
 	prometheus_histogram:new([
 		{name, ar_http_request_duration_seconds},
-		{buckets, [0.01, 0.1, 0.5, 1, 5, 10, 30, 60]},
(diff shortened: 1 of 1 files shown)


humaite (approved):
LGTM. I tested the patch on a test server, and it produces the following output:

```
# TYPE ar_http_request_duration_seconds histogram                                                                                                                                                                  
# HELP ar_http_request_duration_seconds The total duration of an ar_http:req call. This includes more than just the GUN request itself (e.g. establishing a connection, throttling, etc...)
ar_http_request_duration_seconds_bucket{http_method="POST",route="/peers",status_class="unknown",le="+Inf"} 2                      
ar_http_request_duration_seconds_count{http_method="POST",route="/peers",status_class="unknown"} 2                                                                                                                 
ar_http_request_duration_seconds_sum{http_method="POST",route="/peers",status_class="unknown"} 0.287701837            
ar_http_request_duration_seconds_bucket{http_method="POST",route="/peers",status_class="success",le="+Inf"} 110             
ar_http_request_duration_seconds_count{http_method="POST",route="/peers",status_class="success"} 110                                                                                                               
ar_http_request_duration_seconds_sum{http_method="POST",route="/peers",status_class="success"} 24.444222049
```

I was unsure of the behavior when setting it to `infinity`.

PR #930, opened 2025-12-10
fix: remove some more unused metrics and series

JamesPiechota:


config/sys.config
```diff
@@ -23,6 +23,7 @@
 	]},
 	{prometheus, [
 		{cowboy_instrumenter, [
+			{duration_buckets, [infinity]},
```
khetzl, line 26: do i understand correctly that this will set the histograms to have a single bucket up to infinity for the cowboy metrics?

khetzl (approved):

config/sys.config
JamesPiechota: yep! In truth I'd love to remove the `_bucket` metric entirely since we really only use `_count` and `_sum` but this is the closest I've been able to get to that.

If we end up ever needing the histogram functionality we can re-add - but we haven't needed it to date (and it explodes the metrics)

PR #931, opened 2025-12-12
New Argument Parser Support

humaite:
This PR is adding support for the new argument parser, using Unix/Linux like convention and backed by arweave_config. The main idea is to set arweave from the command line with the same naming convention present in arweave parameters. Those parameters will then reflect the naming from environment or configuration file. Here an example:

```sh
./arweave --debug
./arweave --debug true
```

These arguments will be converted using the specification present in `arweave_config_parameters` and will set `[debug]` to `true`. By default, if long argument is defined in the specification, the full parameter key will be converted:

```sh
./arweave --data.directory ./data
```

see: https://github.com/ArweaveTeam/arweave-dev/issues/971

Warning: this part of the application is a work in progress, including
many new features to be tested in real world environment. arweave_config
is considered unstable for now and its interfaces will probably change
in a near future.

This commit adds partial support for the new arweave command line
arguments format, using a more conventional Unix/Linux like convention
and backed by arweave_config. The main idea is to set arweave from
command line using the very same naming structure than the parameter
keys used in the specification/parameters (mostly found in
arweave_config_parameters module at this time of writing). Then, it will
be consistent with environment variables as well.

Long and short arguments are supported, both of them can be
enabled/disabled/configured by setting the right value in the
specifications when specifying a new parameter. The code below will
automatically generate an argument using the parameter key as reference.
The final result will be --test.argument.

    #{
      parameter_key => [test,argument],
      long_argument => true
    }.

In some situation, it can also be useful to define a non-standard
argument from command line. In this case, it can be overwritten by
configuring a binary instead of a boolean.

    #{
      parameter_key => [test,argument],
      long_argument => <<"--some-other-format">>,
    }.

A short argument can be configured using short_argument specification
key, and using an integer.

    #{
      parameter_key => [test,argument],
      short_argument => $t
    }.

The new format is disabled by default, and only legacy format (including
legacy configuration format) is being used. If one wants to use the new
format, AR_CONFIG_MODE environment variable must be set to "new".

    export AR_CONFIG_MODE="new"

ARWEAVE LEGACY FORMAT REFACTORING

The legacy parser was mostly defined in ar module, a new module
containing it called ar_cli_parser was created to isolate this code in
its own namespace. A small cleanup has been done on ar module as well,
like removing unused functions.

ARWEAVE CONFIG BOOTSTRAP

arweave_config is usually started before arweave, but this one is not
aware of the configuration file or any other arguments passed through
the command line. A startup phase is then required to initialize it.
This part of the process is now managed by arweave_config_bootstrap, in
particular via arweave_config_bootstrap:start/1 function. This one is
acting as a finite state machine, taking the arguments from the command
lines and returning the final legacy configuration. This part is
temporary and was created only to deal with legacy and new format at the
same time.

MISCELLANEOUS

  - Fixed many small bugs and typos

  - Fixed some logic issues

  - Added more function specifications

  - Improved the documentation

NEXT STEP

Add support for the new configuration file format.

---

> [!NOTE]
> Introduces a new, spec-driven CLI argument parser and startup bootstrap while preserving legacy behavior by default.
> 
> - New modules: `arweave_config_arguments` (long/short args → params), `arweave_config_arguments_legacy`, `arweave_config_bootstrap` (env → config file → args → runtime), and supporting updates in `arweave_config_spec`, `arweave_config_parameters` (e.g., `--data.directory`), `arweave_config_type` (boolean parsing), `arweave_config_environment`, `arweave_config_legacy` (merge/set), and `arweave_config_sup` (starts new services)
> - `ar.erl` refactored: legacy CLI moved to `ar_cli_parser`; startup now uses `arweave_config_bootstrap`; removed large inline parser/help and added `warnings_as_errors`
> - `ar_config` gains `parse_config_file/1` and file reading helpers
> - CI: run common tests/eunit under `as test` with coverage, generate/upload coverage artifacts; include `ar_cli_parser` tests in VDF/full workflows; add new CT suites for arguments/bootstrap
> 
> Risk: Medium. Touches startup/config path; new mode gated by `AR_CONFIG_MODE=new`, but legacy flow extraction may affect initialization.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e77b0f1aab0b2c2e07d58ccc0bafd0112e0f0483. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/arweave_config/src/arweave_config_bootstrap.erl
```diff
@@ -0,0 +1,204 @@
...
+	% @todo enable arweave_config_file_legacy.
+	case ar_cli_parser:parse_config_file(Args) of
+		Config when is_record(Config, config)  ->
+			arweave_config_legacy:merge(Config),
+			{next, init_arguments, State};
+		{error, Reason, _} ->
+			{error, Reason};
+		{error, Reason} ->
+			{error, Reason}
+	end.
```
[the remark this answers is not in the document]
humaite: can you recheck?

apps/arweave_config/src/arweave_config.erl
```diff
@@ -211,7 +211,18 @@ set(Key, Value) when is_atom(Key) ->
...
+			case arweave_config_spec:get_legacy(Key) of
+				{ok, PK} ->
+					_ = set(PK, Value),
+					{ok, V};
+				Else ->
+					Else
+			end;
+		Else ->
+			Else
+	end;
```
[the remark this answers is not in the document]
humaite: This is a "temporary" issue. This function should not be called a lot and will be removed when the new format will be totally integrated in arweave.

.github/workflows/x-test-full.yml
```diff
@@ -135,7 +133,8 @@ jobs:
             ar_vdf_server_tests,
             ar_vdf_external_update_tests,
             ar_post_block_tests,
-            ar_reject_chunks_tests
+            ar_reject_chunks_tests,
+            ar_cli_parser
```
JamesPiechota, line 140: Do we use `on-demand.yml`? I've been adding all my test files to that .yml in addition to x-test-full and x-test-vdf

apps/arweave/src/ar_cli_parser.erl
```diff
@@ -0,0 +1,1172 @@
...
+%%% The GNU General Public License (GPL-2.0)
+%%% Version 2, June 1991
+%%%
+%%% ------------------------------------------------------------------
+%%%
+%%% @copyright 2025 (c) Arweave
+%%% @author Arweave Team
+%%% @doc `ar_cli_parser' module legacy command line argument parser.
+%%%
+%%% This module has been created from `ar' module. The code is barely
```
JamesPiechota:
```suggestion
%%% This module has been created from `ar' module. The code is mostly
```

I think right? "barely the same" means not the same

apps/arweave_config/src/arweave_config.erl
```diff
@@ -211,7 +211,18 @@ set(Key, Value) when is_atom(Key) ->
 		{key, Key},
 		{value, Value}
 	]),
-	arweave_config_legacy:set(Key, Value);
+	case arweave_config_legacy:set(Key, Value) of
+		{ok, V} ->
+			case arweave_config_spec:get_legacy(Key) of
+				{ok, PK} ->
```
JamesPiechota, line 217: `PK` in this function is the legacy config option name, right? e.g. `mining_server_chunk_cache_size_mb`?

.github/workflows/x-test-full.yml
humaite: Not a lot. I think I used one or two times. We can probably remove it.

apps/arweave_config/src/arweave_config.erl
humaite: correct.

apps/arweave_config/src/arweave_config_arguments.erl
```diff
@@ -0,0 +1,568 @@
...
+%%%   }
+%%% }
+%%% '''
+%%%
+%%% Another example with storage modules. One wants to configure
+%%% storages module with many different options.
+%%%
+%%% ```
+%%% --storage.module 0 \
+%%%   --protocol unpacked \
```
JamesPiechota:
```suggestion
%%%   --format unpacked \
```

apps/arweave_config/src/arweave_config_arguments.erl
```diff
@@ -0,0 +1,568 @@
...
+%%%   --enabled \
+%%% --storage.module 0
+%%%   --protocol replica.2.9 \
+%%%   --start 0 \
+%%%   --end 7200000000000 \
+%%%   --pubkey L-cPRwGcnMFyQW-APh_fS4lMGioLg76-ECovqXIlmJ4 \
```
JamesPiechota:
```suggestion
%%%   --address L-cPRwGcnMFyQW-APh_fS4lMGioLg76-ECovqXIlmJ4 \
```

apps/arweave_config/src/arweave_config_arguments.erl
```diff
@@ -0,0 +1,568 @@
...
+%%% storages module with many different options.
+%%%
+%%% ```
+%%% --storage.module 0 \
+%%%   --protocol unpacked \
+%%%   --enabled \
+%%% --storage.module 0
+%%%   --protocol replica.2.9 \
+%%%   --start 0 \
+%%%   --end 7200000000000 \
```
JamesPiechota: Unfortunately th current storage_module format doesn't use a strict "start, end" - it defines a size and then an offset of those size buckets.

e.g. `3,10000` means that storage module covers 10000 bytes and is the 4th bucket of that. e.g. 30000-40000

apps/arweave_config/src/arweave_config_arguments.erl
```diff
@@ -0,0 +1,568 @@
...
+%%%           "protocol: "replica.2.9",
+%%%           "start": "0",
+%%%           "end": "7200000000000",
+%%%           "pubkey": "L-cPRwGcnMFyQW-APh_fS4lMGioLg76-ECovqXIlmJ4",
+%%%           "status": "enabled"
+%%%         }
+%%%      ]
+%%%   }
+%%% }
+%%% '''
```
JamesPiechota: I'd suggest we don't use storage_module as an example here as I think it may confuse readers. Either that or we should probably align the documentation with the current storage_module implementation (format, address, bucket size, count, etc...)

apps/arweave_config/src/arweave_config_legacy.erl
```diff
@@ -221,6 +252,37 @@ terminate(_, _) ->
...
+			MergedProplist = lists:foldr(fun
+				% same values, nothing to change
+				({{K, NV}, {K, OV}}, Acc) when NV =:= OV ->
+					[{K, OV}|Acc];
+				% different values, we set the new one
+				({{K, NV}, {K, OV}}, Acc) when NV =/= OV ->
+					[{K, NV}|Acc];
+				% something wrong, the configuration
+				% is bad
+				(Else, _Acc) ->
```
JamesPiechota, line 270: This case is triggered if there's a Key in one config and not in the other, right? But I;m guessing somewhere else we make sure that `proplist` is intialized to the set of properties in `#config`?

JamesPiechota: ah, `config_to_proplist` is that place, right?

apps/arweave_config/src/arweave_config_parameters.erl
```diff
@@ -723,7 +775,8 @@ init() ->
 		},
 		#{
 			parameter_key => [config,http,api,listen,port],
-			environment => <<"AR_CONFIG_HTTP_API_LISTEN_PORT">>,
+			environment => true,
+			long_argument => true,
```
JamesPiechota, line 779: When `long_argument` is true, we set the long argument to the `parameter_key` joined with `.`?

humaite: correct. it auto-generate the long argument based on the parameter key, by default, separated by `.`. We can eventually change that in the future, or depending on the context.

apps/arweave_config/src/arweave_config_legacy.erl
humaite: yes, we have `config_to_proplist/1` and `proplist_to_config/1`, both functions are only for legacy configuration. In the future, they will not be required. Note: the proplist is a bit slow, converting it to map will be done in the future.

PR #932, opened 2025-12-12
Fix Github Workflow Issue Due to Missing Files

humaite:
Adding a new application in rebar3 tree requires specific files, at least on arweave point of view. After the release is compiled, a github job will create artifacts (to be reused in other jobs), but an error is produced instead:

    tar: ./_build/default/lib/arweave_diagnostic/include: File removed before we read it
    tar: ./_build/default/lib/arweave_diagnostic/priv: File removed before we read it
    tar: ./_build/default/rel/arweave/lib/arweave_diagnostic-0.0.1/include: File removed before we read it
    tar: ./_build/default/rel/arweave/lib/arweave_diagnostic-0.0.1/priv: File removed before we read it
    tar: ./_build/test/lib/arweave_diagnostic/include: File removed before we read it
    tar: ./_build/test/lib/arweave_diagnostic/priv: File removed before we read it
    tar: ./_build/test/rel/arweave/lib/arweave_diagnostic-0.0.1/include: File removed before we read it
    tar: ./_build/test/rel/arweave/lib/arweave_diagnostic-0.0.1/priv: File removed before we read it

include and priv folders are required. Those must be present in the github repository as well.


humaite:
It fixed the issues, github workflows are working again.

PR #933, opened 2025-12-15
Rate limit restructure (server-side)

khetzl:
Multiple rate limiter actors to be called from middleware/cowboy_handler processes.

Implementation based on Leaky Bucket Token (https://gist.github.com/humaite/21a84c3b3afac07fcebe476580f3a40b) with additional limits for concurrency (similarly to Ranch concurrency limiter).

It could potentially allow complex rules set to map requests to limiter pools and/or custom settings via (reworked) ar_config.

The current implementation is not "wired" in. So the new middleware module is not added to `ar_http_iface_server`, and the supervisor is not yet part of the `ar_node` supervision tree.

The functionality could be extended with values helpful to regulate clients via the `Status 429` response headers.

---

> [!NOTE]
> Replaces legacy request throttling with a new, modular server-side rate limiter and integrates it into the HTTP stack.
> 
> - Adds `arweave_limiter` app (supervisor, `arweave_limiter_group`, time helper, metrics + collector) with sliding-window + leaky-bucket + concurrency controls per limiter group
> - Wires limiter via new `ar_http_iface_rate_limiter_middleware` (mapped by path; local peers bypass) and replaces `ar_blacklist_middleware` in `ar_http_iface_server`
> - Trims `ar_blacklist_middleware` to ban-only; removes in-request rate limiting; `handle_post_tx_accepted` now calls `arweave_limiter:reduce_for_peer(general, IP)`
> - Extends `ar_config` and `arweave_config.hrl` with comprehensive limiter settings (per group: general, chunk, data_sync_record, recent_hash_list_diff, block_index, wallet_list, get_vdf, get_vdf_session, get_previous_vdf_session, metrics)
> - Starts limiter at node/test startup and during shutdown (`ar.erl`, `ar_test_node.erl`, `ar_test_runner.erl`); updates `rebar.config` (app inclusion, meck bump) and test runner script to include multi-app test paths
> - Adds unit tests for limiter group and metrics collector; updates `ar_http_iface_tests` to use new limits and resets; CI workflows include new test modules
> - Minor: `.gitignore` tweak for backup files
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c0d21906aa510b89fc2aa51a00b43d747073be20. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


JamesPiechota:
Ha. Interesting - I would have thought the ++ would have caused an error. Good to know!

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,325 @@
...
+    Result.
+
+reset_all(LimiterRef) ->
+    whereis(LimiterRef) == undefined orelse gen_server:call(LimiterRef, reset_all).
+
+stop(LimiterRef) ->
+    gen_server:stop(LimiterRef).
+
+%% gen_server callbacks
+init([Args]) ->
```
humaite: Why using `[Args]` and not just `Args`?

bin/arweave
```diff
@@ -514,9 +514,19 @@ arweave_test_run() {
 	    TEST_NODE="${TEST_NODE_NAME}@${TEST_NODE_HOST}"
 	fi
 
-	TEST_PATH="$(./rebar3 as ${TEST_PROFILE} path)"
-	TEST_PATH_BASE="$(./rebar3 as ${TEST_PROFILE} path --base)/lib/arweave/test"
-	PARAMS="-pa ${TEST_PATH} ${TEST_PATH_BASE} -config ${TEST_CONFIG} -noshell"
+        TEST_PATH="$(./rebar3 as ${TEST_PROFILE} path)"
+        
+        ## TODO: Generate path for all apps -> Should we fetch this from somewhere?
+        APPS=("arweave" "arweave_config" "arweave_limiter" "arweave_diagnostic")
```
humaite: You could remove the "bash-like" list. It's not really portable and could cause trouble. On POSIX shell, a list is a string made of words separated with spaces or null chars.

```
APPS="arweave arweave_config arweave_limiter arweave_diagnostic"
```

In this case, the code on L523 will simply looks like that:

```
for app in ${APPS}
do
  ...
done
```

Using specific bash features is usually not a good idea.

bin/arweave
```diff
@@ -514,9 +514,19 @@ arweave_test_run() {
 	    TEST_NODE="${TEST_NODE_NAME}@${TEST_NODE_HOST}"
 	fi
 
-	TEST_PATH="$(./rebar3 as ${TEST_PROFILE} path)"
-	TEST_PATH_BASE="$(./rebar3 as ${TEST_PROFILE} path --base)/lib/arweave/test"
-	PARAMS="-pa ${TEST_PATH} ${TEST_PATH_BASE} -config ${TEST_CONFIG} -noshell"
+        TEST_PATH="$(./rebar3 as ${TEST_PROFILE} path)"
+        
+        ## TODO: Generate path for all apps -> Should we fetch this from somewhere?
```
humaite, line 569: In theory, those paths are generated using rebar3, but it seems for the case of tests, it was not correctly done.

apps/arweave/src/ar_http_iface_rate_limiter_middleware.erl
```diff
@@ -0,0 +1,69 @@
+-module(ar_http_iface_rate_limiter_middleware).
```
humaite, line 20: can you add an header with a bit of documentation to explain the role of this middleware?

apps/arweave/src/ar_http_iface_rate_limiter_middleware.erl
```diff
@@ -0,0 +1,69 @@
...
+      Req
+     ).
+
+get_peer_key(Req) ->
+    {{A, B, C, D}, Port} = cowboy_req:peer(Req),
+    {A, B, C, D, Port}.
+
+config_peer_to_ip_addr({A, B, C, D, _}) -> {A, B, C, D}.
+
+path_to_limiter_ref([<<"chunk">> | _]) -> chunk;
```
humaite, line 69: If I understand correctly, this function is reading the path from the client request, and will return the kind of limiter used. Using a data structure like a proplist or a map can probably be easier to modify in the future. Not sure if could be easy to do right now though. 

The problem I see here is, if we are adding a new path, we will need to add it in this function (at least if we need a special rate limit for it).

khetzl: Yes, it's certainly going to some work to add a new rate limiter group. 
1) add Rate Limiter Group (RLG) process to `arweave_limiter_sup`
2) add Rate Limiter Group config parameters as fields to arweave_config `config` record.
2b) add defaults for the parameters.
3) add Rate Limiter routing logic to `ar_http_iface_rate_limiter_middleware`

IMHO, this mapping function is a quite idiomatic way to map paths to atoms. (No dynamically named atoms, no need for lookups)

- I'd debate that 2) is a bigger burden than 3) (extending this function). 
- I'd like to point out: this PR doesn't aim to introduce dynamic RLGs.
- With a different approach on how we deal with the configuration, I think we can introduce RLGs started in a dynamic manner in the future, if necessary.

humaite: We are not adding a lot of new end-point, but perhaps documenting this procedure somewhere could be nice.

apps/arweave/src/ar_blacklist_middleware.erl
```diff
@@ -1,35 +1,17 @@
 -module(ar_blacklist_middleware).
 
--behaviour(cowboy_middleware).
-
--export([start/0, execute/2, reset/0, reset_rate_limit/3,
-		ban_peer/2, is_peer_banned/1, cleanup_ban/1, decrement_ip_addr/2]).
+-export([start/0, ban_peer/2, is_peer_banned/1, cleanup_ban/1]).
 -export([start_link/0]).
 
+-ifdef(TEST).
```
JamesPiechota:
```suggestion
-ifdef(AR_TEST).
```

We've had some name collisions with the `?TEST` define in some dependencies, so we use `?AR_TEST` now to indicate whether or not Arweave is being run in test-mode (it gets set in [rebar.config](https://github.com/ArweaveTeam/arweave/blob/master/rebar.config#L442))

JamesPiechota:
The `RPM_BY_PATH` logic as well as the `ar_rate_limiter` module - they're both related to the client-side rate limiting, right? i.e. no change to them in this PR, but they will likely be removed/updated in a future PR when the client-side functionality is added.

apps/arweave/src/ar_blacklist_middleware.erl
```diff
@@ -1,35 +1,17 @@
...
+
 -include_lib("arweave/include/ar.hrl").
 -include_lib("arweave_config/include/arweave_config.hrl").
 -include_lib("arweave/include/ar_blacklist_middleware.hrl").
 -include_lib("eunit/include/eunit.hrl").
 
-execute(Req, Env) ->
-	IPAddr = requesting_ip_addr(Req),
-	{ok, Config} = arweave_config:get_env(),
-	case lists:member(blacklist, Config#config.disable) of
```
JamesPiechota, line 17: To be honest I didn't realize we had a `disable blacklist` option, and I'm not sure if it is currently in use. It looks like it would completely disable rate limiting on the server side. @ldmberman @humaite are you aware of any nodes that use it? 

It's *possible* some large miners use it on internal nodes used to server data to mining nodes (for example). But I'm not sure it would even be that effective since by default clients will self-throttle regardless of the server-side behavior.

This is a long way of asking:
1. @khetzl did you mean to remove this option? (e.g. because you've replaced it with another configuration option)
2. Even if the removal was unintentional... I'm not sure we actually need or want this particular flag. In general the `enable` and `disable` options are poorly understood and rarely used, so if we want to disable server-side rate limiting a more formal config is a better choice. Only question is whether anyone is using the current flag?

apps/arweave/src/ar_blacklist_middleware.erl
khetzl: There will be quite a few of these, I can change change all to `AR_TEST`

apps/arweave_config/include/arweave_config.hrl
```diff
@@ -173,6 +173,100 @@
 -define(DEFAULT_COWBOY_TCP_SEND_TIMEOUT, 15_000).
 -define(DEFAULT_COWBOY_TCP_LISTENER_SHUTDOWN, 5000).
 
+%% Common RLG Settings
+-define(DEFAULT_HTTP_API_LIMITER_TIMESTAMP_CLEANUP_INTERVAL, 120000).
+-define(DEFAULT_HTTP_API_LIMITER_TIMESTAMP_CLEANUP_EXPIRY, 120000).
+-define(DEFAULT_HTTP_API_LIMITER_IS_MANUAL_REDUCTION_DISABLED, false).
+
+%% General RLG
+-define(DEFAULT_HTTP_API_LIMITER_GENERAL_SLIDING_WINDOW_LIMIT, 150).
```
JamesPiechota: These are the defaults that apply across the board to all endpoints unless specifically overridden?

apps/arweave/src/ar_blacklist_middleware.erl
khetzl: I intended to remove this, we shouldn't allow unrestricted access to resources to the public.
Trusted peers can be listed so their requests are limited via a less restrictive Rate Limiter Group. (Currently, no limiting at all)

Note: Even if the `disable blacklist` flag is enabled in the current release, a global concurrency limit is enforced by cowboy/ranch.

JamesPiechota: Got it. I agree it's correct to remove. However worth providing some context:
- In cases like this where it is the right call to remove an option, we may still need to flag and communicate the removal for any miners that are relying on it. Both in the release notes and in Discord. I suspect this option isn't used much (or at all) so probably fine to silently remove.
- Many miners implement their own security in front of their node. Specifically they will run nodes on a LAN without direct internet connection - so even if they opt in to using the `disable blacklist` flag it doesn't necessarily mean they or we are providing unrestricted access to the public.
- I don't think this flag is used, but if it is used it is likely due to the above. We've had a few miners run large networks with many internal (non publicly accessible nodes). In the past they tried using `local_peers` but since arweave doesn't currently support any sort of wildcards/subnetting, and because the ndoe shutdown/boot time can take a long time, some of the miners had trouble managing rate limiting for their internal nodes. It's in that scenario where I could see `disable blacklist` being useful (setup an internal data-sharing node, disable blacklist, and then let all your other nodes connect to without worrying about needing to restart periodically to update local_peers)

All that said: I do think fine to remove. But I wanted to call out that context as our operating model is often different than that of non-blockchain companies (e.g. we build infrastructure software to be run by 3rd parties as well as by us internally - and those 3rd parties vary from novice to enterprise-level)

apps/arweave_config/include/arweave_config.hrl
```diff
@@ -314,7 +408,199 @@
...
+	'http_api.limiter.get_previous_vdf_session.leaky_tick_intercal' =
+                     ?DEFAULT_HTTP_API_LIMITER_GET_PREVIOUS_VDF_SESSION_LEAKY_TICK_INTERVAL,
+	'http_api.limiter.get_previous_vdf_session.leaky_tick_reduction' =
+                     ?DEFAULT_HTTP_API_LIMITER_GET_PREVIOUS_VDF_SESSION_LEAKY_TICK_REDUCTION,
+	'http_api.limiter.get_previous_vdf_session.concurrency_limit' =
+                     ?DEFAULT_HTTP_API_LIMITER_GET_PREVIOUS_VDF_SESSION_CONCURRENCY_LIMIT,
+	'http_api.limiter.get_previous_vdf_session.is_manual_reduction_disabled' =
+                     ?DEFAULT_HTTP_API_LIMITER_IS_MANUAL_REDUCTION_DISABLED,
+
+	'http_api.limiter.metrics.sliding_window_limit' =
```
JamesPiechota, line 605: @khetzl @humaite What's the recommended approach for the new config system regarding how deep to nest configs?

E.g. Is there a clear preference `http_api_limiter.metrics.sliding_window_timestamp_cleanup_interval` vs. `http_api_limiter.metrics.sliding_window.timestamp_cleanup_interval`?

Especially in a JSON format I could see grouping all the `sliding_window` or `leaky_tick` options together making it a little easier to keep everything straight.

apps/arweave_limiter/src/arweave_limiter.erl
```diff
@@ -0,0 +1,86 @@
...
+%%%
+%%% ------------------------------------------------------------------
+%%%
+%%% @copyright 2025 (c) Arweave
+%%% @author Arweave Team
+%%% @author Kristof Hetzl
+%%% @doc Arweave Rate Limiter.
+%%%
+%%% `arweave_limiter' module is an interface to the Arweave
+%%% Rate Limiter functionality.
```
JamesPiechota, line 14: Nice. I like this approach - really helps with modularization 👍

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,325 @@
...
+            prometheus_counter:inc(ar_limiter_rejected_total,
+                                   [atom_to_list(LimiterRef), atom_to_list(Reason)]),
+            Rejection;
+        Accept ->
+            Accept
+    end.
+
+reduce_for_peer(LimiterRef, Peer) ->
+    Result = gen_server:call(LimiterRef, {reduce_for_peer, Peer}),
+    Result == ok andalso prometheus_counter:inc(ar_limiter_reduce_requests_total,
```
JamesPiechota, line 76: I haven't seen this structure before. Any benefit vs.

```
ok = Result = gen_server:call(LimiterRef, {reduce_for_peer, Peer}),
prometheus_counter:inc(ar_limiter_reduce_requests_total,
```

Ah wait, I think I see the distinction. My version would crash, your version would just silently not call prometheus_count:inc, right?

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,325 @@
...
+
+reset_all(LimiterRef) ->
+    whereis(LimiterRef) == undefined orelse gen_server:call(LimiterRef, reset_all).
+
+stop(LimiterRef) ->
+    gen_server:stop(LimiterRef).
+
+%% gen_server callbacks
+init([Args]) ->
+    process_flag(trap_exit, true),
```
JamesPiechota: We've gone back and forth on `process_flag(trap_exit, true),`. At one point we were adding it to all gen_servers, and then we switched to only adding it when there was a specific cleanup action we wanted to take on exit. 

What's the rationale for trapping the exit here?

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,325 @@
...
+    SlidingTimestampsForPeer0 =
+        expire_and_get_requests(Peer, SlidingTimestamps, SlidingWindowDuration, Now),
+    case IsDisabled of
+        true ->
+            {reply, {register, no_limiting_applied}, State};
+        _ ->
+            case Concurrency > ConcurrencyLimit of
+                true ->
+                    %% Concurrency Hard Limit
+                    ?LOG_WARNING([{event, ar_limiter_reject}, {reason, concurrency},
```
JamesPiechota: We may want to remove all of these ?LOG_WARNINGs or, at best, flip them to ?LOG_DEBUG.

We wrestle with how verbosely to log events that are initiated by clients. We've had issues in the past where a malicious client would attempt to flood a miner's logs via inexpensive HTTP requests.

I guess as a rule of thumb if a log can be generated by a client, we should aim to have the event that triggers the log be difficult/rare. Although log bloat is a constant battle, and I'm sure there are places in the code where this rule of thumb has been violated

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,325 @@
...
+                                                  {leaky_rate_limit, LeakyRateLimit},
+                                                  {peer, Peer}, {id, Id}]),
+                                    {reply, {reject, rate_limit, data}, State};
+                                false ->
+                                    NewLeakyTokens = update_token(Peer, Tokens, LeakyTokens),
+                                    {NewRequests, NewMonitors} =
+                                        register_concurrent(
+                                          Peer, FromPid, ConcurrentRequests, ConcurrentMonitors),
+                                    {reply, {register, leaky},
+                                     State#{leaky_tokens => NewLeakyTokens,
```
JamesPiechota, line 180: Question for my own understanding: what's the rationale of `=>` vs. `:=` here? `=>` is like upsert, right? and `:=` will fail if the key doesn't already exists? Why do we only want to fail on concurrent_monitors?

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,325 @@
...
+                                            concurrent_requests => NewRequests,
+                                            concurrent_monitors := NewMonitors}}
+                            end;
+                        _ ->
+                            {NewRequests, NewMonitors} =
```
JamesPiechota, line 185: It's possible that we hit this branch (i.e. we have capacity under the sleding window limit) while also still having a non-zero `Tokens` count, right? i.e. we have requests registered against the burst limit *and* also have space in the liding window.

I guess I could see this happening if we recently exceeded the sliding window limit and started registering tokens against the burst limit, and then some capacit opened up in the sliding window before we had finished "draining the burst limit bucket".

Is that correct? And that's expected, right?

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,325 @@
...
+    case gen_server:call(LimiterRef, {register_or_reject, Peer}) of
+        {reject, Reason, _Data} = Rejection ->
+            prometheus_counter:inc(ar_limiter_rejected_total,
+                                   [atom_to_list(LimiterRef), atom_to_list(Reason)]),
+            Rejection;
+        Accept ->
+            Accept
+    end.
+
+reduce_for_peer(LimiterRef, Peer) ->
```
JamesPiechota, line 74: Where is this called? I only see it called from `ar_http_iface_middleware:handle_post_tx_accepted`. I was expecting to see it called whenever an inbound HTTP request is processed, but perhaps I misunderstood

JamesPiechota: Ah yeah looks like I misunderstood. Reading the comment reduce_for_peer is a special case so that we don't penalize peers for sharing valid transactions with us. The normal leaky bucket and window counter tracking is handled elswhere.

JamesPiechota: Maybe add a comment to this function to describe it? Main relevant point that I see is that this function is *not* the standard way to reduce the request count for a peer, but is an edge case handler.

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,325 @@
...
+    Timestamps = maps:get(Peer, SlidingTimestamps, []),
+    drop_expired(Timestamps, SlidingWindowDuration, Now).
+
+drop_expired([TS|Timestamps], WindowDuration, Now) when TS + WindowDuration =< Now ->
+    drop_expired(Timestamps, WindowDuration, Now);
+drop_expired(Timestamps, _WindowDuration, _Now) ->
+    Timestamps.
+
+add_and_order_timestamps(Ts, Timestamps) ->
+    lists:reverse(do_add_and_order_timestamps(Ts, lists:reverse(Timestamps))).
```
JamesPiechota, line 282:
```suggestion
    %% Timestamps is ordered oldest to newest. However we expect Ts to usually be the most recent timestamp. To optimize for this case we reverse, add (likely to the front), then re-reverse
    lists:reverse(do_add_and_order_timestamps(Ts, lists:reverse(Timestamps))).
```

Or some other comment just to explain the double-reverse

apps/arweave_limiter/src/arweave_limiter_metrics.erl
```diff
@@ -0,0 +1,63 @@
...
+%%%===================================================================
+%%% Public interface.
+%%%===================================================================
+
+%% @doc Declare Arweave Rate Limiter metrics.
+register() ->
+    ok = prometheus_histogram:new([
+                                   {name, ar_limiter_response_time_milliseconds},
+                                   {help, "Time it took for the limiter to respond to requests"},
+                                   {buckets, [infinity]}, %% For meaningful results on this, we need buckets
```
JamesPiechota: Without buckets we can still get an average over time because the histogram includes `xxx_count` and `xxx_sum` fields. Most of our internal charts just use this average over time value to track performance changes. But if some degree of bucketing would be helpful for this metric (e.g. if the distribution is skewed per timeslice), defintely add them!

We have a broader issue of how to deal with our growing number of metrics, but I think we'll likely need to tackle that holistically - until then feel free to add whatever metrics you think are necessary

rebar.config
```diff
@@ -434,7 +435,7 @@
 		]}
 	]},
 	{test, [
-		{deps, [{meck, "0.8.13"}]},
+		{deps, [{meck, "1.1.0"}]},
```
JamesPiechota, line 437: Did upgrading meck address the bug you noticed?

JamesPiechota:
PR looks great! Most of my comments are minor.

One annoying comment: unfortunately we're still using tab indenting rather than space. We'd like to reformat everything, but until then would be good to have new code indented with tabs.

One question: have you been able to run any performance tests using the new code? My guess is the impact of the extra logic is minimal. Only potential bottleneck I can see if the `arweave_limiter_group` process itself, now that all requests for a given group are sent to a single process I wonder if it could get backed up under load.

apps/arweave_limiter/src/arweave_limiter_group.erl
khetzl: Yes, that's correct. The previous solution immediately reduces when a transaction is accepted, to deal with high load.
I'd like to point out: this is a double reduction, we are just following the previous behaviour and should be rather handled with correct set up of config parameters.

apps/arweave_config/include/arweave_config.hrl
khetzl: Yes, that's correct. These are all defaults, and can be overriden by providing parameters in the config.

apps/arweave_config/include/arweave_config.hrl
khetzl: I have no preference, I'm happy to introduce further grouping, if you guys think it makes more sense.
I tried to keep it as close to what we use internally in the RLG implementation for readability, but `.` vs `_` wouldn't make much of a difference in this sense.

JamesPiechota: :+1: I don't feel strongly either. We can defer to @humaite and if he doesn't have a clear preference, leave as is.

apps/arweave_limiter/src/arweave_limiter_group.erl
khetzl: `ok = Result = gen_server:call...` would crash if `Result` is not the atom `ok`.

`Result == ok andalso erlang_term` returns the `erlang_term` when `Result` is the atom `ok`.  `andalso` is a lazy operation, so the value of `erlang_term` is only calculated when the first bit is true.

So practically, it is just a shorthand for this:

```
case gen_server:call(LimiterRef, {reduce_for_peer, Peer}) of
    ok -> prometheus_counter:inc(ar_limiter_reduce_requests_total);
    _ -> false
end,
```

Since the server has the configuration, we can only determine whether the reduction should be/was performed after the call, and only want to increment the counter, if it indeed was.

JamesPiechota: Makes sense!

apps/arweave_limiter/src/arweave_limiter_group.erl
khetzl: Good catch. The RLG process doesn't start child processes, so it makes no difference. It is just a leftover from the emacs Erlang-mode template I use for different modules.

apps/arweave_limiter/src/arweave_limiter_group.erl
khetzl: Fair point. I'll change all similar ones to `LOG_DEBUG`

apps/arweave_limiter/src/arweave_limiter_group.erl
khetzl: Yes, that's expected. 

We don't care how much of the burst limit we have burned, if the sliding window allows us to accept a request.
The expiry still works, and the leaky token burns are not counted into the sliding window requests.

Also, the leaky token reduction still occurs in the background, so eventually it will reset the tokens to 0, if the sliding window limit is enough for that period of time.

JamesPiechota: Got it. Makes sense.

rebar.config
khetzl: No, it didn't. It is important to have call meck:new/1 separately for each module still.
Since, the update didn't break anything, I though might as well have the newer version then. One less dependency to update later.

apps/arweave_limiter/src/arweave_limiter_group.erl
khetzl: Good catch, These are typos, but they work as they are intended. I fixed them.

apps/arweave/src/ar_http_iface_rate_limiter_middleware.erl
```diff
@@ -0,0 +1,88 @@
...
+execute(Req, Env) ->
+    LimiterRef = get_limiter_ref(Req),
+    PeerKey = get_peer_key(Req),
+
+    case arweave_limiter:register_or_reject_call(LimiterRef, PeerKey) of
+        {reject, Reason, Data} ->
+            {stop, reject(Req, Reason, Data)};
+        _ ->
+            {ok, Req, Env}
+    end.
```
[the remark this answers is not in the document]
khetzl: This is intended, as discussed above. Good catch though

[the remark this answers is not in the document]
JamesPiechota: haha wow. The AI sounds just like me.

apps/arweave_limiter/src/arweave_limiter_sup.erl
```diff
@@ -0,0 +1,184 @@
...
+       sliding_window_limit => Config#config.'http_api.limiter.metrics.sliding_window_limit',
+       sliding_window_duration => Config#config.'http_api.limiter.metrics.sliding_window_duration',
+       timestamp_cleanup_tick_ms =>
+           Config#config.'http_api_limiter.metrics.sliding_window_timestamp_cleanup_interval',
+       timestamp_cleanup_expiry =>
+           Config#config.'http_api_limiter.metrics.sliding_window_timestamp_cleanup_expiry',
+       leaky_rate_limit => Config#config.'http_api.limiter.metrics.leaky_limit',
+       leaky_tick_ms => Config#config.'http_api.limiter.metrics.leaky_tick_intercal',
+       concurrency_limit => Config#config.'http_api.limiter.metrics.concurrency_limit',
+       is_manual_reduction_disabled => Config#config.'http_api.limiter.metrics.is_manual_reduction_disabled'},
```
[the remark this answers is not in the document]
khetzl: Good catch, none of us noticed this.

apps/arweave/src/ar.erl
```diff
@@ -37,6 +37,7 @@ main("") ->
 	show_help();
 main(Args) ->
 	arweave_config:start(),
+	arweave_limiter:start(),
```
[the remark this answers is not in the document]
khetzl: This was indeed a problem, with relying on relx to start the limiter app on the erlang node. It is now only loaded, so the code is available but we manually start the limiter from the code.

This has to be reworked once the config refactor is done, so we can define the right dependency tree in .app.src files, and rely on that to determine the right app start order.

apps/arweave_limiter/src/arweave_limiter_sup.erl
```diff
@@ -0,0 +1,184 @@
...
+    [
+     #{id => chunk,
+       sliding_window_limit => Config#config.'http_api.limiter.chunk.sliding_window_limit',
+       sliding_window_duration => Config#config.'http_api.limiter.chunk.sliding_window_duration',
+       timestamp_cleanup_tick_ms =>
+           Config#config.'http_api_limiter.chunk.sliding_window_timestamp_cleanup_interval',
+       timestamp_cleanup_expiry =>
+           Config#config.'http_api_limiter.chunk.sliding_window_timestamp_cleanup_expiry',
+       leaky_rate_limit => Config#config.'http_api.limiter.chunk.leaky_limit',
+       leaky_tick_ms => Config#config.'http_api.limiter.chunk.leaky_tick_intercal',
```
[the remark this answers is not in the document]
khetzl: This typo has been fixed

apps/arweave/src/ar_http_iface_rate_limiter_middleware.erl
```diff
@@ -0,0 +1,88 @@
...
+      #{},
+      <<"Too Many Requests">>,
+      Req
+     ).
+
+get_peer_key(Req) ->
+    {{A, B, C, D}, _Port} = cowboy_req:peer(Req),
+    {A, B, C, D}.
+
+config_peer_to_ip_addr({A, B, C, D}) -> {A, B, C, D}.
```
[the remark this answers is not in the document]
khetzl: Good catch from the AI

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,371 @@
...
+           sliding_window_duration => SlidingWindowDuration,
+           sliding_window_limit => SlidingWindowLimit,
+           sliding_timestamps => #{} %% Peer -> Ordered list of timestamps
+          }}.
+
+handle_call(reset_all, _From, State) ->
+    {reply, ok, State#{concurrent_requests => #{},
+                       concurrent_monitors => #{},
+                       leaky_tokens => #{},
+                       sliding_timestamps => #{}}};
```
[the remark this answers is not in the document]
khetzl: This has been handled in the `remove_concurrent/5`, if the monitor reference is not found in `concurrent_monitors/ConcurrentMonitors`, we don't do anything.

apps/arweave/src/ar_http_iface_rate_limiter_middleware.erl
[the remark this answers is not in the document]
khetzl: It feels like the bot has commented this once already, and it has been changed since then

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,377 @@
...
+          MonitorRef, Pid, Reason, ConcurrentRequests, ConcurrentMonitors),
+    {noreply, State#{concurrent_requests => NewConcurrentRequests,
+                     concurrent_monitors => NewConcurrentMonitors}};
+handle_info(Info, State = #{id := Id}) ->
+    ?LOG_WARNING([{event, unhandled_info}, {id, Id}, {module, ?MODULE}, {info, Info}]),
+    {noreply, State}.
+
+terminate(_Reason, #{leaky_tick_timer_ref := _LeakyRef,
+                     timestamp_cleanup_timer_ref := _TsRef} = _State) ->
+    ok.
```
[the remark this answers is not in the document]
khetzl: cancelling the timers crashes the tests.

apps/arweave_limiter/src/arweave_limiter_group.erl
```diff
@@ -0,0 +1,379 @@
...
+        remove_concurrent(
+          MonitorRef, Pid, Reason, ConcurrentRequests, ConcurrentMonitors),
+    {noreply, State#{concurrent_requests => NewConcurrentRequests,
+                     concurrent_monitors => NewConcurrentMonitors}};
+handle_info(Info, State = #{id := Id}) ->
+    ?LOG_WARNING([{event, unhandled_info}, {id, Id}, {module, ?MODULE}, {info, Info}]),
+    {noreply, State}.
+
+terminate(_Reason, _State) ->
+    ok.
```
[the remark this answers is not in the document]
khetzl: this would crash the nodes on ending tests

apps/arweave_limiter/src/arweave_limiter_metrics.erl
```diff
@@ -0,0 +1,64 @@
...
+%%%===================================================================
+
+%% @doc Declare Arweave Rate Limiter metrics.
+register() ->
+    ok = prometheus_histogram:new([
+                                   {name, ar_limiter_response_time_microseconds},
+                                   {help, "Time it took for the limiter to respond to requests"},
+                                   %% buckets might be reduced for production
+                                   {buckets, [0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1, 5, 10, 50]},
+                                   {labels, [limiter_id]}]),
```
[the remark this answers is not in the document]
khetzl: This is simple not true; the buckets for some reason ar in milliseconds, and values on the metrics interface returned in milliseconds, even though we report in microseconds.

apps/arweave_limiter/src/arweave_limiter_group.erl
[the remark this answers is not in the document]
khetzl: Yeah, he's commenting this over and over again, but cancelling would crash the tests, so we are not doing it. these processes should live for the lifetime of the node anyway.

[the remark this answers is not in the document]
JamesPiechota: Maybe if you resolve all instances of the comment it will stop?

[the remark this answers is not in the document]
JamesPiechota: Oh looks like a known issue: https://forum.cursor.com/t/bugbot-should-filter-out-previously-resolved-dismissed-issues/147748

apps/arweave_limiter/src/arweave_limiter_group.erl
khetzl: Fixed

apps/arweave/test/ar_http_iface_tests.erl
```diff
@@ -281,18 +282,22 @@ test_single_regossip(_) ->
 test_node_blacklisting_get_spammer() ->
 	{ok, Config} = arweave_config:get_env(),
 	{RequestFun, ErrorResponse} = get_fun_msg_pair(get_info),
+	LimitWithBursts = Config#config.'http_api.limiter.general.sliding_window_limit'
+		+ Config#config.'http_api.limiter.general.leaky_limit',
 	node_blacklisting_test_frame(
 		RequestFun,
 		ErrorResponse,
-		Config#config.requests_per_minute_limit div 2 + 1,
+		LimitWithBursts,
```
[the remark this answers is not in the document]
khetzl: This is incorrect, the test asserts a range correctly

apps/arweave/test/ar_http_iface_tests.erl
```diff
@@ -24,6 +24,7 @@ start_node() ->
 
 reset_node() ->
 	ar_blacklist_middleware:reset(),
+	arweave_limiter_sup:reset_all(),
```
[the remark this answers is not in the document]
khetzl: That's ok, we set a much higher rate (like previous solution) for tests, it won't reach the threshold. But good catch anyway, probably at some point we should call reset on the remote nodes.

apps/arweave/src/ar_http_iface_rate_limiter_middleware.erl
```diff
@@ -0,0 +1,90 @@
...
+execute(Req, Env) ->
+    LimiterRef = get_limiter_ref(Req),
+    PeerKey = get_peer_key(Req),
+
+    case arweave_limiter:register_or_reject_call(LimiterRef, PeerKey) of
+        {reject, Reason, Data} ->
+            {stop, reject(Req, Reason, Data)};
+        _ ->
+            {ok, Req, Env}
+    end.
```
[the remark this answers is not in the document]
khetzl: Discussed above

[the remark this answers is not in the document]
khetzl: don't post this again, please

PR #951, opened 2026-01-28
Lb/localnet

ldmberman:
> [!NOTE]
> **High Risk**
> Touches mining/nonce-limiter validation, recall-range computation, storage/rocksdb open semantics, and mempool overspend logic; regressions could impact consensus validation, syncing, or node startup from state snapshots.
> 
> **Overview**
> Adds a **LOCALNET** runtime mode and tooling to run a single-node, snapshot-seeded network. This introduces `ar_localnet` (start from a snapshot dir, seed data roots/tx data, and create new snapshots) plus a `localnet_shell` entrypoint and a dedicated `ar_localnet_mining_server`/supervisor wired via `ar_node_worker` to support deterministic “mine one block / mine until height” workflows.
> 
> Extends “start from state” to accept an explicit `start_from_state` folder (CLI/config), validates it against `data_dir`, and updates startup/join logic to read block index/history/wallet state from that folder; related changes include `ar_node:read_recent_blocks/3`, disk-cache lookups with an optional custom directory, and multiple DB open callsites updated to use the new `ar_kv` API.
> 
> Refactors `ar_kv` database opening to a map-based `open/1` (plus `open_readonly/1` and `close/1`) with explicit paths/log paths and read-only support, then updates storage/sync components accordingly; also adds a synchronous `store_data_roots_sync/4` path and a new HTTP `POST /data_roots/{offset}` endpoint to accept validated data-root indices.
> 
> Makes LOCALNET-specific consensus relaxations for testing (e.g., optional precalculated recall ranges, step-count/timeline comparisons, and difficulty checks), and improves mempool overspend filtering by tracking per-origin spent totals (with denomination-aware redenomination) to drop the lowest-priority spends when an origin exceeds confirmed balance.
> 
> CI/dev tooling updates include enabling Git LFS on checkouts (and tracking `localnet_snapshot/**` via LFS), adding headless notebook jobs to the full test workflow, and adding Jupyter config to strip notebook outputs by default.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 94d71668a779303b3dad6ec900f6a835799b4cef. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/arweave/src/ar_node_worker.erl
```diff
@@ -680,12 +716,25 @@ handle_task({cache_missing_txs, BH, TXs}, State) ->
 			{noreply, State}
 	end;
 
-handle_task(mine, State) ->
+handle_task(mine_one_block, State) ->
```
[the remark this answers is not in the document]
ldmberman: mine_one_block is not supposed to be called when the mining server is running. I do not like stopping the already running mining server by mine_one_block, because the outcome won't be deterministic and it may create wrong expectations.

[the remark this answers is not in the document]
ldmberman: mine_one_block now returns an error if the mining server is already running.

apps/arweave/src/ar_localnet_mining_server.erl
```diff
@@ -0,0 +1,329 @@
...
+	end.
+
+find_offset_in_intervals([{End, Start} | Rest], Offset) ->
+	Len = End - Start,
+	case Offset < Len of
+		true ->
+			Start + Offset;
+		false ->
+			find_offset_in_intervals(Rest, Offset - Len)
+	end.
```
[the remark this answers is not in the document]
ldmberman: It should never happen, a clearer error message thus does not seem that valuable here.

apps/arweave/src/ar_node_worker.erl
```diff
@@ -98,12 +127,12 @@ init([]) ->
 	{ok, Config} = arweave_config:get_env(),
 	validate_trusted_peers(Config),
 	StartFromLocalState = Config#config.start_from_latest_state orelse
-			Config#config.start_from_block /= undefined,
+			Config#config.start_from_block /= not_set,
```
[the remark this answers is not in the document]
ldmberman: It is not_set by default in apps/arweave_config/include/arweave_config.hrl

apps/arweave/src/ar_node_worker.erl
```diff
@@ -328,6 +357,14 @@ calculate_delay(Bytes) ->
...
 	BaseDelay + NetworkDelay.
 
+handle_call(mine_one_block, _From, State) ->
+	case maps:get(miner_state, State) of
+		undefined ->
+			State2 = start_mining(State),
+			{reply, ok, State2};
+		_ ->
+			{reply, {error, mining_server_running}, State}
+	end;
```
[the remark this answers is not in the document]
ldmberman: A block will be mined by construction, we won't implement this scenario here

apps/arweave/src/ar_node_worker.erl
```diff
@@ -1656,14 +1717,14 @@ start_mining(State) ->
...
-			ar_mining_server:start_mining({DiffPair, MerkleRebaseThreshold, Height}),
 			State#{ miner_state => running };
-		_ ->
-			ar_mining_server:set_difficulty(DiffPair),
-			ar_mining_server:set_merkle_rebase_threshold(MerkleRebaseThreshold),
-			ar_mining_server:set_height(Height),
+		running ->
+			?MINING_SERVER:set_difficulty(DiffPair),
+			?MINING_SERVER:set_merkle_rebase_threshold(MerkleRebaseThreshold),
+			?MINING_SERVER:set_height(Height),
```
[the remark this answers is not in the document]
ldmberman: I prefer not to hide the potential issue behind _ here.

apps/arweave/src/ar_mining_server.erl
```diff
@@ -174,10 +176,7 @@ get_packing_type(Packing) ->
 %%%===================================================================
 
 init([]) ->
-	%% Trap exit to avoid corrupting any open files on quit.
-	process_flag(trap_exit, true),
```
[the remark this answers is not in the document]
ldmberman: This (opening all files in advance) was an optimization no longer needed in 2.9 where read rate is always low.

apps/arweave/src/ar_mining_worker.erl
```diff
@@ -661,19 +652,45 @@ process_all_sub_chunks(WhichChunk, Rest, Candidate, Nonce, State) ->
...
+			case ar_mining_cache:with_cached_value(
+				?CACHE_KEY(Candidate#mining_candidate.cache_ref, Candidate#mining_candidate.nonce),
+				Candidate#mining_candidate.session_key,
+				State#state.chunk_cache,
+				fun
+					(#ar_mining_cache_value{chunk2_missing = true}) ->
+						{ok, drop, -SubChunkSize};
+					(#ar_mining_cache_value{chunk2 = Chunk2}) when is_binary(Chunk2) ->
+						{ok, drop, -SubChunkSize};
+					(#ar_mining_cache_value{chunk2 = undefined} = CachedValue) ->
```
JamesPiechota: it's been a while since I've been in the mining cache. Can you remind me why we need to set chunk1_missing here rather than just dropping the cache entry entirely? Is it because if chunk2 arrives an dthere's no cache entry it will create one? (i.e. that's the "orphaned entry" you mention)

ldmberman: Yes, chunk2 needs to know chunk1 failed and just drop the case. I want to also rename chunk1_missing into chunk1_failed, because missing feels like missing "so far" to me.

apps/arweave/src/ar_localnet_mining_server.erl
```diff
@@ -0,0 +1,339 @@
...
+		solution_hash = SolutionHash,
+		start_interval_number = StartIntervalNumber,
+		step_number = StepNumber,
+		packing_difficulty = ?REPLICA_2_9_PACKING_DIFFICULTY,
+		replica_format = 1,
+		last_step_checkpoints = Checkpoints,
+		steps = Steps
+	},
+	ar_node_worker:found_solution(miner, Solution, undefined, undefined),
+	ok.
```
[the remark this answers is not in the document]
ldmberman: In localnet we allow same VDF step. Won't fix

apps/arweave/src/ar_bridge.erl
```diff
@@ -38,6 +39,12 @@ block_propagation_parallelization() ->
 start_link(Name, Workers) ->
 	gen_server:start_link({local, Name}, ?MODULE, Workers, []).
 
+start_gossip() ->
```
JamesPiechota, line 42: Are these really only for use in localnet?

apps/arweave/src/ar_block_pre_validator.erl
```diff
@@ -717,9 +732,37 @@ pre_validate_pow_2_6(B, PrevB, PartitionUpperBound, Peer) ->
 			end
 	end.
 
+-ifdef(LOCALNET).
+get_precalculated_recall_range(B) ->
```
JamesPiechota, line 740: Maybe a short comment about what each of the LOCALNET-only function definitions are doing/why?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -1530,10 +1516,32 @@ handle_call({add_block, B, SizeTaggedTXs}, _From, State) ->
...
+	BlockSize = BlockEnd - BlockStart,
+	lists:foreach(
+		fun({DataRoot, TXSize, TXStartOffset, TXPath}) ->
+			ok = update_data_root_index(DataRoot, TXSize, TXStartOffset, TXPath, ?DEFAULT_MODULE)
+		end,
+		Entries
+	),
+	ok = ar_kv:put({data_root_offset_index, ?DEFAULT_MODULE},
+			<< BlockStart:?OFFSET_KEY_BITSIZE >>,
+			term_to_binary({TXRoot, BlockSize, DataRootIndexKeySet})),
```
JamesPiechota, line 1542: Does this need a [safe]?

JamesPiechota: (since it looks like the values can come from a client?)

apps/arweave/src/ar_data_sync.erl
```diff
@@ -2355,40 +2363,49 @@ init_kv(StoreID) ->
 		{"disk_pool_chunks_index", BasicOpts ++ BloomFilterOpts},
 		{"migrations_index", BasicOpts}
 	],
+	{ok, Config} = arweave_config:get_env(),
+	DataDir = Config#config.data_dir,
 	Dir =
 		case StoreID of
 			?DEFAULT_MODULE ->
-				?ROCKS_DB_DIR;
+				filename:join(DataDir, ?ROCKS_DB_DIR);
```
JamesPiechota, line 2371: How did this work before? Wouldn't it have put the ROCKS_DB_DIR relative to the CWD? Or maybe relative to the build path?

apps/arweave/src/ar_data_sync.erl
ldmberman: term_to_binary does not accept safe. Yes, we later decode it back via binary_to_term, however we validate the format is {binary(), integer(), sets:set(binary())} before storing so it is safe to read it.

apps/arweave/src/ar_data_sync.erl
ldmberman: Before we built the absolute path at the last moment in ar_kv.erl. This is [the commit](https://github.com/ArweaveTeam/arweave/pull/951/changes/fa45f07b8c74c3d857686dcb8bdca8d7bd4b34a3#diff-fe20359376b0502ac56aa6096531a0fb999833b2a6ae0c72fa6c7cf89946d627) where I changed the approach. Now the absolute path is passed by the client - I needed to pass the localnet snapshot path when reading from snapshot databases.

apps/arweave/src/ar_bridge.erl
ldmberman: `start_gossip` and `stop_gossip` are used in `ar_test_node:with_gossip_paused/2` I implemented [here](https://github.com/ArweaveTeam/arweave/pull/951/changes/38132f046199a4b72c793e7e8cb8cb2938d07c41) to fix some flakiness in `ar_vdf_block_validation_tests`. `connect_from_peer/disconnect_from_peer` do not work because they also interfere with VDF communication.

apps/arweave/src/ar_block_pre_validator.erl
ldmberman: Added some comments

apps/arweave/src/ar_kv.erl
```diff
@@ -81,26 +83,39 @@ create_ets() ->
...
-%% relative to the data directory and identified by the given Name.
-open(DataDirRelativePath, CfDescriptors, UserOptions, CfNames) ->
-	gen_server:call(
-		?MODULE, {open, {DataDirRelativePath, CfDescriptors, UserOptions, CfNames}}, ?DEFAULT_CALL_TIMEOUT
-	).
+%% @doc Open a key-value store in read-only mode.
+%% This will not modify any database files (no WAL writes, no compaction, no manifest updates).
+%% Useful for reading snapshot data without altering it.
+%% Args is a map with the same keys as open/1.
+open_readonly(Args) ->
```
JamesPiechota, line 117: Oh nice - do you know if this would allow 2 arweave nodes to read from the same set of storage modules? e.g. perhaps to have two nodes both sharing unpacked data from the same disks, or perhaps to have two nodes both doing a cross-module repack with the same unpacked data as the readonly source?

apps/arweave/src/ar_block_pre_validator.erl
JamesPiechota: Thanks!

apps/arweave/src/ar_bridge.erl
JamesPiechota: ah got it, so these are test-only functions?

apps/arweave/src/ar_data_sync.erl
```diff
@@ -2355,40 +2363,49 @@ init_kv(StoreID) ->
...
-			[{ar_data_sync, StoreID}, {chunks_index, StoreID}, {data_root_index_old, StoreID},
+	ok = ar_kv:open(#{
+		path => filename:join(Dir, "ar_data_sync_db"),
+		cf_descriptors => ColumnFamilyDescriptors,
+		cf_names => [{ar_data_sync, StoreID}, {chunks_index, StoreID}, {data_root_index_old, StoreID},
 			{data_root_offset_index, StoreID}, {tx_index, StoreID}, {tx_offset_index, StoreID},
-			{disk_pool_chunks_index_old, StoreID}, {migrations_index, StoreID}]),
-	ok = ar_kv:open(filename:join(Dir, "ar_data_sync_chunk_db"), [{max_open_files, 10000},
+			{disk_pool_chunks_index_old, StoreID}, {migrations_index, StoreID}]}),
+	ok = ar_kv:open(#{
```
JamesPiechota, line 2381: Loving these changes. Makes it a lot easier to read

apps/arweave/src/ar_localnet.erl
```diff
@@ -0,0 +1,1150 @@
...
+-define(DEFAULT_SNAPSHOT_DIR, "localnet_snapshot").
+
+-ifndef(START_FROM_STATE_SEARCH_DEPTH).
+	-define(LOCALNET_START_FROM_STATE_SEARCH_DEPTH, 100).
+-else.
+	-define(LOCALNET_START_FROM_STATE_SEARCH_DEPTH, ?START_FROM_STATE_SEARCH_DEPTH).
+-endif.
+
+-define(LOCALNET_DATA_DIR, ".tmp/data_localnet_main").
+
```
JamesPiechota, line 20: What do you think about having codex/claude generate function comments for all the functions in this file? I realize I'm not sure what they are all supposed to do.

apps/arweave/src/ar_bridge.erl
ldmberman: yes, they are

apps/arweave/src/ar_kv.erl
ldmberman: Yes, I think, we could implement that

apps/arweave/src/ar_localnet.erl
ldmberman: Done

apps/arweave/src/ar_kv.erl
JamesPiechota: nice. not a priority now, but good to know for later

apps/arweave/src/ar_localnet.erl
```diff
@@ -0,0 +1,1259 @@
...
+		case Config#config.mining_addr of
+			not_set ->
+				ar_wallet:to_address(ar_wallet:new_keyfile({?ECDSA_SIGN_ALG, secp256k1}, wallet_address, DataDir));
+			Addr ->
+				Addr
+		end,
+	StorageModules =
+		case Config#config.storage_modules of
+			[] ->
+				[{21 * ?MiB, 0, {replica_2_9, MiningAddr}}];
```
JamesPiechota, line 55: An particular reason for 21 MiB? Or not important?

apps/arweave/src/ar_localnet.erl
```diff
@@ -0,0 +1,1259 @@
...
+	ar_util:do_until(
+		fun() -> ar_node:is_joined() end,
+		100,
+		?WAIT_UNTIL_JOINED_TIMEOUT
+	 ).
+
+%% @doc Read recent block heads, tx headers, block time history, and account tree data
+%% from the snapshot directory and store them in the data directory. Does not do anything
+%% if recent blocks are already available locally.
+store_snapshot_data2(SnapshotDir) ->
```
JamesPiechota, line 150: I don't see a `store_snapshot_data` should this be renamed to `submit_snapshot_data2`?

apps/arweave/src/ar_localnet.erl
```diff
@@ -0,0 +1,1259 @@
...
+	InterimRewardHistoryLength0 = (Height - ar_fork:height_2_8()) + 21600,
+	InterimRewardHistoryLength =
+		case InterimRewardHistoryLength0 > 0 of
+			true ->
+				InterimRewardHistoryLength0;
+			false ->
+				0
+		end,
+	RewardHistoryBI0 = ar_rewards:trim_buffered_reward_history(Height, BI),
+	lists:sublist(RewardHistoryBI0, InterimRewardHistoryLength).
```
JamesPiechota, line 673: Can this be deduped with: https://github.com/ArweaveTeam/arweave/blob/lb/localnet/apps/arweave/src/ar_node_worker.erl#L1833-L1837

apps/arweave/src/ar_localnet.erl
```diff
@@ -0,0 +1,1259 @@
...
+	io_lib:format("spora_2_6(~s)", [ar_util:encode(Addr)]);
+format_packing({composite, Addr, Diff}) ->
+	io_lib:format("composite(~s, ~B)", [ar_util:encode(Addr), Diff]);
+format_packing({replica_2_9, Addr}) ->
+	io_lib:format("(replica_2_9, ~s)", [ar_util:encode(Addr)]);
+format_packing(unpacked) ->
+	"unpacked";
+format_packing(Other) ->
+	io_lib:format("~p", [Other]).
+
```
JamesPiechota, line 227: Possible to dedupe with ar_serialize:encode_packing?

apps/arweave/src/ar_localnet.erl
```diff
@@ -0,0 +1,1259 @@
...
+	),
+	io:format("Seeding completed. Total size: ~B bytes (~B big chunks, ~B small chunks).~n",
+		[TotalBigChunk + TotalSmallChunk, TotalBigChunk, TotalSmallChunk]),
+	{TotalBigChunk, TotalSmallChunk}.
+
+%% @doc For a given block start offset and its transactions, generate Merkle
+%% paths and data roots, register them with ar_data_root_sync, then write the
+%% raw transaction data to storage.
+%% Return {TotalBigChunkBytes, TotalSmallChunkBytes}.
+submit_block_data(BlockStart, TXs) ->
```
JamesPiechota, line 273: For these `submit_xxx` functions is the reason we call modules directly rather than going through the "normal" HTTP API for performance reasons? Or maybe because we're loading a snapshot from disk rather than normal syncing the normal API would reject the requests?

apps/arweave/src/ar_localnet.erl
```diff
@@ -0,0 +1,1259 @@
...
+		end,
+		{0, 0},
+		TXs2
+	).
+
+%% @doc Return the hardcoded map of {BlockStartOffset => [TXID]} for the seed
+%% data included in the localnet snapshot. These are real mainnet transactions
+%% whose data is bundled in the snapshot's seed_txs/ directory so the localnet
+%% node has chunk data to mine with.
+snapshot_txs() ->
```
JamesPiechota, line 320: Are the selected block height special? Or just picked to divide the blockchain into N roughly equal snapshots?

apps/arweave/src/ar_mempool.erl
```diff
@@ -199,7 +229,9 @@ drop_txs2(DroppedTXs, RemoveTXPrefixes, DropFromDiskPool) ->
 		{tx_priority_set, PrioritySet2},
 		{tx_propagation_queue, PropagationQueue2},
 		{last_tx_map, LastTXMap2},
-		{origin_tx_map, OriginTXMap2}
+		{origin_tx_map, OriginTXMap2},
+		{origin_spent_total_map, OriginSpentTotalMap2},
+		{origin_spent_total_denomination, Denomination}
```
JamesPiechota, line 234: I suspect it's not really important, but Denomination may still include the impact of TXs that were dropped in the loop above, right? Given how seldom denomination changes (if ever), my guess is this is not relevant in practice. Just trying to confirm my understanding

apps/arweave/src/ar_node_worker.erl
```diff
@@ -150,7 +179,7 @@ init([]) ->
 	%% May be start mining.
 	case Config#config.mine of
 		true ->
-			gen_server:cast(?MODULE, automine);
+			gen_server:cast(?MODULE, start_mining);
```
JamesPiechota, line 182:
```suggestion
			start_mining();
```

apps/arweave/src/ar_node_worker.erl
```diff
@@ -23,12 +24,22 @@
 
 -include_lib("arweave_config/include/arweave_config.hrl").
 
+-ifdef(LOCALNET).
+-define(MINING_SERVER, ar_localnet_mining_server).
+-else.
+-define(MINING_SERVER, ar_mining_server).
```
JamesPiechota, line 30: I still see `ar_mining_server` referenced explicitly in other modules - that okay?  i.e. is it only here that we need to swap `ar_localnet_mining_server` for `ar_mining_server`?

JamesPiechota: Ah it's only when these functions are called that we need to swap things out:
```
-callback start_mining({DiffPair, MerkleRebaseThreshold, Height}) -> ok when
	DiffPair :: {non_neg_integer() | infinity, non_neg_integer() | infinity},
	MerkleRebaseThreshold :: non_neg_integer() | infinity,
	Height :: non_neg_integer().

-callback pause() -> ok.

-callback is_paused() -> boolean().

-callback set_difficulty(DiffPair :: {non_neg_integer() | infinity, non_neg_integer() | infinity}) -> ok.

-callback set_merkle_rebase_threshold(Threshold :: non_neg_integer() | infinity) -> ok.

-callback set_height(Height :: integer()) -> ok.
```

JamesPiechota (approved):

